PR #48, opened 2015-11-20
Added online help

dfinke:


dfinke:
Don't know. I'm using VS 2015, not sure where they came from or what they are either.

PR #55, opened 2015-12-02
Add support for more variable scopes in debugger

rkeithhill:
This PR adds support for displaying Global and Script scope variables.  It also supports showing local variables for stack frames other than the topmost frame (frame 0).


daviwil:
Awesome work, thanks a lot Keith!

PR #59, opened 2015-12-03
Add "Expand Alias" support

dfinke:
Adds Expand Alias to the PowerShell extension


src/PowerShellEditorServices.Host/LanguageServer.cs
```diff
@@ -152,6 +153,46 @@ public void Initialize()
...
+            $lhs=$targetScript.Substring(0, $token.Start)
+            $rhs=$targetScript.Substring($token.Start + $token.Length)
+            
+            $targetScript=$lhs + $definition + $rhs
+       }
+    }
+
+    $targetScript
+}
+M
```
daviwil: Is the 'M' here a typo?

src/PowerShellEditorServices.Host/LanguageServer.cs
```diff
@@ -152,6 +153,46 @@ public void Initialize()
...
+Expand-Alias @'" + "\r\n" + content + "\r\n'@";
```
daviwil: This is a security risk since the content string is being inserted and executed directly as a script.  I think it might be safer if you take the "Expand-Alias" call out of the script string and then add another command to the psCommand to call it.  Something like this:

```csharp
psCommand
    .AddScript(script)
    .AddCommand("Expand-Alias")
    .AddArgument(@'" + "\r\n" + content + "\r\n'@");
```

daviwil: I wonder if it also makes sense to obfuscate the "Expand-Alias" command somehow so that it doesn't overwrite a similarly-named command that the user has imported/created in their session.  Maybe call it `__Expand-Alias` (two leading underscores)?

dfinke: Looks like it, I removed it on my box. Interestingly it works with and without it.

src/PowerShellEditorServices.Protocol/LanguageServer/PSModuleMessage.cs
```diff
@@ -0,0 +1,20 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
```
daviwil: Looks like this needs to be removed until the Gallery PR

daviwil:
Cool, looks like everything is good now!  Thanks for the awesome contribution!

Issue #61, opened 2015-12-07
Variables in the debugger are not displayed if type is PSObject


rkeithhill:
I see this when debubbing a filter definition like:
```
filter foo {
    "`$_ is $_"
}
```
In the Variables window `$_` shows up with no value.  I believe the problem is that the code that checks to see if the variable is expandable returns true on primitives wrapped in a PSObject.  I have a fix I'll submit in a few.

PR #62, opened 2015-12-07
Fix for #61 - PSObject not unwrapped when determining if a PSVariable…

rkeithhill:
…'s value is expandable.


daviwil:
Thanks Keith!

PR #64, opened 2015-12-10
Add new client/server API for the host process

daviwil:
This set of changes introduces a new client/server API which makes it much easier to create a language server or debug adapter implementation with matching C# client.  It also abstracts a "channel" model which enables different transport channels to be used with this client/server API.  For now only the standard I/O channel is implemented but other channels could easily be added.


daviwil:
Still finishing up this PR, need to clean up a few things and merge Doug's latest PR.

@adamdriscoll: Take a look at the changes in src\PowerShellEditorServices.Protocol\MessageProtocol\Channel and see if you think the HTTP/WebSockets implementation could be based on ChannelBase.

PowerShellEditorServices.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.40629.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.24720.0
```
daviwil: Revert?

src/PowerShellEditorServices.Host/LanguageServer.cs
```diff
@@ -3,98 +3,76 @@
...
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
-using System.IO;
+using DebugAdapterMessages = Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter;
 
 namespace Microsoft.PowerShell.EditorServices.Host
 {
-    internal class LanguageServer : IMessageProcessor
+    internal class LanguageServer : LanguageServerBase
```
adamdriscoll: Maybe I'm missing intention here but it seems like this class should be public since it implements a bunch of handy message handling stuff and right now it only allows the use StdioServer channel class. I know there aren't any other channels yet but I would think this class should have a constructor that accepts a ChannelBase.

src/PowerShellEditorServices.Host/DebugAdapter.cs
```diff
@@ -5,124 +5,84 @@
...
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Management.Automation;
 using System.Threading.Tasks;
 
 namespace Microsoft.PowerShell.EditorServices.Host
 {
-    internal class DebugAdapter : IMessageProcessor
+    internal class DebugAdapter : DebugAdapterBase
```
adamdriscoll: Same comment as with the LanguageServer class. But put me in my place if I'm not understanding the intent. :)

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/ChannelBase.cs
```diff
@@ -0,0 +1,118 @@
...
+using Newtonsoft.Json.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol.Channel
+{
+    /// <summary>
+    /// Defines a base implementation for servers and their clients over a
+    /// single kind of communication channel.
+    /// </summary>
+    public abstract class ChannelBase
```
adamdriscoll: This is great. I love how simple it is. I think it will make it much easier to implement additional channels.

src/PowerShellEditorServices.Host/LanguageServer.cs
daviwil: Yep, need to add a second constructor that takes a ChannelBase.  The reason why I didn't make this class public yet is that it's only meant to be used in the host process for now.  The way I generally see this working is that other channel implementations would be added to the Protocol library and then exposed through command line flags in the host process.

However, maybe I should just move the server implementations to the Protocol project for maximum flexibility.  Thoughts?

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/ChannelBase.cs
daviwil: Yeah, thought it presumes that every channel can be created from input/output System.IO.Streams which may not be true in all cases.  If it turns out that's not so easy, I may need to find a way to abstract MessageReader/Writer into interfaces that can expose other implementation types.

src/PowerShellEditorServices.Host/LanguageServer.cs
daviwil: I think moving these classes to Protocol is a good idea, there's no particular reason to make them only accessible through the host.

By the way, the reason why there are *Base implementations for the language/debug servers is that my next big change is factoring out the common protocol implementation pieces into a shared .NET library for speaking the VS Code protocol.  Much of the stuff that's currently in the Protocol library will get moved out to (yet another) library that will get used by other projects.

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/ChannelBase.cs
```diff
@@ -0,0 +1,118 @@
...
+        protected abstract void Shutdown();
+
+        /// <summary>
+        /// Sends an event to the channel's endpoint.
+        /// </summary>
+        /// <typeparam name="TParams">The event parameter type.</typeparam>
+        /// <param name="eventType">The type of event being sent.</param>
+        /// <param name="eventParams">The event parameters being sent.</param>
+        /// <returns>A Task that tracks completion of the send operation.</returns>
+        public Task SendEvent<TParams>(
```
daviwil: Need to move this logic to ProtocolServer since it isn't needed this low in the chain

src/PowerShellEditorServices.Host/LanguageServer.cs
adamdriscoll: My idea was to use something like Owin and then the actual HTTP hosting could be done in the host process, Nancy or IIS. So I would think having those classes exposed would be helpful.

daviwil: Yep, sounds like a good idea to me!

PowerShellEditorServices.sln
daviwil: Looks like the solution opens in VS 2013 with no complaints, will leave it like this.

daviwil:
After fighting with test failures all day, finally merging this!

adamdriscoll:
Sweet! Looks great!

daviwil:
Thanks!  I'll finally get NuGet packages posted sometime next week :)  Until then you might be able to get by with hacking in the host process while you're writing your new channel implementation.

PR #66, opened 2015-12-11
Fix casing of event messages from debug adapter

daviwil:
This change fixes the casing of event messages raised from the debug
adapter.  This issue caused the VS Code debugger UI to not display details
correctly when stopped at a breakpoint.


daviwil:
Merging the pull request despite what appears to be a test failure.  Still dealing with test hangs, will fix that after others are unblocked.

PR #70, opened 2015-12-11
Use a minimal set of Script Analyzer rules

daviwil:
This change greatly reduces the set of Script Analyzer rules that are used
by default in the analysis service.  In the future, this list of rules
will be fully configurable by the user.  This change is meant to simplify
the ruleset in the meantime.


daviwil:
cc @Jaykul @rkeithhill

daviwil:
Keith recommends adding `PSUseDeclaredVarsMoreThanAssigments`, any disagreement there?  I'll merge this PR sometime tomorrow if nobody has objections to the list of rules.

daviwil:
Need to take out `PSUseSingularNouns` before merging.

Issue #72, opened 2015-12-12
Display variable's type name


rkeithhill:
Visual Studio uses a three column grid to display a variable's name, value and type name.  In VSCode we have a tree view and the tool window it is in is typically narrow.  It is common to display variable name followed by its value.  However PowerShell quite often has objects (composite variables) where the current variables display shows no value for objects.  

I propose is that for expandable objects, we take advantage of the opportunity to display something.  One option is the variable's short type name e.g.:

![vscodevariabletypenames1](https://cloud.githubusercontent.com/assets/5177512/11760729/4a124dde-a063-11e5-90f0-274125696cc1.png)

Of course since we've chosen to prefix variables with `$` then perhaps we display type names using PowerShell syntax:
![vscodevariabletypenames2](https://cloud.githubusercontent.com/assets/5177512/11760733/6e0ff5ba-a063-11e5-8ba7-44e4b4fee487.png)
 
Another variation is to test the expandable object's ToString() return and *if* it is different than the type name, then display that instead.  Here's an example of where this is useful.  I can see the useful value of `$pwd` without having to open the node:
![vscodevariabletypenames3](https://cloud.githubusercontent.com/assets/5177512/11760773/65637480-a065-11e5-83ef-8e49f41aec8f.png)
In this case, I think seeing the path (ToString() value) is more useful than seeing `[PathInfo]`.  

What are your thoughts on this?  The implementation is a trivial change to VariableDetails.cs.  Personally, I'm leaning towards the last example of using either the type name or `ToString()` if it is different.

daviwil:
I think this will be really useful, and I'd go with the "type name or ToString() if different" option as well.  Thanks Keith!

rkeithhill:
Just about ready to check this in with a few additional goodies.  We now display string types in double quotes.  This helps you see empty strings easier and VSCode has nice syntax coloring for strings which also makes them stand out better. 
![vscodevariabletypenames4](https://cloud.githubusercontent.com/assets/5177512/11764077/cd697262-a0df-11e5-9735-f60e19d6c31b.png)
Also note that we are looking specifically at one dimensional arrays and object's implement ICollection and for those, we display the Length/Count of the array/collection.  See the $args object array above contains 2 items. 

Some collections like HashTable and ArrayList don't use `[]` in their type name, for those I'm using this approach `ArrayList: 1` e.g. see the $PSVersionTable below:
[...]

PR #74, opened 2015-12-12
Addresses proposal #72 - for expandable objects, display something us…

rkeithhill:
…eful.  This is typically the short type name.  However, for rank 1 arrays and objects implementing ICollection we now also display the number of items in the array/collection.  Also changed objects of type string to display their value in double quotes.


daviwil:
Excellent work man, another big improvement to the Variables view.

Issue #75, opened 2015-12-13
Variables display isn't showing contents for generic-based dictionaries


rkeithhill:
Specifically the `$PSBoundParameters` variable in `Local` scope of a function that specifies parameters. I've modified the code in `VariableDetails.GetChildren` like so:

```csharp
else if (dictionary != null)
{
    var res1 = dictionary.OfType<object>().ToList();
    var res2 = dictionary.OfType<KeyValuePair<string,object>>().ToList();
    foreach (var item in dictionary)
    {
        if (item.GetType() != typeof(DictionaryEntry))
        {
            break;
        }

        var entry = (DictionaryEntry)item;
        childVariables.Add(
            new VariableDetails(
                entry.Key.ToString(),
                entry.Value));
    }
    //childVariables.AddRange(
    //    dictionary
    //        .OfType<DictionaryEntry>()
    //        .Select(e => new VariableDetails(e.Key.ToString(), e.Value)));
}
```

The commented out code at the bottom is the original code.  The LINQ operator OfType<DictionaryEntry> returns nothing for `PSBoundParametersDictionary` because it derives from the generic type `Dictionary<string, object>`.  I can see in res1 that the type of object enumerated using the `OfType<>` method is `KeyValuePair<string,object>`.  However, if I just do the enumeration via a foreach loop, the item type works out to `DictionaryEntry` even for PSBoundParametersDictionary. 

It is late and I'm not quite grokking how this is working and what the *best* way is to handle both `IDictionary` and `IDictionary<TKey,TValue>`.  If anybody has any suggestions, let me know.  At the very least, the code above (sans the res1/res2 experiment) is better than what we have right now.

daviwil:
That is...  weird.  There seems to be some implicit coercion going on when using the foreach approach.  I'm fine with doing it this way though!  I'd leave a comment above the loop explaining why it needs to remain a loop so that nobody changes it :)

rkeithhill:
OK I “think” I know the “why” now after spending some time in dotPeek.  

Here are the two ways to get an enumerator from a Dictionary<TKey,TValue) based object, that the `GetChildren()` method has tried:
```csharp
IDictionaryEnumerator IDictionary.GetEnumerator() 
{
    return new Enumerator(this, Enumerator.DictEntry);
}

IEnumerator IEnumerable.GetEnumerator() 
{
    return new Enumerator(this, Enumerator.KeyValuePair);
}
```
[...]

PR #76, opened 2015-12-14
For for issue #75, variables display isn't displaying anything for ge…

rkeithhill:
…neric dictionaries like PSBoundParametersDictionary.  Also enhanced the determination of what is expandable.  We were not allowing structs like DateTime, TimeSpan and DictionaryEntry to expand.  Changed test from !IsValueType to !IsPrimitive.  Also enhanced the display of property values where we encounter an exception when trying to retrieve the property.  Following VS debugger's lead on this on.   You can see the display by looking at $Host.Runspace.  And we also weren't displaying .NET properties for colletions and dictionaries.  We now do that - in addition to the elements in the collection or dictionary.  Also, don't try to add properties that are indexers.

@daviwil - this one is a bit more change in VariablesDetails.cs than I was anticipating.  Please give it a good look over.

### DIFF
--- a/src/PowerShellEditorServices/Debugging/VariableDetails.cs
+++ b/src/PowerShellEditorServices/Debugging/VariableDetails.cs
@@ -130,6 +130,7 @@ namespace Microsoft.PowerShell.EditorServices
             return
                 valueObject != null &&
                 !valueType.IsPrimitive &&
+                !(valueObject is decimal) &&
                 !(valueObject is UnableToRetrievePropertyMessage) &&
                 !(valueObject is string); // Strings get treated as IEnumerables
         }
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Debugging/VariableDetails.cs
```diff
@@ -223,59 +241,55 @@ private static VariableDetails[] GetChildren(object obj)
...
-                        objectType.GetProperties(
-                            BindingFlags.Public | BindingFlags.Instance);
-
-                    foreach (var property in properties)
-                    {
-                        try
+                    // We're in the realm of regular, unwrapped .NET objects
+                    if (dictionary != null)
+                    { 
+                        // Buckle up kids, this is a bit weird.  We could not use the LINQ
```
daviwil: Great comment!  Thanks for adding that

daviwil:
Excellent work yet again, thank you for carrying us through the Hack Week :)

PR #80, opened 2015-12-16
WebSocket Channel

adamdriscoll:
Implements WebSocket support using OWIN.

This PR isn't quite done but wanted to started a discussion.

Questions:
- WebSocket unit testing is tricky. I can self-host OWIN and test that way but it's not quite a unit test then. 
- Should this be a separate library or is it cool in the main protocol library? It does introduce OWIN dependencies.


PowerShellEditorServices.sln
```diff
@@ -1,99 +1,99 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
daviwil: Weird, is there a line ending conversion happening that makes these files seem like every line changed?  What response do you get from this command: `git config core.autocrlf`

adamdriscoll:
After looking at this PR, I'm going to close it and create one where it's actually readable...

PowerShellEditorServices.sln
adamdriscoll: It returns true. Is that why this merge looks like crap?

daviwil: Mine returns `true` also so I'm not sure why this would be happening.  Strange!

adamdriscoll: I'm going to just kill my fork and copy in my changes. It's messy anyways.

daviwil:
Regarding your question about which library this should be in:  I think this might be a good time to establish a new channel library pattern by creating a new PowerShellEditorServices.Channel.WebSocket project under the `src` folder.  I think I'll keep the stdio channel in-box but all other channels can be extensions to the existing design.  Is it possible that the code that's in your Host.IIS library could be put there?

For unit testing, I'm totally fine with self-hosting the socket with OWIN and setting up a local client/server in the unit test code.  I'm doing that pretty extensively with the stdio implementation so I don't see a big difference (so long as we can catch the inevitable configuration problems that self-hosting OWIN on a CI server will cause).

PowerShellEditorServices.sln
daviwil: Check out this StackOverflow answer: http://stackoverflow.com/a/20653073

Maybe core.autocrlf needs to be set to 'input' ?  I've never had to use that before, though

adamdriscoll: Thanks. I'll futz with it. I have a nice clean fork now. :)

adamdriscoll:
Channel library convention sounds good to me. I'll make that change. As for the host IIS project, I think I'll just get rid of that. It's really just a standard OWIN ASP.NET host. It would make a good example but not a part of the SDK. 

Unit testing: Perfect. I'll spend some time on that after I move things around.

daviwil:
My ultimate goal is to have one set of LanguageServer and DebugAdapter unit tests that get used commonly across all channel and interface implementations, so don't worry about writing a bunch of unit tests for the new channel other than just making sure that the client/server can send/receive messages.  Once I get back to the ILanguageService/IDebugService interface stuff this will all light up pretty fast :)

adamdriscoll:
Good to know. I was thinking of inheriting from them someone how for my tests but if you are working on it already I'll let it be. :D

Jaykul:
By the way @adamdriscoll, this is GitHub -- around here it's perfectly acceptable to rebase and `push -f` your pull requests to clean stuff up ;-)

daviwil:
That's basically what I was going to do, make an abstract base that accepts a channel as constructor parameter.  I've got a working version of it in an unfinished branch.

daviwil:
Haha, yeah, that's what I do, `push -f` about 20 times before I'm happy with it.

PR #82, opened 2015-12-17
Added more DebuggerDisplayAttributes to make debugging of the editor …

rkeithhill:
…services easier.


daviwil:
Awesome, thanks for settings those up!

Issue #87, opened 2015-12-19
Preference variables show as expandable but there is nothing to show when expanded


rkeithhill:
It looks like the only property is `value__`.   I think the more general case is how we handle Enums in the variables window.

PR #88, opened 2015-12-21
Fix for #87, enums display as expandable when they have no properties…

rkeithhill:
….  VS also doesn't show enum values as expandable.  Note that this is a specific fix for enums but I think we need a more general fix where we don't show any variable as expandable *if* it has no properties to show.


daviwil:
Excellent, thanks Keith!

Issue #73, opened 2015-12-12
Support PowerShell v3 and v4


daviwil:
Currently we support only PowerShell v5 but it wouldn't take a great deal of effort to support PowerShell 3 and 4.  [PoshTools](github.com/adamdriscoll/poshtools) has accomplished this already so there's some work there to be learned from.

The newly released PowerShell reference assemblies could help build version-specific implementation DLLs:  http://blogs.msdn.com/b/powershell/archive/2015/12/12/powershell-sdk-reference-assemblies-available-via-nuget-org.aspx

adamdriscoll:
I'm curious your thoughts on how this will be accomplished. In PoshTools, it really isn't doing anything with S.M.A.dlls. It's more lucky that the assembly version didn't change and we do some checking to prevent MissingMethodExceptions. I think this will eventually lead to a problem if the version does change so it would be good to figure out a way to accomplish it correctly.

One thing that is done via AppVeyor is to build 2013 and 2015 extensions. This is done using some custom MSBuild targets\props. 

Are you thinking that we would have different project files or do something similar with custom MSBuild targets?  I think the latter would keep the VS solution tidy but is a little more work. The one thing I'm a bit unclear about would be the packages.config file. It seems like we'd need a separate one for each build type or have it patched before package restore kicks in.

I'm sure I'm just totally overthinking this. :)

daviwil:
The way you've done it is probably the best way to do it for now.  For PowerShell v3 through v5 we can probably use the same implementation DLL.  If for some reason we have a different SMA assembly version in the future which breaks compatibility with previous code contracts, we'll need to have a different core logic assembly for Editor Services that we hide behind a well-defined interface.

Here's a potential design:

- `Microsoft.PowerShell.EditorServices.dll` shifts from providing implementations for all features to defining the core feature interfaces that the Editor Services library exposes.  It would also provide the library 'lifetime' code that simplifies the use of feature implementations for any version.  The caller would specify which implementation assembly to use when they create a new EditorSession.

- Introduce `Microsoft.PowerShell.EditorServices.v345.dll` which contains the feature implementations that were previously found in the core DLL.

- There could potentially be some 'common' DLL that could share any feature implementations that are version-independent but I'm not sure how likely (or easy) that will be.  This would be used as a dependency to every version-specific implementation assembly.

[...]

PR #103, opened 2016-01-04
PowerShell v3/v4 Support

adamdriscoll:
Step towards #73 

This PR adds a test and some logic into PSES to account for different PS versions. That said, there is an issue with the current code base that we need to work out before this actually works completely. The use of the AstVisitor2 class is going to cause problems for v3 and v4. Not quite sure how to work around this. 

- Adds a test that runs for each version of PS to make sure we can build for that version
- Pre-compiler constants added for each PS version (defaults to v5), added for tests
- Added some logic to to avoid MissingMethodExceptions when running on v3\v4
- Added new property to PowerShellContext that exposes PS version


adamdriscoll:
Annnddd we'll see if this build passes. Doing some interesting stuff with MSBuild in this PR.... :D

adamdriscoll:
Tests hang in my environment as well when I run the entire batch. If I run just the new tests they complete successfully. Something is wonky....

src/PowerShellEditorServices/PowerShellEditorServices.csproj
```diff
@@ -49,12 +52,15 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Management.Automation, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\..\..\..\..\..\Windows\Microsoft.NET\assembly\GAC_MSIL\System.Management.Automation\v4.0_3.0.0.0__31bf3856ad364e35\System.Management.Automation.dll</HintPath>
```
daviwil: Should we take out the HintPath?  I dunno if it might be differnet somehow on other people's machines

src/PowerShellEditorServices/Session/PowerShellContext.cs
```diff
@@ -120,6 +129,21 @@ private void Initialize(Runspace initialRunspace)
             // TODO: Should this be configurable?
             this.SetExecutionPolicy(ExecutionPolicy.RemoteSigned);
 
+            var psVersionTable = this.currentRunspace.SessionStateProxy.GetVariable("PSVersionTable") as Hashtable;
```
daviwil: Any chance this could fail?  Should we put a try/catch around it?

daviwil: If so I'd say make it into its own private function, something like GetPowerShellVersion

src/PowerShellEditorServices/PowerShellEditorServices.csproj
adamdriscoll: Yeah. I'll fix that. I'm using it for the tests but it should be put in dynamically.

src/PowerShellEditorServices/Session/PowerShellContext.cs
adamdriscoll: Good point. Will do.

test/PowerShellEditorServices.Test/Language/LanguageServiceTests.cs
```diff
@@ -289,6 +298,96 @@ public void LanguageServiceFindsSymbolsInNoSymbolsFile()
             Assert.Equal(0, symbolsResult.FoundOccurrences.Count());
         }
 
+        [Theory]
+        [InlineData("3")]
+        [InlineData("4")]
+        [InlineData("5")]
+        public void CompilesWithPowerShellVersion(string version)
+        {
+            var assemblyPath = InstallPackage(string.Format("Microsoft.PowerShell.{0}.ReferenceAssemblies", version), "1.0.0");
```
daviwil: Awesome!  This is a neat way to do it.  Two things I'd recommend:

1. Maybe we can just make the reference assemblies a NuGet dependency of the test project so that we don't need to call NuGet from within the test code?  That way you can just reference the already-pulled dependencies directly.  Might make things easier for AppVeyor

2. It'd be nice to move these tests into their own test class/file, maybe something like PowerShellVersionTests.cs

test/PowerShellEditorServices.Test/PowerShellEditorServices.Test.csproj
```diff
@@ -104,7 +104,11 @@
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Content Include="NuGet.exe">
```
daviwil: The solution already has NuGet.exe in the root \.nuget folder so you could probably just use that directly rather than including another one here.  Might not be necessary at all if we pull in the reference assemblies as NuGet dependencies for the test project

daviwil:
Sweet, the tests passed!

adamdriscoll:
I was treating PSVersion as a string when it was actually a Version. Broke EVERYTHING. ;)

daviwil:
I'm trying to decide whether this is the best way to enable PowerShell 3-5 support.  Let me break down the pros/cons I see, feel free to suggest more:

**Pros**
- We can build against the official reference assemblies and be completely sure whether our version-specific implementations target the appropriate APIs for each version

**Cons**
- Requires independent Microsoft.PowerShell.EditorServices implementation assemblies for each of the three versions
- Potentially requires 3 more NuGet packages for the different versions (though it may be better to package them with the core DLL)
- Complicates the host startup strategy since we need to determine the system's PowerShell version and dynamically load the right implementation DLL.  We may need to do this anyway for full PS vs Nano Server PS, so I'm not completely sure whether this is a negative for now
- Feature implementation code gets a little more complicated with compiler define usage, but thankfully not by much

What do you think?  Are there more pros to this approach that I'm missing?

daviwil:
Though I just realized that maybe the point of this isn't to have 3 different assemblies, but just to verify that we're using the right APIs for the 3 different versions.  Is that what you're going for?

adamdriscoll:
Yeah. Your last comment is correct. I wasn't suggesting three different assemblies (even though the defined constants make that possible). It was just a validation step. PowerShellv5 is defined for all builds except the one done in that test because it overrides the MSBuild property. It's really easy to miss something without this kind of check because it will only happen on a system running that version of PS. You'll hear about it pretty quickly from users but won't necessarily catch it in development. Honestly, I really hate the use of constants like this because it's really ugly to read and is really easy to get wrong so I would understand if you felt we should go another way with this.

The implementation of the PowerShellVersion property and checking it in various places would be the real guard for running on different versions of PS. The one caveat is any extending from AstVisitor2. We might just need to guard against instantiating and have an alternate version that extends from AstVisitor for down level machines.

daviwil:
Now that I understand the intent, I think this is a really clever way to do it!  It will really help make sure we're using the right APIs for the 3 versions.  Let's go with this.

Here's a thought on the AstVisitor2 problem: maybe instead of duplicating the code for the 2 classes, we can have a standard AstVisitor implementation and then wrap it with an AstVisitor2 implementation.  For example, FindSymbolVisitor would derive from AstVisitor and implement the necessary methods there.  FindSymbolVisitor2 would derive from AstVisitor2 and implement methods specific to that interface, but also it would use the AstVisitor methods to call a wrapped FindSymbolVisitor instance that gets created in the constructor.  That way the AstVisitor method implementations get implemented once even though both classes ultimately have to contain the same methods.

Since the visitor classes are invoked from the methods in the AstOperations class, we can choose which visitor class implementation to use there based on a (new) version parameter.

What do you think?

adamdriscoll:
Sounds perfect to me! Will update this PR with the changes.

adamdriscoll:
Easy fix ;)

daviwil:
This is awesome man, thanks so much for putting it together!

PR #108, opened 2016-01-08
Rkeithhill/fix getvariables failed

rkeithhill:
Fix for Get-Variable -Scope doesn't always work. 

Ideally we should be using CallStackFrame.GetFrameVariables() but it currently seems to be broken.  We've been informed that using Get-Variable -Scope <num> isn't great because scope numbers and frames don't always line up.  For instance, dot source introducing a new stack frame but not a new scope.  But until GetFrameVariables is fixed in a future drop, we have no option but to use Get-Variable -Scope <num>.  This will result in stack frames that have no Auto or Local variables.  But the good news is that the debug host will no longer crash.  :-)

BTW this commit introduces a new optional parameter on PowerShellContext.ExecuteCommand().  While there is a parameter to suppress "output" there was no parameter to suppress "errors" from being written to the debug console.  I see no benefit in littering the user's debug console with a bunch of "get-variable : The scope number '#' exceeds the number of active scopes".  So there is now a "sendErrorToHost" parameter that defaults to true.  However, even when that parameter is set to false, the error info shows up in the DebugAdapter.log file.

### DIFF
--- a/src/PowerShellEditorServices/Debugging/DebugService.cs
+++ b/src/PowerShellEditorServices/Debugging/DebugService.cs
@@ -360,16 +360,19 @@ namespace Microsoft.PowerShell.EditorServices
                 new VariableContainerDetails(this.nextVariableId++, "Scope: " + scope);
             this.variables.Add(scopeVariableContainer);
 
-            var results = await this.powerShellContext.ExecuteCommand<PSVariable>(psCommand);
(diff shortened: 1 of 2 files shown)


daviwil:
Argh, looks like the test failure is a timeout due to a DebugAdapter test hang, sorry about that.  The second run just passed so I'm going to go ahead and merge this.  Thanks for the fix!

PR #120, opened 2016-01-31
Convert JSON|CSV to PS|C# class syntax

dfinke:
Convert JSON|CSV to PS|C# class syntax


daviwil:
The test failure is due to an intermittent lingering issue with debug adapter tests, so don't worry about that.  I'll wait to merge this until we figure out what to do about the dependency on the vscode-powershell side.

daviwil:
Discussed with Doug that we'll hold off on integrating this PR until we get the new command extensibility model going.  This feature would be a great validation of the new extensibility approach.  I'm keeping this PR open as a reminder to do that :)

daviwil:
Doug just shipped an update of his ConvertToClass module which leverages the $psEditor API! https://www.powershellgallery.com/packages/ConvertToClass/1.5.  Closing this PR now.

PR #121, opened 2016-02-02
Add text input prompt support

daviwil:
This change adds a PSHostUserInterface implementation for both the Prompt
and ReadLine methods so that input can be requested from the user.  The
implementation is made with the IPromptHandler interface so that the
prompt handling strategy can be changed based on the context of the
user's session.


daviwil:
Still finishing this up, just getting the PR started to make sure tests are passing.

src/PowerShellEditorServices.Protocol/Server/PromptHandlers.cs
```diff
@@ -21,6 +22,11 @@ public ChoicePromptHandler GetChoicePromptHandler()
         {
             return new ProtocolChoicePromptHandler(this.eventWriter);
         }
+
+        public InputPromptHandler GetInputPromptHandler()
+        {
+            throw new NotImplementedException();
```
daviwil: Need to add an implementation here

src/PowerShellEditorServices/Console/FieldDetails.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        private static Type GetFieldTypeFromTypeName(string assemblyFullName)
+        {
+            Type fieldType = typeof(string); 
+
+            if (!string.IsNullOrEmpty(assemblyFullName))
+            {
+                if (!LanguagePrimitives.TryConvertTo<Type>(assemblyFullName, out fieldType))
+                {
+                    // TODO: Log lack of ability to convert to type?
```
daviwil: Log here

src/PowerShellEditorServices/Console/InputPromptHandler.cs
```diff
@@ -0,0 +1,280 @@
...
+        /// <param name="responseString">The string representing the user's response.</param>
+        /// <returns>
+        /// True if the prompt is complete, false if the prompt is 
+        /// still waiting for a valid response.
+        /// </returns>
+        public bool HandleResponse(string responseString)
+        {
+            if (this.currentField == null)
+            {
+                // TODO: Assert
```
daviwil: Do something about this

src/PowerShellEditorServices/Console/InputPromptHandler.cs
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            // TODO: Is string empty?  Use default or finish prompt?
+            object responseValue = responseString;
```
daviwil: What does PowerShell console host do about default values?

daviwil:
Woot!  Merged.

daviwil:
Jeeeeez this feature is never-ending.  Still finishing up all the polish to make sure prompt cancellation works right and doesn't block future operations...

Issue #98, opened 2016-01-02
File path completions should not contain the current path prefix


rkeithhill:
This is what I'm referring to here:

![filecompletionlist](https://cloud.githubusercontent.com/assets/5177512/12072931/f88fca44-b0bc-11e5-88a4-90b277a98fc6.png)

This would also make the extension behave like ISE WRT completing absolute paths.  

This would also mitigate an issue where the tooltip clips the end of the path:
![clipfilecompletionlist](https://cloud.githubusercontent.com/assets/5177512/12072939/93098236-b0bd-11e5-8b70-10d319951f2c.png)

daviwil:
Interesting, looks like the ISE must be doing some post-processing of the completion results for paths.  I agree, this would be helpful!

rkeithhill:
So I've been looking at this.  I figured out how to get VSCode to display different text in the tooltip vs what it injects into the doc.  It requires modifying this LanguageServer.cs method:
```csharp
internal static CompletionDetails Create(CompletionResult completionResult, int replacementLength)
{
    Validate.IsNotNull("completionResult", completionResult);

    // Some tooltips may have newlines or whitespace for unknown reasons
    string toolTipText = completionResult.ToolTip;
    if (toolTipText != null)
    {
        toolTipText = toolTipText.Trim();
    }

    var completionDetails = new CompletionDetails
    {
        CompletionText = completionResult.CompletionText,
        ToolTipText = toolTipText,
        SymbolTypeName = ExtractSymbolTypeNameFromToolTip(completionResult.ToolTip),
        CompletionType = 
            ConvertCompletionResultType(
                completionResult.ResultType)
    };

    if (completionDetails.CompletionType == CompletionType.Path)
    {
        completionDetails.CompletionText = completionDetails.CompletionText.Substring(replacementLength);
        completionDetails.ToolTipText = completionDetails.ToolTipText.Substring(replacementLength);
    }

    return completionDetails;
}
```
The problem is determining what text to remove from the tooltip.  I thought maybe that the `ReplacementLength` in the CommandCompletion object would work e.g.:
```csharp
internal static CompletionDetails Create(CompletionResult completionResult, int replacementLength)
{
    Validate.IsNotNull("completionResult", completionResult);

    // Some tooltips may have newlines or whitespace for unknown reasons
    string toolTipText = completionResult.ToolTip;
    if (toolTipText != null)
    {
        toolTipText = toolTipText.Trim();
    }

[...]

PR #122, opened 2016-02-02
Implementation for issue #98 - shorting file path completion list.

rkeithhill:
This commit not only addresses issue #98 but also improves completion list support overall for commands (shows resolved command or path to exe), parameters (shows parameter type) as well as .NET properties and methods (shows signature).

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -994,10 +994,10 @@ function __Expand-Alias {
                 // Look for type encoded in the tooltip for parameters and variables.
                 // Display PowerShell type names in [] to be consistent with PowerShell syntax
                 // and now the debugger displays type names.
-                var matches = Regex.Matches(completionDetails.ToolTipText, @"^\[(.+)\]");
+                var matches = Regex.Matches(completionDetails.ToolTipText, @"^(\[.+\])");
                 if ((matches.Count > 0) && (matches[0].Groups.Count > 1))
                 {
-                    detailString = "[" + matches[0].Groups[1].Value + "]";
+                    detailString = matches[0].Groups[1].Value;
                 }
 
                 // PowerShell returns ListItemText for parameters & variables that is not prefixed
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -431,10 +431,6 @@ function __Expand-Alias {
 
             if (completionResults != null)
             {
-                // By default, insert the completion at the current location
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -989,23 +985,59 @@ private static CompletionItemKind MapCompletionKind(CompletionType completionTyp
...
-            if (completionDetails.CompletionType == CompletionType.Variable)
+            if ((completionDetails.CompletionType == CompletionType.Variable) ||
+                (completionDetails.CompletionType == CompletionType.ParameterName))
             {
-                // Look for variable type encoded in the tooltip
+                // Look for type encoded in the tooltip for parameters and variables.
+                // Display PowerShell type names in [] to be consistent with PowerShell syntax
+                // and now the debugger displays type names.
                 var matches = Regex.Matches(completionDetails.ToolTipText, @"^\[(.+)\]");
+                if ((matches.Count > 0) && (matches[0].Groups.Count > 1))
```
daviwil: This could be simplified by moving the parentheses outside of the square brackets in the regex so that the contents of match group 1 include the brackets.  Not worth making the change if you're already done with the work though :)

rkeithhill: Good point.  I can make that change and push it and it should show up on the PR.

test/PowerShellEditorServices.Test/Language/PowerShellVersionTests.cs
```diff
@@ -58,13 +58,23 @@ private void Compile(string project, string version)
             }
 
             FileInfo fi = new FileInfo(project);
+            string solutionDir = fi.Directory.Parent.Parent.FullName;
             var outPath = Path.GetTempPath();
 
             var p = new Process();
             p.StartInfo.FileName = msbuild;
-            p.StartInfo.Arguments = string.Format(@" {0} /p:Configuration=Debug /t:Build /fileLogger /flp1:logfile=errors.txt;errorsonly  /p:SolutionDir={1} /p:SolutionName=PowerShellEditorServices /p:DefineConstants=PowerShellv{2} /p:OutDir={3}", project, fi.Directory.Parent.Parent.FullName, version, outPath);
+            p.StartInfo.Arguments = string.Format(@" {0} /p:Configuration=Debug /t:Build /fileLogger " +
```
daviwil: Good catch, I wonder how this never affected me before

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
daviwil: Cool!  I'll merge it as soon as you're done with that.  Nice work!

daviwil:
Sweet!  Merged.

Issue #40, opened 2015-11-18
Improve parameter Intellisense sorting


rkeithhill:
Right now, command-specific parameters are mixed in with common parameters.  

![vscintellisense](https://cloud.githubusercontent.com/assets/5177512/11246425/d83fb44e-8dd4-11e5-9419-df74a24ff0d0.png)

Ideally, I'd love to see VSC support a tabbed Intellisense popup (like I've seen before in VS) where there would be two tabs as shown I the C# popup:

![vsintellisense](https://cloud.githubusercontent.com/assets/5177512/11246396/bacbbc50-8dd4-11e5-9422-dcbad1d81eef.png)

Where the first tab would show the command-specific parameters and the second tab would show common parameters.

If VSC doesn't support this then I propose that the parameters are listed in the following order:
* Sorted non-common parameters
* Line separator (if possible)
* Optional common parmeters
* Line separator (if possible)
* Sort common parameters

For example, this is an example of the above for Stop-Process:
```
Force
Id
InputObject
Name
PassThru
----
Confirm
WhatIf
----
Debug
ErrorAction
ErrorVariable
InformationAction
InformationVariable
OutBuffer
OutVariable
PipelineVariable
Verbose
WarningAction
WarningVariable
```
I think this order would be more useful to PowerShell scripters.  Personally, I'm almost always more interested in the command-specific parameters.

rkeithhill:
I looked at this a bit over the weekend.  It appears the extension is passing the parameters in the same (non-alpha) order to VSCode that I observe in ISE.  VSCode must be sort the list.  That's too bad, wish we could suppress that.

daviwil:
For the JSON protocol there's a 'SortText' field that I haven't been using yet.  I bet we'd get the right ordering if we prefix the label text to populate SortText based on the type of parameter.  Something like:

```
1 Force
1 Id
...
----
2 Confirm
2 WhatIf
----
3 Debug
3 ErrorAction
3 ErrorVariable
...
```

Could also just use an ordered integer converted to a string if you want to completely eliminate the sorting.  Check out the CreateCompletionItem method in `src\PowerShellEditorServices.Protocol\Server\LanguageServer.cs`.

rkeithhill:
Will do.  Thanks.

rkeithhill:
So just putting an increasing integer converted to string in the SortText field didn't do the trick.  Guess I need to read up on the CompletionItem semantics.

daviwil:
[...]

PR #124, opened 2016-02-02
Addresses issue #40 - improved parameter completion list order

rkeithhill:
This takes advantage of the fact that PowerShell returns us a list of parameter names already in the preferred order (common params at the end).  We just need to make the SortText for each completion result preserve the initial order.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -431,13 +431,15 @@ function __Expand-Alias {
 
             if (completionResults != null)
             {
+                int sortIndex = 1;
                 completionItems =
                     completionResults
                         .Completions
                         .Select(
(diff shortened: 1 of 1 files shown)


rkeithhill:
Forgot to mention.  This only works with a >= `0.10.7-insiders` build of VSCode.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1031,6 +1034,18 @@ private static CompletionItemKind MapCompletionKind(CompletionType completionTyp
                 }
             }
 
+            // We want a special "sort order" for parameters that is not lexicographical.
```
daviwil: Excellent comment, thank you!

daviwil:
Argh, that debug adapter test is really starting to annoy me.  Your change is great though, merging it!

Issue #96, opened 2015-12-29
Plan for packaging of the x86/x64 host EXE and CMD files


rkeithhill:
What is the plan for allowing a user to choose between the x86 and x64 host?  I was expecting to see two debugger configurations in the package.json file: “PowerShell x86” and “PowerShell x64”.  

I’m not sure which should be the default (probably the x64 host).  Anyway, I was thinking about changing the package.json file but I realize I’m not sure how the two EXE and CMD files will be placed in the extension's install dir.  The exe’s have different names so they could be co-located in the Bin dir but the CMD files have the same name.  Is this a case where the x86 CMD should have x86 put in its filename so it can be co-located in the Bin dir?

Thoughts?

adamdriscoll:
Different CMD names is probably a good idea. The x64 host is technically AnyCPU so if we default to that it would run on x86 as well.

daviwil:
Yep, agree with both your points.  We'll need a separate .cmd to run the x86 host and then have a separate debugger configuration for that.  x64 should be the default, and like Adam says, it should work fine on 32-bit systems.  

I was thinking about adding a new preference setting in the VS Code extension to choose which bitness to use for the language service.  It kinda sucks to have to change the bitness setting in two different places but maybe later we can find a clever way to make one setting that controls both.

Feel free to go for it, Keith!

rkeithhill:
I'm looking at this - trying to under the debuggers section of the package.json format.  Ideally, I would prefer that users did not have to change a user preference setting to select the host.  It would be cool if it were something that could be selected via the Debug target dropdown in the VSCode Debug viewlet's toolbar.  But I'm not quite seeing how to do that with the package.json file.  I created two different debuggers: entries (it's set up to take an array) but when you click the gear icon in the Debugger toolbar to create a new launch.json file, PowerShell no longer shows up in the list.  So while the format seems to indicate it will take an array of debugger configurations, it isn't working for me (or more likely I'm doing something wrong).

Anyway, I filed on issue on the VSCode repo. https://github.com/Microsoft/vscode/issues/1696

rkeithhill:
BTW what does it look like to "launch" a remote host debug adapter?  Do we need to be able to specify x86 for that as well? I would think so.

daviwil:
[...]

PR #125, opened 2016-02-02
PSES side changes for issue #96 - support x86/x64 debugging.

rkeithhill:
Besides **all** the awesome work @adamdriscoll did to enable this; there are some changes that need to be made to package.json in the vscode repo.  In order to support those changes, we need a different filename for the x86 batch file that launches the x86 debug host.


rkeithhill:
@daviwil - is there somewhere this renamed file needs to be added to ensure it is part of the package and installed into the extension's bin dir?

rkeithhill:
I'm not sure the VSCode tweak is going to make it into the next release.  I'm going to hold off on the VSCode change until I can verify the VSCode fix is in place.  But I think this PR is still good to merge (shouldn't hurt anything even if it isn't being used yet).

daviwil:
Yeah, I was thinking about the packaging issue recently...  I think it makes sense for there to be a single PowerShellEditorServices.Host NuGet package which contains both host EXEs and respective cmd files.  Right now I think the only way to accomplish that will be to change the `src\PowerShellEditorServices.Host\PowerShellEditorServices.Host.nuspec` file to include all of the necessary files so that they're packaged up together.  

I believe this requires adding a `files` section under the root `package` node after `metadata`:

```xml
<files>
  <file src="bin\$configuration$\$id$.exe" target="lib\net45\" />
  <file src="bin\$configuration$\$id$.DebugAdapter.cmd" target="content\" />
  <file src="..\PowerShellEditorServices.Host.x86\bin\$configuration$\$id$.x86.exe" target="lib\net45\" />
  <file src="..\PowerShellEditorServices.Host.x86\bin\$configuration$\$id$.DebugAdapter.x86.cmd" target="content\" />
</files>
```

More info: http://docs.nuget.org/Create/Nuspec-Reference

You should be able to test this by running this command (note the .csproj at the end):

`.\.nuget\NuGet.exe pack src\PowerShellEditorServices.Host\PowerShellEditorServices.Host.csproj`

I actually just tried to replicate this and it worked fine for me.  Both DLLs and CMD files show up in the resulting NuGet package for me.  If you make this change, could you also delete the .nuspec file in `src\PowerShellEditorServices.Host.x86` so we don't end up with an extra NuGet package?

rkeithhill:
Is this NuGet package what gets pushed to the VSCode gallery as the PSES extension?

daviwil:
Nah, it's one step earlier in the chain where I publish NuGet packages of the binaries to NuGet.org.  I'm going to start using `Install-Package` to yank those from NuGet to package up the VS Code extension.

rkeithhill:
OK, I'll commit to this branch so it all contained.  Might not get to this until tomorrow.

daviwil:
No problem, thanks a lot for doing it!

rkeithhill:
Man, that Platform=X64 environment variable bit me again.  Nuget.exe failed until I removed that env var from my console.  Anyway, the update to this PR for the requested nuspec changes have been pushed.

daviwil:
Awesome, thanks for making the nuspec changes!  Merged.

Issue #123, opened 2016-02-02
Breakpoints not honored in file name with square brackets in name


rkeithhill:
In testing the `${file}` way of specifying the file to debug, I'm noticing that for a file named `foo[1].ps1` the breakpoint never hits - even though the script clearly executes.  And while VSCode shows the file as having a breakpoint, if I run Get-PSBreakpoint in the dbg repl, it doesn't show up.  My guess is that were we set the breakpoint file we probably need to use `-LiteralPath` instead of `-Path`.

daviwil:
Hah, great catch!

rkeithhill:
@daviwil Check this out.  Create a file in ISE - name it `foo[1].ps1`.  Now put a few lines of script in it.  Put the cursor on a valid line of script and press F9.  Nothing!! No breakpoint.  Turns out that `Set-PSBreakpoint` doesn't have a Path/LiteralPath set of parameters (bad, bad cmdlet).  For files it has a `-Script` parameter that gets tripped up by globbing chars.  If you try to set the breakpoint from ISE's command window, it smartly escapes the square brackets and puts the path in a string:
```
5> Set-PSBreakpoint -Script 'C:\Users\Keith\PowerShellSupport\foo`[1`].ps1' -Line 2
```
From the command line, only this combination seems to work (path in quotes and escaped brackets). Blech.  The `CompletionText` for that path has the correctly quoted/escaped string.  I wonder if we could somehow use that for processing file paths before passing them to Set-PSBreakpoint? Hmm...

daviwil:
Haha, using IntelliSense to resolve the path might be a little overkill, but we should definitely quote and escape any path that gets passed in.  If you want to try the IntelliSense approach just to see how the code looks, go for it!

rkeithhill:
OK fixing it the right way should be easy. I believe the only wildcard chars we need to escape are `[` and `]`.

daviwil:
Sounds good to me!

PR #126, opened 2016-02-03
Fix for issue #123, wildcard chars in script path prevent breakpoints…

rkeithhill:
… from working.

Once I got breakpoints to be able to set for files like foo[1].ps1 and foo][.ps1 I noticed that the script named foo][.ps1 would not execute at all!! You get an error saying something like the specified wildcard character pattern is not valid.  FWIW ISE fails with this same message.  So PSES has one up on ISE.  :-)

### DIFF
--- a/src/PowerShellEditorServices/Debugging/DebugService.cs
+++ b/src/PowerShellEditorServices/Debugging/DebugService.cs
@@ -57,6 +57,16 @@ namespace Microsoft.PowerShell.EditorServices
         #region Public Methods
 
         /// <summary>
+        /// Returns the passed in path with the [ and ] wildcard characters escaped.
+        /// </summary>
+        /// <param name="path">The path to process.</param>
+        /// <returns>The path with [ and ] escaped.</returns>
+        public static string EscapeWilcardsInPath(string path)
+        {
+            return path.Replace("[", "`[").Replace("]", "`]");
+        }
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -57,6 +57,16 @@ public DebugService(PowerShellContext powerShellContext)
         #region Public Methods
 
         /// <summary>
+        /// Returns the passed in path with the [ and ] wildcard characters escaped.
+        /// </summary>
+        /// <param name="path">The path to process.</param>
+        /// <returns>The path with [ and ] escaped.</returns>
+        public static string EscapeWilcardsInPath(string path)
```
daviwil: Not a big deal, but slight typo here: EscapeWil**d**cardsInPath

daviwil: Also I'd say this method probably belongs in PowerShellContext rather than DebugService since it may have more general utility

rkeithhill:
Thanks for catching that typo.  I'll fix that.  Yeah, I wondered where to put that static and dithered between PowerShellContext and DebugService.  :-)

daviwil:
Awesome, thanks!  Merged.

PR #134, opened 2016-02-05
Fix for issue 129 - increase default dbg console to 120

rkeithhill:
I'll also update the MaxWindowSize and MaxPhysicalWindowSize (including height) to all be consistent with the initial default size.  This is all we can do for now as we get no notifications from VSCode on console size changes.


daviwil:
Great!  Merged.

PR #136, opened 2016-02-06
Add intervallic buffering of OutputEvent messages

### LINKED ISSUE #113: Improve performance of console output by buffering OutputEvent messages
Console output is exposed via the OutputEvent in both the LanguageServer and DebugAdapter.  Currently we write an OutputEvent message for each line of output that gets sent to our PSHostUserInterface implementation.  We should buffer these writes within some short time window to reduce the amount of OutputEvent messages that get sent to the client.  This should result in a decent performance boost when commands with many lines of output are executed in the REPL.

daviwil:
This change introduces a new AsyncDebouncer class which is able to gate
invocation requests to only occur after a specified interval.  This class
is used to implement an OutputDebouncer which takes all output messages
written from the PowerShell host and buffers them.  This causes many lines
of output written in a short time window to be sent as one OutputEvent
rather than writing individual events for each line.

Resolves #113.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/PowerShellEditorServices.Protocol.csproj
+++ b/src/PowerShellEditorServices.Protocol/PowerShellEditorServices.Protocol.csproj
@@ -124,6 +124,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="LanguageServer\References.cs" />
     <Compile Include="Server\LanguageServerSettings.cs" />
+    <Compile Include="Server\OutputDebouncer.cs" />
     <Compile Include="Server\PromptHandlers.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/src/PowerShellEditorServices.Protocol/Properties/AssemblyInfo.cs
+++ b/src/PowerShellEditorServices.Protocol/Properties/AssemblyInfo.cs
@@ -41,3 +41,4 @@ using System.Runtime.InteropServices;
 [assembly: AssemblyFileVersion("0.0.0.0")]
 [assembly: AssemblyInformationalVersion("0.0.0.0")]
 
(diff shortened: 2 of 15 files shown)


rkeithhill:
Cool.  Where/when did you learn all this async/await Fu?  I've only seriously started using it in the past 5 months on a WPF app.

daviwil:
I learned a lot from the [Concurrency in C# Cookbook](http://www.amazon.com/Concurrency-C-Cookbook-Stephen-Cleary-ebook/dp/B00KCY2CB4/ref=mt_kindle?_encoding=UTF8&me=).  I had been using async/await for a while but only really started learning the nitty-gritty last year when I started this project.  It's actually a pretty good model and can do quite a lot once you understand all the uses of Task<T>.

dfinke:
Hmm, cool, didn't know about that book. I had Joe Duffy's work on my list. Thanks.

rkeithhill:
Thanks for the tip.  That book will be on order from Amazon soon.  :-)

daviwil:
It's pretty good!  Covers the TPL Dataflow library (which I haven't used yet) and Reactive Extensions (which I'd love to use but would require extra open-source approval :/)

daviwil:
Argh, this change is really biting me right now.  Makes LanguageServer tests much more affected by output timing.  Trying to figure out a predictable way of gathering output for comparison.

rkeithhill:
Can you bypass the delay for testing purposes?

daviwil:
Maybe, but I'd rather have the output be as real as possible just in case anything breaks.  I've got a test fix in progress now, we'll see if it works out.  If not, I may work around it for now.

daviwil:
I've got a much cleaner solution for the tests now, was also able to fix a couple issues thanks to Keith's comments elsewhere.

daviwil:
Argh, looks like the solution wasn't totally bulletproof, the CI tests on master just failed.  Probably a timing issue.  I'm looking into it.

PR #139, opened 2016-02-07
Fixes #137 - working dir should be set to something reasonable.

### LINKED ISSUE #137: Debugger should set working dir to something other than VSCode install dir
I would propose that we set the current working dir to either the workspace root dir or the folder of the initial script file.  I've already run into issues with relative paths that stumped me for a bit until I realized the working dir wasn't set to what what I thought it was.  Do you think this could make it into 0.4.0?  It would help the debug experience.

rkeithhill:
Turns out vscode passes us the cwd.  Just needed to rename the property in the LaunchRequest class to match. Same for args (renamed Arguments).  I have not hooked up script args yet though.

@daviwil - please look over the ExecuteCommand for Set-Location to make sure I'm doing the ContinueWith() correctly.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
@@ -33,10 +33,10 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
         public string RuntimeExecutable { get; set; }
 
     //        /** Optional arguments passed to the runtime executable. */
-        public string[] RuntimeArguments { get; set; }
+        public string[] RuntimeArgs { get; set; }
 
     //        /** Optional environment variables to pass to the debuggee. The string valued properties of the 'environmentVariables' are used as key/value pairs. */
-        public Dictionary<string, string> EnvironmentVariables { get; set; }
+        public Dictionary<string, string> Env { get; set; }
     }
 }
 
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
@@ -24,10 +24,10 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
         public bool StopOnEntry { get; set; }
 
     //        /** Optional arguments passed to the debuggee. */
-        public string[] Arguments { get; set; }
+        public string[] Args { get; set; }
 
     //        /** Launch the debuggee in this working directory (specified as an absolute path). If omitted the debuggee is lauched in its own directory. */
-        public string WorkingDirectory { get; set; }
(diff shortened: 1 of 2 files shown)


rkeithhill:
BTW if this goes through I will be updating the package.json to add configurationProperties for cwd and maybe args (if I can figure out how to pass those args to the script reliably).

daviwil:
Awesome!  I think you can simplify your code a little bit, will leave a comment in the commit

src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
```diff
@@ -24,10 +24,10 @@ public class LaunchRequestArguments
         public bool StopOnEntry { get; set; }
 
     //        /** Optional arguments passed to the debuggee. */
-        public string[] Arguments { get; set; }
+        public string[] Args { get; set; }
 
     //        /** Launch the debuggee in this working directory (specified as an absolute path). If omitted the debuggee is lauched in its own directory. */
-        public string WorkingDirectory { get; set; }
+        public string Cwd { get; set; }
```
daviwil: Apparently they changed this on me a long time ago and I never noticed it...

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -74,23 +74,37 @@ protected override void Shutdown()
...
+            setWorkingDirCommand.AddCommand(@"Microsoft.PowerShell.Management\Set-Location")
+                .AddParameter("LiteralPath", workingDir);
+
             // Execute the given PowerShell script and send the response.
             // Note that we aren't waiting for execution to complete here
             // because the debugger could stop while the script executes.
             Task executeTask =
                 editorSession.PowerShellContext
-                    .ExecuteScriptAtPath(launchParams.Program)
+                    .ExecuteCommand(setWorkingDirCommand)
```
daviwil: Instead of doing the ExecuteCommand here, I'd say do it in the block above with an await:

```csharp
            string workingDir = launchParams.Cwd ?? Path.GetDirectoryName(launchParams.Program);
            var setWorkingDirCommand = new PSCommand();
            setWorkingDirCommand.AddCommand(@"Microsoft.PowerShell.Management\Set-Location")
                 .AddParameter("LiteralPath", workingDir);

            await editorSession.PowerShellContext.ExecuteCommand(setWorkingDirCommand);
            
            Logger.Write(LogLevel.Verbose, "Working dir set to '" + workingDir + "'");
```

Using await adds some additional error handling for us that we don't get with ContinueWith. If the ExecuteCommand fails, the ContinueWith needs to check for the error and deal with it somehow.  Better to let the framework do that in this case I think.  Makes the code a bit simpler too.

rkeithhill:
OK, updated that method as you suggested.

daviwil:
Fantastic, thanks!  Merged it.

PR #140, opened 2016-02-07
Adding support for passing ${file} as cwd.

rkeithhill:
If cwd is a file path we check for that and if so, we grab its folder.  If any error occurs during this test for dir/grab parent dir operation we log it and continue.  I don't think we should crash the debugger.  Unfortunately the call to set-location with a bad path doens't get echoed to the debug console (just the debugAdapter.log) so the user won't know they typo'd a path - if they change the default.  :-(

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -88,6 +88,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             catch (Exception ex)
             {
                 Logger.Write(LogLevel.Error, "cwd path is bad: " + ex.Message);
+                workingDir = Environment.CurrentDirectory;
             }
 
             var setWorkingDirCommand = new PSCommand();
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -74,10 +74,22 @@ protected override void Shutdown()
...
+            try
+            {
+                if ((File.GetAttributes(workingDir) & FileAttributes.Directory) != FileAttributes.Directory)
+                {
+                    workingDir = Path.GetDirectoryName(workingDir);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Write(LogLevel.Error, "cwd path is bad: " + ex.Message);
```
daviwil: I just realized that maybe we should set a default path in the catch in case the path is actually bad, that way we don't try to Set-Location on a full file path or something.

rkeithhill: Good idea.  Give me a sec.

rkeithhill: How about?
```csharp
catch (Exception ex)
{
    Logger.Write(LogLevel.Error, "cwd path is bad: " + ex.Message);
    workingDir = Environment.CurrentDirectory;
}
```
I thought about doing a `Path.GetDirectoryName(launchParams.Program)` but they could have specified a bad path to a script file as well and that could throw.  In this case, Set-Location gets a valid path and when we try to execute the script - hopefully they'll get a good error in the dbg console.

daviwil: Yep, that sounds good to me.

daviwil:
Looks good, thanks man!  Merged.

Issue #141, opened 2016-02-07
Add support for running scripts with arguments


rkeithhill:
I have a prototype.  I think this is pretty doable.

daviwil:
Awesome!

PR #142, opened 2016-02-07
Fixex #141 - passes args to script.

rkeithhill:
This approach concats args (if any passed) with a space delimiter and then appends that to the scriptpath (space separed) and used AddScript() which takes this combination of script and args.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -97,14 +97,22 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 
             await editorSession.PowerShellContext.ExecuteCommand(setWorkingDirCommand);
 
-            Logger.Write(LogLevel.Verbose, "Working dir set to '" + workingDir + "'");
+            Logger.Write(LogLevel.Verbose, "Working dir set to: " + workingDir);
+
+            // Prepare arguments to the script - if specified
+            string arguments = null;
+            if ((launchParams.Args != null) && (launchParams.Args.Length > 0))
+            {
+                arguments = string.Join(" ", launchParams.Args);
+                Logger.Write(LogLevel.Verbose, "Script arguments are: " + arguments);
+            }
 
             // Execute the given PowerShell script and send the response.
             // Note that we aren't waiting for execution to complete here
             // because the debugger could stop while the script executes.
             Task executeTask =
                 editorSession.PowerShellContext
-                    .ExecuteScriptAtPath(launchParams.Program)
+                    .ExecuteScriptAtPath(launchParams.Program, arguments)
                     .ContinueWith(
                         async (t) => {
                             Logger.Write(LogLevel.Verbose, "Execution completed, terminating...");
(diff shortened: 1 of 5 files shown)


daviwil:
Sweet!  Do you mind adding a test for this in `test\PowerShellEditorServices.Test\Debugging\DebugServiceTests.cs` so that we can verify that the behavior keeps working in the future?  You could add another test script to `test\PowerShellEditorServices.Test.Shared\Debugging` which accepts a positional param, an named parameter accepting an array, and a switch param, just to make sure we cover the useful cases.

rkeithhill:
Sure thing.

rkeithhill:
I've got a first test going but I'm not sure if the ExecuteScriptAtPath should be awaited and if there is anything I need to wait on after the abort at the end?
```csharp
[Fact]
public async Task DebuggerAcceptsScriptArgsAsSingleString()
{
    ScriptFile debugWithParamsFile =
        this.workspace.GetFile(
            @"..\..\..\PowerShellEditorServices.Test.Shared\Debugging\DebugWithParamsTest.ps1")

    await this.debugService.SetBreakpoints(
        debugWithParamsFile,
        new int[] { 3 });

    string arguments = "Foo -Param2 @('Bar','Baz') -Force Extra1";

    // Execute the script and wait for the breakpoint to be hit
    this.powerShellContext.ExecuteScriptAtPath(
        debugWithParamsFile.FilePath, arguments);

    await this.AssertDebuggerStopped(debugWithParamsFile.FilePath);

    StackFrameDetails[] stackFrames = debugService.GetStackFrames();

    VariableDetailsBase[] variables =
        debugService.GetVariables(stackFrames[0].LocalVariables.Id);

    var var = variables.FirstOrDefault(v => v.Name == "$Param1");
    Assert.NotNull(var);
    Assert.Equal("\"Foo\"", var.ValueString);
    Assert.False(var.IsExpandable);

    var = variables.FirstOrDefault(v => v.Name == "$Param2");
    Assert.NotNull(var);
    Assert.True(var.IsExpandable);
    
    var childVars = debugService.GetVariables(var.Id);
    Assert.Equal(9, childVars.Length);
    Assert.Equal("\"Bar\"", childVars[0].ValueString);
    Assert.Equal("\"Baz\"", childVars[1].ValueString);

    var = variables.FirstOrDefault(v => v.Name == "$Force");
    Assert.NotNull(var);
    Assert.Equal("True", var.ValueString);
    Assert.True(var.IsExpandable);

    var = variables.FirstOrDefault(v => v.Name == "$args");
    Assert.NotNull(var);
    Assert.True(var.IsExpandable);

    childVars = debugService.GetVariables(var.Id);
    Assert.Equal(8, childVars.Length);
    Assert.Equal("\"Extra1\"", childVars[0].ValueString);

    // Abort execution of the script
    this.powerShellContext.AbortExecution();
}
```

daviwil:
I'd say don't worry about awaiting the execute task, the debug adapter should terminate after AbortExecution is called.  You might want to keep an eye on your process list to make sure this is actually the case after the test has finished running.

The rest of the validation steps look excellent.

rkeithhill:
Hmm, looks like the CI build is running long.

daviwil:
Yeah, my output buffering change caused it, I'm working on some improvements to those tests now.

daviwil:
I see that both of your new tests pass though (nice use of Theory!) so I'll go ahead and merge this.

rkeithhill:
Heh, just learned that trick today.  I remember seeing Adam's (or your) test use InlineData for the different versions of PowerShell.

PR #143, opened 2016-02-07
Fix timing issues with OutputReader in host tests

daviwil:
This change provides a fix with timing issues that can occur with tests
which leverage the OutputReader class to read buffered output from the
host process.  It avoids test hangs by using a timeout of 5 seconds per
read and handles an edge case where some line segments don't end with
newlines.


daviwil:
Not sure whether this will resolve the root cause but hopefully it will prevent future test hangs due to any issues.

PR #144, opened 2016-02-08
Fixes issue with notepad.exe running correctly.

rkeithhill:
For some reason, the PowerShell engine thinks notepad.exe is a command-line app and calls the SessionPSHost BeginNotifyApplication() which used to throw NotImplementedException.  It now just logs the call.  

Also updated the PSHost version to pull the version from the services dll.  I tried several ways to pull from the host app (reflection Process.GetCurrentProcess().MainModule) but both caused the unit tests to hang.


daviwil:
Great!  Thanks for doing this.

PR #150, opened 2016-02-10
Fixes #149 script paths with spaces

### LINKED ISSUE #149: Scripts fail to launch if working directory path contains spaces
Example: D:\powershell code\Get-Something.ps1 will fail to launch with the following error:
```
d:\powershell : The term 'd:\powershell' is not recognized as the name of a cmdlet, function, script file, or operable 
 program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
 At line:1 char:1
•d:\powershell code\Get-OutlookPublicFolders.ps1
•~~~~~~~~~~~~~ ◦CategoryInfo : ObjectNotFound: (d:\powershell:String) [], CommandNotFoundException
◦FullyQualifiedErrorId : CommandNotFoundException
```
launch.json contains the default settings (as of 0.4.0 release of Powershell extension):
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "PowerShell",
            "type": "PowerShell",
            "request": "launch",
            "program": "${file}",
            "args": [],
            "cwd": "${file}"
        }
    ]
}
```

I have also tried placing the full path name in the cwd parameter, using escaped double quotes, single quotes, etc.

rkeithhill:
Updated the EscapePath method to optionally escape spaces.  Turns out that you don't want to do this for setting a breakpoint because it uses AddCommand instead of AddScript.  With AddCommand, the parameter value is passed as a single string so spaces don't need to be escaped.


rkeithhill:
Hold off on merging this. Need to update handle case of already escaped chars.

daviwil:
Good call!  I didn't think about that

rkeithhill:
Let's give the latest update some bake/test time.  I have to head into work but I wanted your review and I want to test this some more tonight.

daviwil:
Yep, makes sense.  In fact, feel free to merge the PR when you're happy with it :)

Here's something to keep in mind for the longer term: as you mentioned in the initial PR comments, this problem occurs because we're using AddScript instead of AddCommand now.  Lee Holmes told me a few months back that it was good that I wasn't using AddScript because of the potential for script injection attacks.  At some point I'd like to find a solution to the args-passing problem that allows us to switch back to using AddCommand for script execution.  This would also help with the space escaping issue.  Since it may not be easy to solve, we can wait a bit on making that change.

rkeithhill:
Agreed. It would be better to use AddCommand.  This particular fix may be worth putting in a 0.4.1 but for 0.5.0 it would be good to figure out a better arg passing solution.

daviwil:
Yep, definitely.  Actually we're up to 0.4.2 now because I had already shipped the Editor Services NuGet packages before we found the native command issue :)

rkeithhill:
@daviwil Can you review the `PowerShellContext.SetWorkingDirectory` method I added?  I think calling the Runspace.SessionStateProxy.Path.SetLocation() method will be better than invoking a pipeline with `Set-Location`.

daviwil:
Looks great to me! Let's go with this solution.

PR #151, opened 2016-02-10
Easier launching of the debugging when /waitForDebugger passed in.

rkeithhill:
OK been doing a lot of debugging of the debug adapater and this has been driving me crazy.   Launch VSCode, open PS1 file so I can find the PID of the editor host.  Then debug the script, hop over to VS attach to process, find the PID of the second host process (the debug instance) and attach.  This new approach will just pop up a dialog box asking which instance of VS to use to debug.  You can press NO to skip attaching the debugger and the script will continue to execute.

### DIFF
--- a/src/PowerShellEditorServices.Host/Program.cs
+++ b/src/PowerShellEditorServices.Host/Program.cs
@@ -27,15 +27,15 @@ namespace Microsoft.PowerShell.EditorServices.Host
                             "/waitForDebugger",
                             StringComparison.InvariantCultureIgnoreCase));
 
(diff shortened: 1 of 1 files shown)


rkeithhill:
This can be merged independent of the script path spaces fix for #149.

daviwil:
This is great, I like it!  Don't know why I didn't think of that before ;)  If this ends up being a smoother process going forward maybe we can rename the argument to something like `/breakOnLaunch` or something like that.  Not important for now, though.

daviwil:
Also, feel free to merge this if you're ready.

PR #152, opened 2016-02-10
Update README.md to convey latest project status

daviwil:
Also introducing the new [Development Roadmap](https://github.com/PowerShell/PowerShellEditorServices/wiki/Development-Roadmap) wiki page.  @rkeithhill, would appreciate your feedback on what I've laid out there so far.  Feel free to add new items that you're interested in working on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ across multiple editors.
   - Real-time semantic analysis of scripts using PowerShell Script Analyzer
   - Basic script evaluation
 - The Debugging Service simplifies interaction with the PowerShell debugger (breakpoints, variables, call stack, etc)
+- The Console Service provides a simplified interactive console interface which implements a rich PSHost implementation:
+  - Interactive command execution support, including basic use of native console applications
+  - Choice prompt support
+  - Input prompt support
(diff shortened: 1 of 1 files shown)


rkeithhill:
Looks good.  I was going to mention attaching the debugging to runspace (both local and remote) but I see you have that in 0.7.0.  Do you think conditional breakpoint support is something easy enough for me to do in 0.4.2 or should that wait until 0.5.0?  As for doc strategy, what do you think about using [DocFX](http://dotnet.github.io/docfx/index.html)?

daviwil:
Since conditional breakpoint would be a new feature, I'd say lets leave it for 0.5.  If you're interested in working on remote session debugging sooner than 0.7, I'm fine with that!  I had originally put it in 0.6 but wasn't sure if it was too early.  Totally up to you.  PoshTools' codebase has a pretty good example of how to accomplish this.

DocFX looks sweet, I've looked at in the past and liked what I saw.  I'll need to talk to Joey to see if we need to use do something that fits in with the new PowerShell documentation strategy. We'll have to play around with some options in the next couple months.

PR #156, opened 2016-02-15
Fixes #155 - selecting outermost scope crashes debug host.

### LINKED ISSUE #155: Debugging script using AddScript introduces an extra scope which when selected, crashes the host
Ugh.  It appears that this crashes the host because VSCode asks for "sourceReference" info with a null value which crashes during JSON deserialization.  We could probably fix the stackFrame's source to return a non-null Name (we don't have a Path for this outer scope).  However we would like to replace AddScript with AddCommand.AddParameter/AddArgument.  And that would eliminate the extra scope!

I have prototyped this and there are several problems.  First, we would no longer accept a "single" string containing all arguments.  I don't want to get into the business of fully replicating PowerShell's argument parser.  So users will have to provide each parameter/argument in a separate string in the args string array.  I've identified several places where that breaks down e.g switch parameters:
```
"-SwitchParameter", "value"
```
Passing switch parameters is tricky.  They really need to be at the end of the args array after all other parameters & args.  This is because we don't know if `"value"` is supposed to be an argument for the `-SwitchParameter` of if it is meant to be a positional (or extra) argument.  We "could" require the user to **always** supply an argument for switches but the `AddParameter` API wants a .NET boolean value as the arg.  We only get strings from VSCode. In the args, I've tried "$true", "1", etc but I get the error:
```
Cannot process argument transformation on parameter 'force'. Cannot convert value 
"System.String" to type "System.Management.Automation.SwitchParameter". 
Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0.
```
Keep in mind I don't really want to know if the parameter is supposed to be a switch or not so I again I don't want to get into the business of specia casing values (converting string "$true" to .NET boolean true).

The third issue is handling argument arrays e.g. `-p1 foo,bar,baz`.  How does the user specify that in the `"args":` field in launch.json.  Who knows what else will crop up.

rkeithhill:
Because we use AddScript, that adds another scope not associated with any filename.  That resulted in a StackFrame ScriptPath with a null value.  Later, VSCode sent us a request for soruce with a sourceReference value of null.  That caused the JSON deserialization of that message to throw and that crashed the debug host.  I made two changes.  First, when we construct the StackFrameDetails, if the PowerShell CallStackFrame.ScriptName is null, we use the string "<No File>" instead for ScriptPath.  The second change was to ExecuteScriptAtPath.  Essentially, if you do not pass args we now use AddCommand(scriptPath) which does not add the extra stack frame.  If you do use args then we still use AddScript.  You will see the extra outermost stack frame of "<ScriptBlock>, <NoFile>".  What is somewhat useful about this is that you can open Script variables to look at $MyInvocation.Line to see how the script was invoked with the args.

### DIFF
--- a/src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
+++ b/src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
@@ -64,7 +64,7 @@ namespace Microsoft.PowerShell.EditorServices
         {
             return new StackFrameDetails
             {
-                ScriptPath = callStackFrame.ScriptName,
+                ScriptPath = callStackFrame.ScriptName ?? "<No File>",
                 FunctionName = callStackFrame.FunctionName,
                 LineNumber = callStackFrame.Position.StartLineNumber,
                 ColumnNumber = callStackFrame.Position.StartColumnNumber,
--- a/src/PowerShellEditorServices/Session/PowerShellContext.cs
+++ b/src/PowerShellEditorServices/Session/PowerShellContext.cs
@@ -455,18 +455,22 @@ namespace Microsoft.PowerShell.EditorServices
         /// <returns>A Task that can be awaited for completion.</returns>
         public async Task ExecuteScriptAtPath(string scriptPath, string arguments = null)
         {
-            // If we don't escape wildcard characters in the script path, the script can
-            // fail to execute if say the script name was foo][.ps1.
-            // Related to issue #123.
-            string escapedScriptPath = EscapePath(scriptPath, escapeSpaces: true);
+            PSCommand command = new PSCommand();
 
             if (arguments != null)
             {
-                escapedScriptPath += " " + arguments;
-            }
+                // If we don't escape wildcard characters in the script path, the script can
+                // fail to execute if say the script name was foo][.ps1.
+                // Related to issue #123.
+                string escapedScriptPath = EscapePath(scriptPath, escapeSpaces: true);
+                string scriptWithArgs = escapedScriptPath + " " + arguments;
 
-            PSCommand command = new PSCommand();
-            command.AddScript(escapedScriptPath);
+                command.AddScript(scriptWithArgs);
+            }
+            else
+            {
+                command.AddCommand(scriptPath);
+            }
 
             await this.ExecuteCommand<object>(command, true);
         }


daviwil:
Looks great!  Had to restart the CI build because of a test hang :/  You can merge it anytime though.

daviwil:
The test hang is an old recurring one, not to imply that your change caused it ;)

daviwil:
Argh, another hang.  Bad luck on Appveyor today!

rkeithhill:
So if I debug that test ServiceExecutesReplCommandAndReceivesInputPrompt it passes.  If I just run it, it fails.  A timing thing?

rkeithhill:
Hmm, if I debug the unit test printf style, it seems the output is off by one:
pses-unittestfail
The `Key Value` should be in index 1 not two.

daviwil:
Yeah, there's a funky timing issue with that one for some reason...  There's some kind of bug in OutputDebouncer but I haven't been able to find a consistent repro to help diagnose a fix.  Only appears on AppVeyor for me.  Strange thing is that I put a hack in that test to avoid it and it still seems to get hit for you.

rkeithhill:
In my case, the extra newline is appearing between `Name: John` and `Key Value` which the test hack doesn't invoke in my case because the extra line is not before `Name: John`.

rkeithhill:
Does this printf (action breakpoint) output tell you anything:
```
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) read output "\r\n                $NameField = New-Object System.Management.Automation.Host.FieldDescription \"Name\"\r\n                $NameField.SetParameterType([System.String])\r\n                $fields = [System.Management.Automation.Host.FieldDescription[]]($NameField)\r\n                $host.ui.Prompt($null, $null, $fields)\r\n\r\nName: John\r\n"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) nextOutputString is ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) foud new line? true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "                $NameField = New-Object System.Management.Automation.Host.FieldDescription \"Name\""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "                $NameField.SetParameterType([System.String])"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "                $fields = [System.Management.Automation.Host.FieldDescription[]]($NameField)"
Lock Dispose 14696841
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "                $host.ui.Prompt($null, $null, $fields)"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "Name: John"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "                $NameField = New-Object System.Management.Automation.Host.FieldDescription \"Name\""  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "                $NameField.SetParameterType([System.String])"  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "                $fields = [System.Management.Automation.Host.FieldDescription[]]($NameField)"  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "                $host.ui.Prompt($null, $null, $fields)"  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  ""  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "Name: John"  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  ""  /  lineHasNewLine true
Lock Dispose 14696841
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) read output "\r\nKey  Value\r\n---  -----\r\nName John \r\n\r\n\r\n\r\nPS C:\\Users\\Keith\\GitHub\\rkeithhill\\PowerShellEditorServices\\test\\PowerShellEditorServices.Test.Host\\bin\\Debug\r\n"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) nextOutputString is ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) foud new line? true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "Key  Value"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "---  -----"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "Name John "
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: "PS C:\\Users\\Keith\\GitHub\\rkeithhill\\PowerShellEditorServices\\test\\PowerShellEditorServices.Test.Host\\bin\\Debug"
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) enqueuing line: ""
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "Key  Value"  /  lineHasNewLine true
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  "---  -----"  /  lineHasNewLine true
```

daviwil:
Awesome!  I was just about to do something like that, thanks for the log.  I'll analyze this for a bit and let you know.

After some thought, I think this is a bug in the OutputReader that I'm using to read buffered output. I can see where there might be a cause that an empty string is being returned when no output is ready yet.  Your log should help me come up with the reason why

daviwil:
This line seems wrong to me:

```
Microsoft.PowerShell.EditorServices.Test.Host.OutputReader.ReadLine(string) buffered nextOutputString:  ""  /  lineHasNewLine true
```

I think the logic for lineHasNewLine is incorrectly marking the last item as a newline when the final segment is an empty string.  Let me see if I can come up with a fix

daviwil:
And I think this is only when the lines that are read are being enqueued into the buffer.

daviwil:
I think I have a fix for this, will submit as a separate PR

daviwil:
Lets see if this fixes it: https://github.com/PowerShell/PowerShellEditorServices/pull/158

rkeithhill:
Hmm, is this hanging again?  The unit tests did all pass before I pushed the merged changes from PR #158.

daviwil:
Damn...  Hanging on the same test.  This is really annoying.

daviwil:
The unfortunate thing is that there's no reliable way to debug these issues on the CI server.  The best I can usually do is enable RDP and then remote desktop into the machine and try to read the logs.  I'll take a look at it some more tomorrow.  You can go ahead and merge the PR if you like or you can keep it open so we can try any new bug fixes that I figure out.

rkeithhill:
FWIW unit tests pass much more consistently now on my dev PC.  Occasionally the tests hang and when I attach the debugger, I see that Shutdown is waiting for output to flush on one thread:
```
 	mscorlib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 662	C#
 	mscorlib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 3320	C#
 	mscorlib.dll!System.Threading.Tasks.Task.InternalWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 3252	C#
 	mscorlib.dll!System.Threading.Tasks.Task.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 3167	C#
 	mscorlib.dll!System.Threading.Tasks.Task.Wait() Line 3054	C#
>	Microsoft.PowerShell.EditorServices.Protocol.dll!Microsoft.PowerShell.EditorServices.Protocol.Server.LanguageServer.Shutdown() Line 86	C#
...
```

daviwil:
Jeez, even with a CPU load generator I can't make these issues repro on my machine.  Seems like I need to specify a timeout there at the very least, but I need to check and see why it would be hanging anyway.  Might be that the client's stream has disconnected before the final flush can occur so the SendEvent method in OutputDebouncer.OnFlush can't complete.  That should throw an error, though...

If you're able to reproduce that issue locally again sometime today, could you email me a dump of the host process?  It'd be helpful to see what state everything is in at that point.

daviwil:
Thanks to the dump you sent I was able to reproduce and fix the latest hang issue.  Sync, merge, and try again if you don't mind :)

rkeithhill:
Yay!  The tests passed with no hangs!

daviwil:
That's awesome. Thanks a lot for the debugging help!

rkeithhill:
No problem.  Hopefully that's the end of the unit test hangs (crossing my fingers).  :-)

daviwil:
Haha, I'm sure I'll find a way to create some more soon ;)

PR #168, opened 2016-02-25
Fixes #164 - adds capabilities response to initialize request.

### LINKED ISSUE #164: Add capabilities data to InitializeResponse body in debug adapter
The VS Code team has added a capabilities list that our debug adapter can opt into:

https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/parts/debug/common/debugProtocol.d.ts#L495

In the following issue, Isidor recommended that we set the `supportsEvaluateForHovers` to true so that data tips are evaluated correctly.  We should also determine which other capabilities we'd like to support and file separate issues for enabling those.

Related: https://github.com/Microsoft/vscode/issues/1804

rkeithhill:
This commit is in preparation for adding conditional breakpoint and function breakpoint support.  Also cleaned up some doc comments.  The two primary files that were changed functionally are InitializeRequest.cs and DebugAdapaterBase.cs.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
@@ -3,8 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
 using System.Collections.Generic;
+using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
 {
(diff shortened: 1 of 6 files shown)


src/PowerShellEditorServices.Protocol/DebugAdapter/LaunchRequest.cs
```diff
@@ -17,25 +17,43 @@ public static readonly
 
     public class LaunchRequestArguments
     {
-    //        /** An absolute path to the program to debug. */
```
daviwil: Thanks for fixing those!

daviwil:
Damn, didn't see the notification for this last night, sorry about that.  Looks great!

rkeithhill:
Thanks. I want to start looking at both conditional and function breakpoints next.  Given PowerShell's support for this, I think we can get also get VS style `action` breakpoints.  :-)

daviwil:
That will be awesome!

PR #169, opened 2016-02-25
Partial fix for issue 95 - set all scopes but Local as expensive.

rkeithhill:
This makes the debug hover tips work better.  But if you want to look at a variable in a higher scope, you have to select that scope in the callstack so the variable is a "local" variable in that scope.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/Scope.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/Scope.cs
@@ -29,6 +29,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
             return new Scope {
                 Name = scope.Name,
                 VariablesReference = scope.Id,
+                // Temporary fix for #95 to get debug hover tips to work well at least for the local scope.
                 Expensive = (scope.Name != VariableContainerDetails.LocalScopeName)
             };
         }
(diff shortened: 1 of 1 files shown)


daviwil:
Looks good!

daviwil:
By the way, I was thinking about releasing an 0.4.3 of Editor Services for a couple small fixes I need for initial ISE Preview integration.  I was going to avoid shipping a new update of the VS Code extension with this update since it wouldn't be relevant.

Would you like this change to be pulled into 0.4.3 even though we probably won't update VS Code yet, or would it be alright to put this in Editor Services 0.5.0?

rkeithhill:
If you're not releasing an update via the VSCode marketplace, then we might as well wait until 0.5.0 to release it.

rkeithhill:
Um yeah, so it just occurs to me now that you might have been waiting on accepting this PR until after you create 0.4.3.   If so, sorry.  I don't think there should be any issue with this being in 0.4.3.  BTW are we using the develop branch?  Should I be creating PRs to develop instead of master?

daviwil:
Nah you did the right thing.  I had already created a release/0.4.3 branch to take fixes for that release.  Though I guess the question now becomes whether release tags work against non-master branches.  There hasn't been a reason to use the develop branch before, but if it turns out to be a pain to have tags on branches other than master, we may start using develop.

rkeithhill:
Sounds good.

PR #172, opened 2016-02-28
Fixes #94 - adds support for conditional breakpoints

### LINKED ISSUE #94: Add support for conditional breakpoints
The VS Code debug adapter protocol just gained support for conditional breakpoints. We should adopt the protocol changes and add this support in our debug service.  See the commit related to this issue for the protocol changes:

https://github.com/Microsoft/vscode-debugadapter-node/issues/6

rkeithhill:
This adds some unit tests for conditional breakpoints and refactors some existing unit tests I wrote to be a bit more DRY.  Also, VSCode supports column breakpoints in its protocol.  And PowerShell supports breakpoints on columns other than 1.  So why don't I see any UI in VSCode to create a breakpoint at a specific column?  Is there something we have to do to declare support for column breakpoints?

### DIFF
--- a/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
+++ b/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
@@ -19,6 +19,7 @@ namespace Microsoft.PowerShell.EditorServices.Test.Debugging
         private Workspace workspace;
         private DebugService debugService;
         private ScriptFile debugScriptFile;
+        private ScriptFile variableScriptFile;
         private PowerShellContext powerShellContext;
         private SynchronizationContext runnerContext;
 
@@ -35,6 +36,10 @@ namespace Microsoft.PowerShell.EditorServices.Test.Debugging
             this.debugScriptFile =
                 this.workspace.GetFile(
                     @"..\..\..\PowerShellEditorServices.Test.Shared\Debugging\DebugTest.ps1");
+
+            this.variableScriptFile =
+                this.workspace.GetFile(
+                    @"..\..\..\PowerShellEditorServices.Test.Shared\Debugging\VariableTest.ps1");
 
             this.powerShellContext = new PowerShellContext();
             this.powerShellContext.SessionStateChanged += powerShellContext_SessionStateChanged;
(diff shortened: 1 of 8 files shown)


test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -195,6 +207,79 @@ public static IEnumerable<object[]> DebuggerAcceptsScriptArgsTestData
...
+
+            Assert.Equal(2, breakpoints.Length);
+            Assert.Equal(5, breakpoints[0].LineNumber);
+            Assert.True(breakpoints[0].Verified);
+            Assert.Null(breakpoints[0].Message);
+
+            Assert.Equal(10, breakpoints[1].LineNumber);
+            Assert.False(breakpoints[1].Verified);
+            Assert.NotNull(breakpoints[1].Message);
+            Assert.Contains("Unexpected token '-ez'", breakpoints[1].Message);
```
daviwil: This is awesome!

daviwil:
This is great work man, I'm really excited for this feature to come online!  If you're ready you can merge it.

daviwil:
Regarding column breakpoints, I haven't seen anything in the UI for that yet.  Was sort of expecting that I could put the cursor somewhere, right click and Add Breakpoint but it's not there.  We might have to ask about that.

rkeithhill:
It is weird that the protocol supports column breakpoints but the UI doesn't.  Oh well.  At least we will have conditional breakpoints. BTW I ran into a "gotcha" - ScriptBlock parser will catch a number of parse errors but it won't catch this `$i == 3`.  It parses as an assignment with $i on the LHS and, get this, `= 3` on the RHS.  So yeah, in PowerShell you can actually create a command named `=` e.g.:
```powershell
function = { Write-Warning "What do you think this is - C#?" }
$i == 1
```
Outputs:
```
WARNING: What do you think this is - C#?
```
This is a bummer because somebody will invariably forget to use `-eq` and the breakpoint won't hit. I guess I could try to handle two "special" cases e.g.`LHS == RHS` and `LHS > RHS` by inspecting the AST and generating an error message? The other combinations of `!=`, `>=`, `<` and `<=` generate parse errors.  What do you think?

daviwil:
Yeah, if there's a way to catch that and throw an error (especially suggesting the usage of `-eq` and `-gt`) then that would be ideal.  This is a weird case where I'd expect the language would throw an error since it's kind of a common issue for people coming over from C#.  I know I've hit it in the past...

rkeithhill:
Ask and you shall receive:
vscodebadcondmessage1
vscodebadcondmessage2

daviwil:
That looks great!

PR #178, opened 2016-03-02
This is a *proposed* change to address the issue that the launch requ…

rkeithhill:
…est could come after the configurationDone request (https://github.com/Microsoft/vscode/issues/3548).

This commit "should" allow the messages to come in either order and still work.  **However, it needs a good review. ** BTW it appears all the messages are processed on the same thread (or maybe that was coincidence).  If that is the case then perhaps the lock is not needed.  Still it is a fairly lightweight lock - one bool check and an assignment.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -106,7 +106,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             RequestContext<object> requestContext)
         {
             // Ensure that only the second message between launch and
-            // configurationDone - actually launches the script.
+            // configurationDone requests, actually launches the script.
             lock (syncLock)
             {
                 if (!this.isLaunchRequestComplete)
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -81,6 +101,31 @@ protected override void Shutdown()
 
         #region Built-in Message Handlers
 
+        protected async Task HandleConfigurationDoneRequest(
+            object args,
+            RequestContext<object> requestContext)
+        {
+            // Ensure that only the second message between launch and
+            // configurationDone requests, actually launches the script.
+            lock (syncLock)
```
daviwil: Theoretically you shouldn't need a lock for this since all requests are handled serially by the MessageReader (a SynchronizationContext handles reads in a queue).  Did you run into a case where there was a race condition or are you just being cautious?

rkeithhill: No. Was just being overly cautious. I'll drop the locks.

daviwil: No problem!  There can be some more subtle behavior with async methods that could require locking but we can discuss that another time ;)

rkeithhill:
@daviwil Are you OK with this PR after commit f1be4b6?

daviwil:
Yeah, sorry I didn't follow up on that!  I'll merge it.

rkeithhill:
No problemo.  Thanks.

PR #179, opened 2016-03-04
Provide a bit more accurate "reason" for debug stop.

rkeithhill:
It turns out that when the debugger stops at a function breakpoint, it isn't necessarily obvious why the debugger stopped.  There is no breakpoint glyph for a function breakpoint.  So if we return "function breakpoint", then VSCode will display "PAUSED ON FUNCTION BREAKPOINT" in the call stack viewlet.  Hopefully that will help the user know why the debugger stopped.

See related VSCode issue - https://github.com/Microsoft/vscode/issues/3648

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -474,6 +474,21 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             // Flush pending output before sending the event
             await this.outputDebouncer.Flush();
 
+            // Provide the reason for why the debugger has stopped script execution.
+            // See https://github.com/Microsoft/vscode/issues/3648
+            // The reason is displayed in the breakpoints viewlet.  Some recommended reasons are: 
+            // "step", "breakpoint", "function breakpoint", "exception" and "pause".
+            // We don't support exception breakpoints and for "pause", we can't distinguish 
+            // between stepping and the user pressing the pause/break button in the debug toolbar.
+            string debuggerStoppedReason = "step";
+            if (e.Breakpoints.Count > 0)
+            {
+                debuggerStoppedReason =
+                    e.Breakpoints[0] is CommandBreakpoint
+                        ? "function breakpoint"
+                        : "breakpoint";
+            }
+
             await this.SendEvent(
                 StoppedEvent.Type,
                 new StoppedEventBody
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -474,6 +474,21 @@ protected override void Shutdown()
...
+            // The reason is displayed in the breakpoints viewlet.  Some recommended reasons are: 
+            // "step", "breakpoint", "function breakpoint", "exception" and "pause".
+            // We don't support exception breakpoints and for "pause", we can't distinguish 
+            // between stepping and the user pressing the pause/break button in the debug toolbar.
+            string debuggerStoppedReason = "step";
+            if (e.Breakpoints.Count > 0)
+            {
+                debuggerStoppedReason =
+                    e.Breakpoints[0] is CommandBreakpoint
+                        ? "function breakpoint"
```
daviwil: I like this style for ternary operators, cleaner than the one I normally use.

daviwil:
Awesome!  Looks great.

KirkMunro:
Could you please call them command breakpoints?  They aren't function breakpoints, they are command breakpoints.  They work for aliases, functions, cmdlets, and even executable applications.

daviwil:
At the protocol layer we're speaking in VS Code's terms so that we're consistent with their experience.  In the core library layer we'll definitely want to call them command breakpoints for sure.  We'll get all of our terminology straight as we start cleaning up the APIs for 1.0 later this year.

rkeithhill:
Yeah, we are trying to match the VSCode UI experience:

vscodefuncbreakpointbutton

This is one of those compromises you make when you're plugging into a multi-language editor like VSCode.  BTW this is where the reason is used:

vscodefuncbreakpointreason

When plugging in these services into ISE, ISE would use more PowerShell specific terms.

Hmm, looking at this again I see a type in the service that I do need to rename `FunctionBreakpointDetails` to `CommandBreakpointDetails`.  I'll do that later tonight.

KirkMunro:
Since VSCode is a multi-language editor, and since some of these languages may either (a) not have functions/have functions called something else (subroutine, procedure, etc), or (b) not have function breakpoints/have function breakpoints that are called something else, maybe a suggestion should be logged against VSCode that the terminology should change to command breakpoint in order to serve all languages better.

rkeithhill:
Or perhaps VSCode could allow the extension to define a preferred alias for `function breakpoints` which it would honor in the UI (primarily the tooltip you get when you hover over the Breakpoints "+" button)?  That could get tricky because VSCode is pushing for localization and I don't think the PowerShell extension is "there" yet.

daviwil:
Localization is something I'm not looking forward to...

PR #180, opened 2016-03-05
Renamed FunctionBreakpointDetails to CommandBreakpointDetails

rkeithhill:
The services API should be oriented more towards PowerShell (and not VSCode) terminology.


daviwil:
Hmmm did the tests hang the first time?

rkeithhill:
No, failed on a debouncer output test I think.

daviwil:
That's the first time I've seen that one fail due to a timeout, I'll have to keep an eye on it.  Oh well, change looks good, I'm merging it.

PR #183, opened 2016-03-08
Fix #99: 'using module' relative path resolution

### LINKED ISSUE #99: Relative paths not correctly resolved by 'using module' statement
@AndrewGaspar identified this issue and filed a bug at the PowerShell/vscode-powershell repo:

https://github.com/PowerShell/vscode-powershell/issues/64

We need to provide the proper working directory to the PowerShell parser when parsing each file so that module path resolution works correctly.

daviwil:
This change fixes an issue with relative module path resolution with the
'using module' statement in more recent versions of PowerShell v5.  A new
method overload was added to the PowerShell API's Parser.ParseInput method
which allows the script's path to be specified.  By using this parameter,
the PowerShell parser resolves relative module paths correctly.

This change also required a change to how PowerShell API usage is verified
for different PowerShell versions.  Since the PowerShell v5 reference
assemblies don't (yet) contain Parser.ParseInput overload, we now have to
make a distinction between the PowerShell v5 which is specified by the
reference assemblies and the one that is currently shipped in the latest
official Windows 10 release (10586).  Once updated reference assemblies
have been shipped for PowerShell v5 we may revisit this approach.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 version: '$(core_version).{build}'
-os: Unstable
+os: WMF 5
 configuration: Release
 clone_depth: 10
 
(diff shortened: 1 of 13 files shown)


daviwil:
I'm going to verify that this works against a Windows 10 RTM machine before I merge it.

daviwil:
Seems to be working fine on Win 10 RTM with PowerShell 5.0.10240.16410

PR #184, opened 2016-03-09
Finish PowerShell v3 and v4 support

daviwil:
This change finishes our PowerShell v3 and v4 support by conditionally
changing behavior for those versions.  This works by using an interface
facade in front of any version-specific behavior so that the CLR doesn't
try to resolve method names that aren't available in a particular version.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -295,7 +295,14 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             object pauseParams,
             RequestContext<object> requestContext)
         {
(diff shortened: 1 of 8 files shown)


daviwil:
@adamdriscoll @rkeithhill 

What do you think about the name of `IVersionSpecificOperations`?  Got a better name idea?

One cool thing, Keith's conditional and command breakpoint support works in PowerShell v3 and v4 :)

adamdriscoll:
@daviwil Looks good to me! I think the interface name is fine. Very clear what it is doing. :+1: 

Awesome that the new breakpoint types still work!!

rkeithhill:
> conditional and command breakpoint support works in PowerShell v3 and v4 

I was hoping that would be the case.  :-)

PR #207, opened 2016-04-07
Fixes #206 - completion list for $env: now works.

### LINKED ISSUE #206: $ label prefix interfering with completion list for $env:
I can fix this issue by removing a "feature" I had added to present variables and parameters in the completion list with a prefix - namely `$` and `-`.  In ISE, it displays the text as returned by the AST with no prefix **but** ISE is able to use custom icons e.g.:

ise_proplist
ise_paramlist

AFAIK we can't provide custom icons in VSCode so I changed the code to "prefix" variables and parameters so they would look a bit more "PowerShelly" e.g.:

vscode_proplist
vscode_paramlist

I can remove this special code which makes completion of `$env:` behave correctly but the completion lists look a little, well, naked IMO but functional is better than pretty.  :-)

vscode_proplist2
vscode_paramlist2

rkeithhill:
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/207)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -1009,7 +1009,6 @@ function __Expand-Alias {
         {
             string detailString = null;
             string documentationString = null;
-            string labelString = completionDetails.ListItemText;
 
             if ((completionDetails.CompletionType == CompletionType.Variable) ||
                 (completionDetails.CompletionType == CompletionType.ParameterName))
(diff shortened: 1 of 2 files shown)


daviwil:
:+1:

dfinke:
@rkeithhill Did you add the reviewable button?

daviwil:
Nah, someone added it to the PowerShell GitHub org as a whole so it keeps showing up on our PRs.  I think it's extra noise on top of the PR interface so I just ignore it and don't use it.

PR #210, opened 2016-04-11
Rkeithhill/is194 ctrl f5 run

rkeithhill:
Take a good look at the DebugAdapter change.  The LaunchScript method is not async.  I "think" I'm waiting on the awaitable DisableAllBreakpoints() method correctly by using Task.WhenAll().  If not, let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/210)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -73,14 +73,6 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 
         protected Task LaunchScript(RequestContext<object> requestContext)
         {
-            // If this is a run without debugging session, disable all breakpoints.
-            if (this.noDebug)
-            {
-                Task disableBreakpointsTask = 
-                    this.editorSession.DebugService.DisableAllBreakpoints();
-                Task.WhenAll(disableBreakpointsTask);
-            }
-
             return editorSession.PowerShellContext
                     .ExecuteScriptAtPath(this.scriptPathToLaunch, this.arguments)
                     .ContinueWith(
@@ -235,9 +227,11 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                     LogLevel.Warning, 
                     $"Attempted to set breakpoints on a non-existing file: {setBreakpointsParams.Source.Path}");
 
+                string message = this.noDebug ? null : "Source does not exist, breakpoint not set.";
+
                 var srcBreakpoints = setBreakpointsParams.Breakpoints
                     .Select(srcBkpt => Protocol.DebugAdapter.Breakpoint.Create(
-                        srcBkpt, setBreakpointsParams.Source.Path, "Source does not exist, breakpoint not set."));
+                        srcBkpt, setBreakpointsParams.Source.Path, message, verified: this.noDebug));
 
                 // Return non-verified breakpoint message.
                 await requestContext.SendResult(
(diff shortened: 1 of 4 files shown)


rkeithhill:
It's our profile friend again. When I try to debug this I see the `DidChangeConfigurationNotification` event and `EvaluateRequest` message sent but when I put breakpoints in the corresponding methods, those breakpoints never hit.  Eventually the test times out on the line `outputString = await outputReader.ReadLine();`.

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -58,6 +58,20 @@ public DebugService(PowerShellContext powerShellContext)
         #region Public Methods
 
         /// <summary>
+        /// Disables all breakpoints in the runspace.
+        /// </summary>
+        /// <returns></returns>
+        public async Task DisableAllBreakpoints()
```
daviwil: Technically this shouldn't be necessary since a new process is created each time so there would be no breakpoints to disable.  Have you seen breakpoints persist across debugging sessions?

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -72,6 +73,14 @@ protected override void Initialize()
 
         protected Task LaunchScript(RequestContext<object> requestContext)
         {
+            // If this is a run without debugging session, disable all breakpoints.
+            if (this.noDebug)
+            {
+                Task disableBreakpointsTask = 
+                    this.editorSession.DebugService.DisableAllBreakpoints();
+                Task.WhenAll(disableBreakpointsTask);
```
daviwil: Task.WhenAll actually returns a Task that aggregates the await for a list of Tasks to complete.  You'd have to await it before it actually waits for completion.  The correct way to wait in this function would either be to change the function to async and await the `DisableAllBreakpoints` method or to just call `.Wait()` on the task that's returned from `DisableAllBreakpoints`.  The latter is a little more risky though because it can cause the

daviwil:
Seems like all we really have to do in "NoDebug" mode is ignore any breakpoint requests that come in.  Did you see a case where breakpoints already existed in the session by the time we launch the script?

src/PowerShellEditorServices/Debugging/DebugService.cs
rkeithhill: No I haven't seen that.  The problem is the ordering of the messages.  Right now we get the launchRequest, setBreakpoints and configurationDone.  But at one point we were getting the launchRequest at the end of this sequence (which is I the code for those two messages tolerates order).  If the launchRequest ever comes after the setBreakpoints then it is too late to "not set" them.  

I guess my confidence in the stability of the debug protocol message ordering isn't extremely high.  :-)  But at the moment, the given order I see on the insiders build suggests that we could just "not set" them.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
rkeithhill: Yeah, it would be much easier to make that LaunchScript method async but I was worried about the ripple effects of that.  :-)

src/PowerShellEditorServices/Debugging/DebugService.cs
daviwil: Here's a possible strategy.  Let's go ahead and check in a solution where we just assume the message ordering comes in correctly and that we know the value of NoDebug before the breakpoint request comes in.  If NoDebug is on, we'll just ignore any breakpoint request that comes through.

Since the 1.0 release of VS Code is supposed to land at the end of April, we can keep an eye on it and see if the debug messages continue to be sent in the right order.  We'd then release our update on or just after the 1.0 release is done.  If it turns out people see issues with running in NoDebug mode, we can release a quick bugfix update that re-introduces the breakpoint disabling code.

I think I'd rather take this strategy than have to make further changes to our code to avoid a possible message ordering problem.  What do you think?

rkeithhill: Works for me.  I 'll make the changes tonight.  BTW we are broken again in the just released 1.0.0 alpha.  If you try to start DebugTest.ps1 under the debugger with no breakpoints set, you get no output.  :-(

daviwil: Argh...  Does it look like the script is getting executed on our end?  Are they just not showing our output in the UI?

rkeithhill: Nevermind. I'm an idiot. Every time I re-install the alpha I have to remember to update the package.json package for the PS extension to use my debug bits.  Forgot to do that. :-(

daviwil: Man, I hate having to do that.  There needs to be some way to refer to a setting string for the extension in a debug config parameter...

rkeithhill: You and I both. It would be great if that path was a user setting/preference like the editor session host path and waitForDebugger, etc.

rkeithhill:
@daviwil Check out the latest update and see what you think.

daviwil:
Looks good to me!

PR #214, opened 2016-05-03
Introduce new editor extensibility API

daviwil:
This change introduces a new PowerShell-based API that allows scripts and
modules to extend the behavior of the host editor.  Scripts can either
automate the editor directly or register commands that can later be
invoked by the user.  The ExtensionService provides this behavior for any
PowerShellContext by injecting a new variable '$psEditor' into the
session.

This version of the API is very early and will be expanded upon in future
releases.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/214)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,6 +30,7 @@ install:
 - git submodule -q update --init
 
 before_build:
+- ps: if (Test-Path 'C:\Tools\NuGet3') { $nugetDir = 'C:\Tools\NuGet3' } else { $nugetDir = 'C:\Tools\NuGet' }; (New-Object Net.WebClient).DownloadFile('https://dist.nuget.org/win-x86-commandline/v3.3.0/nuget.exe', "$nugetDir\NuGet.exe")
 - nuget restore
 
 build:
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -1,0 +1,112 @@
+# PowerShell Editor Services Extensibility Model
+
+PowerShell Editor Services exposes a common extensibility model which allows
+a user to write extension code in PowerShell that works across any editor that
+uses PowerShell Editor Services.
+
+## Using Extensions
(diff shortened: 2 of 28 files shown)


daviwil:
Still need to clean up the docs I started writing and test once again to make sure everything works in VS Code.

rkeithhill:
You've been busy.  Nice PR!

daviwil:
Yeah, I might have bit off more than I could chew for one release but I think it's a solid basis for future work :)  I'll get this merged later today and then knock out a couple other issues for 0.6.0.

By the way, according to VS Code's April Iteration Plan they'll be releasing 1.1.0 next Monday. I think I'll line up our release with that so we've got a few extra days for polish. I'll have family visiting this weekend so I'll get everything I can done before EOD Wednesday and then ship on Monday. That sound OK?

dfinke:
If you'd stop gallivanting around Europe, you might be done.

Looking forward to this.

rkeithhill:
Yup.  I still haven't been able to track down that issue with single stepping in the debugger but the workaround is easy-ish - just slow down.  :-)

daviwil:
@dfinke I took my time just to annoy you ;)

@rkeithhill If it isn't something that gets hit that often I'm fine with fixing it in a later release.  I think that we've got some issues with async Task usage in the debug adapter that might be causing some sequencing issues.  When I fix all that up this problem might go away.

dfinke:
@daviwil +100 on that front :+1:

daviwil:
Taking a little extra time on this because I think I might be able to get an actual documentation site generated from the C# XML doc comments :)  [DocFX](http://dotnet.github.io/docfx/) is pretty cool!

daviwil:
Going to finish the documentation as a separate pull request.  Merging this now.

PR #217, opened 2016-05-10
Fix return value of EditorObject.RegisterCommand

daviwil:
This change fixes the EditorObject.RegisterCommand function to pass
through the return value of the ExtensionService.RegisterCommand function
so that the Register-EditorCommand cmdlet writes the correct verbose
output depending on whether the command was already registered.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/217)


daviwil:
Small issue with the last PR.

PR #229, opened 2016-05-16
Fix #221: Occasional init hang in LanguageServer

### LINKED ISSUE #221: Timing issue in LanguageServer initialization
After updating it appears PowerShellEditorServices is not properly starting.

I am getting the following error.

"No Handler found for the command: 'PowerShell.RunSelection'. An etension might be missing an activation event."

Tried restarts and a reboot. Will try reinstalling the extension next.

daviwil:
This change fixes an issue in the LanguageServer where the length of time
that it takes to evaluate the extension API script can cause the
"initialize" request to not complete.  This leaves the VS Code language
server client in an initializing state where no language features will
work for the rest of the session.  The fix is to execute the extension API
script after the "initialize" handler has been registered.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/229)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -74,11 +74,6 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 
         protected override void Initialize()
         {
-            // Initialize the extension service
-            // TODO: This should be made awaited once Initialize is async!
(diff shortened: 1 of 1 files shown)


daviwil:
This one was pretty dumb, I was executing the extension API script before registering message handlers.  This caused the handler to be registered after that script finishes.  On some machines the "initialize" event would come before the script finished execution which meant that there would never be a handler available to respond to that request.

PR #234, opened 2016-05-17
Fix #232: Language server crashes when ErrorActionPreference is "Stop"

### LINKED ISSUE #232: Language server sometimes crashes then $ErrorActionPreference = "Stop"
The language server sometimes crashes when gathering details about a command either for signature help or symbol hover when the user has set their `$ErrorActionPreference` to `Stop`.  This will require handling the `System.Management.Automation.ActionPreferenceStopException` appropriately.

Original issue: https://github.com/PowerShell/vscode-powershell/issues/187

daviwil:
This change fixes an issue where the language server can sometimes crash
when the user has configured their $ErrorActionPreference variable to
"Stop" in the language server's session.  The fix is to refactor existing
code to use a common code path that handles RuntimeExceptions
appropriately.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/234)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -528,21 +528,16 @@ function __Expand-Alias {
         {
             if (completionItem.Kind == CompletionItemKind.Function)
             {
-                RunspaceHandle runspaceHandle =
-                    await editorSession.PowerShellContext.GetRunspaceHandle();
-
                 // Get the documentation for the function
                 CommandInfo commandInfo =
-                    CommandHelpers.GetCommandInfo(
+                    await CommandHelpers.GetCommandInfo(
                         completionItem.Label,
-                        runspaceHandle.Runspace);
+                        this.editorSession.PowerShellContext);
 
                 completionItem.Documentation =
-                    CommandHelpers.GetCommandSynopsis(
+                    await CommandHelpers.GetCommandSynopsis(
                         commandInfo,
-                        runspaceHandle.Runspace);
-
-                runspaceHandle.Dispose();
+                        this.editorSession.PowerShellContext);
             }
 
             // Send back the updated CompletionItem
--- a/src/PowerShellEditorServices/Language/CommandHelpers.cs
+++ b/src/PowerShellEditorServices/Language/CommandHelpers.cs
@@ -4,12 +4,11 @@
 //
 
 using System.Linq;
+using System.Management.Automation;
+using System.Threading.Tasks;
 
 namespace Microsoft.PowerShell.EditorServices
 {
-    using System.Management.Automation;
-    using System.Management.Automation.Runspaces;
-
     /// <summary>
     /// Provides utility methods for working with PowerShell commands.
     /// </summary>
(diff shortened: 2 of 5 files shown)


src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -7,13 +7,12 @@
...
+using System.Management.Automation;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.PowerShell.EditorServices
 {
-    using System.Management.Automation;
-    using System.Management.Automation.Runspaces;
-    using System.Threading;
     /// <summary>
```
rkeithhill: Ugh, I hope that wasn't me.  I have to watch were Resharper wants to add using statements.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -528,21 +528,16 @@ RequestContext<object> requestContext
...
 
                 completionItem.Documentation =
-                    CommandHelpers.GetCommandSynopsis(
+                    await CommandHelpers.GetCommandSynopsis(
                         commandInfo,
-                        runspaceHandle.Runspace);
-
-                runspaceHandle.Dispose();
+                        this.editorSession.PowerShellContext);
             }
```
rkeithhill: Now with more async/await coolness.  :-)

src/PowerShellEditorServices/Language/LanguageService.cs
daviwil: I dunno, I catch Visual Studio doing it from time to time.  Seems like if one `using` ends up there then Visual Studio puts all future ones there!

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
daviwil: Haha, yep.  In this case I'm just trying to reuse existing logic in `PowerShellContext.ExecuteCommand` instead of using `PowerShell.Invoke` directly.  A bit of a perf hit to go through ExecuteCommand and async twice in a row, I'm thinking I might come up with my own async pipeline helper class at some point in the future.

src/PowerShellEditorServices/Language/CommandHelpers.cs
```diff
@@ -57,13 +51,12 @@ public class CommandHelpers
...
-                using (PowerShell powerShell = PowerShell.Create())
-                {
-                    powerShell.Runspace = runspace;
-                    powerShell.AddCommand("Get-Help");
-                    powerShell.AddArgument(commandInfo);
-                    helpObject = powerShell.Invoke<PSObject>().FirstOrDefault();
-                }
+                PSCommand command = new PSCommand();
+                command.AddCommand("Get-Help");
+                command.AddArgument(commandInfo);
```
rkeithhill: BTW over time should we nail these commands down to the ones that ship with PowerShell e.g. `Microsoft.PowerShell.Core\Get-Help`?  Now that the extension runs profiles, users could be pulling in modules that proxy commands like Get-Help.  Heck, PSCX does this.  This (our proxy) has caused some end user problems and we've had to back out some features.

daviwil: That's a great point, I'll go ahead and fix these right now since I'm in here

daviwil:
Argh, I realized that this actually constitutes an API change on `CommandHelpers` methods so I might not be able to include it in 0.6.2.

daviwil:
I guess I'll worry about that once we hit 1.0 ;)

PR #241, opened 2016-05-21
Need to escape regex pattern with external component.

### LINKED ISSUE #240: Completion item with regex metachars can cause editor host to crash
See VSCode issue: https://github.com/PowerShell/vscode-powershell/issues/193#issuecomment-220784291

In this case the user (@Sharpiro) had `notepad++.exe` in their path.  I was inject that text into a regex pattern without escaping it.  My bad!

rkeithhill:
LanguageServer.CreateCompletionItem was causing a editor host crash because it (er *I*) was injecting tooltip text (returned by the PS engine) into a regex pattern without escaping it.  One of the completion items was 'Notepad++.exe' which resulted in a regex parsing error.  Would like to add a test for this at some point but need to think on it.  CreateCompletionItem is private.

Fixes #240

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/241)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -1128,7 +1128,7 @@ function __Expand-Alias {
                 // from the ToolTipText - if there is any ToolTipText.
                 if (completionDetails.ToolTipText != null)
                 {
-                    // Fix for #193 - notepad++.exe in tooltip text caused regex parser to throw.
+                    // Fix for #240 - notepad++.exe in tooltip text caused regex parser to throw.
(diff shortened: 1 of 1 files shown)


daviwil:
Sweet!  That was fast, man.

daviwil:
Man, I dunno why that debugger test has started failing so frequently now.  I'll look into it soon.  Merge anyway despite the failure

rkeithhill:
That's weird because that test passed for me when I ran it before submitting the PR.  It is the ServiceLoadsProfileOnDemand test that always times out for me.

daviwil:
That test should have been fixed with one of my recent changes...  Strange that it'd time out for you still

PR #254, opened 2016-06-10
Add Editor Commands for opening Profiles

gerane:
This change adds the following built-in Editor Commands

- PowerShellEditorServices.OpenEditorProfile: Opens the current
Editor's profile. Uses $profile so should work when other Editors use
PSES.
- PowerShellEditorServices.OpenProfileList: Prompts user with a list of
 Profiles to open.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/254)

### DIFF
--- a/src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
+++ b/src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
@@ -15,7 +15,8 @@ Register-EditorCommand `
     -ScriptBlock {
         param([Microsoft.PowerShell.EditorServices.Extensions.EditorContext]$context)
         
-        $List = @('Microsoft.VSCode_profile.ps1','Microsoft.PowerShell_profile.ps1','Microsoft.PowerShellISE_profile.ps1','Profile.ps1')
+        $Current = Split-Path -Path $profile -Leaf        
+        $List = @($Current,'Microsoft.VSCode_profile.ps1','Microsoft.PowerShell_profile.ps1','Microsoft.PowerShellISE_profile.ps1','Profile.ps1') | Select-Object -Unique
         $Choices = [System.Management.Automation.Host.ChoiceDescription[]] @($List)
         $Selection = $host.ui.PromptForChoice('Please Select a Profile', '(Current User)', $choices,'0')
         $Name = $List[$Selection]
--- a/src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
+++ b/src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
@@ -1,0 +1,29 @@
+Register-EditorCommand `
+    -Name 'PowerShellEditorServices.OpenEditorProfile' `
+    -DisplayName 'Open Editor Profile' `
+    -SuppressOutput `
+    -ScriptBlock {
+        param([Microsoft.PowerShell.EditorServices.Extensions.EditorContext]$context)
+        If (!(Test-Path -Path $Profile)) { New-Item -Path $Profile -ItemType File }
+        $psEditor.Workspace.OpenFile($Profile)
+    }
+
+Register-EditorCommand `
+    -Name 'PowerShellEditorServices.OpenProfileList' `
+    -DisplayName 'Open Profile from List (Current User)' `
+    -SuppressOutput `
+    -ScriptBlock {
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+Register-EditorCommand `
+    -Name 'PowerShellEditorServices.OpenProfileList' `
+    -DisplayName 'Open Profile from List (Current User)' `
+    -SuppressOutput `
+    -ScriptBlock {
+        param([Microsoft.PowerShell.EditorServices.Extensions.EditorContext]$context)
+        
+        $List = @('Microsoft.VSCode_profile.ps1','Microsoft.PowerShell_profile.ps1','Microsoft.PowerShellISE_profile.ps1','Profile.ps1')
```
daviwil: I missed this part when I looked at this earlier, we should use the filename of the current editor's profile as the first item rather than hardcoding the VS Code option.  Makes sense to hardcode the other profiles though since they are common on Windows.

daviwil: If you want to wait and make this change later, that's alright.  If you want to make it now you can just make an additional commit in your branch and push it to your repo.  The PR should pick up the change automatically.

daviwil:
The AppVeyor failure is an occasionally recurring one, don't worry about it :)

gerane:
lol =)

src/PowerShellEditorServices/Extensions/BuiltInCommands.ps1
gerane: Ok, I will look at that. Would you rather check for duplicates, or make the first one named something like "Current Editor Profile"

daviwil: Don't worry about duplicates at the moment, let's just naively make the first item have the name of the current editor's profile.

gerane: I like this solution. It will always show the current Editor profile as the default choice, and then if a new profile type is added in another editor, it will be in the List. 

```powershell
Register-EditorCommand `
    -Name 'PowerShellEditorServices.OpenProfileList' `
    -DisplayName 'Open Profile from List (Current User)' `
    -SuppressOutput `
    -ScriptBlock {
        param([Microsoft.PowerShell.EditorServices.Extensions.EditorContext]$context)
        
        $Current = Split-Path -Path $profile -Leaf        
        $List = @($Current,'Microsoft.VSCode_profile.ps1','Microsoft.PowerShell_profile.ps1','Microsoft.PowerShellISE_profile.ps1','Profile.ps1') | Select-Object -Unique
        $Choices = [System.Management.Automation.Host.ChoiceDescription[]] @($List)
        $Selection = $host.ui.PromptForChoice('Please Select a Profile', '(Current User)', $choices,'0')
        $Name = $List[$Selection]
        
        $ProfileDir = Split-Path $Profile -Parent
        $ProfileName = Join-Path -Path $ProfileDir -ChildPath $Name
        
        If (!(Test-Path -Path $ProfileName)) { New-Item -Path $ProfileName -ItemType File }
        
        $psEditor.Workspace.OpenFile($ProfileName)
    }
```

daviwil: Sweet!  Works for me

gerane: Ok, done. The only thing that might be an issue, is if another editor is about to set their $profile to a different path than the normal editor paths. Shouldn't be an issue atm and I am unsure if that is possible. Not exactly sure how that portion works.

daviwil: They won't be allowed to do this.  The strategy is to always use the common paths for profiles, the editor should only be able to choose the name of the profile file.

gerane: Nice, that was what I was hoping, but wasn't 100% sure.

daviwil:
Thanks a lot for kicking off the built-in editor commands work!  I'll file a bug to remind myself to wire this up for the next release.

gerane:
No problem! I will start thinking of other commands that make sense as built-in commands.

PR #256, opened 2016-06-11
Convert Editor Services host to PowerShell module

daviwil:
This (unfortunately large) commit includes two important design changes to
the PowerShell Editor Services hosting model.

The first change is the conversion of the host process to be loaded as a
PowerShell module inside of powershell.exe.  This change enables a lot of
interesting possibilities including the ability to run the Editor Services
host on a remote machine and connect to it from a client machine.  Another
benefit of this approach is the ability to load assembly dependencies
based on the version of PowerShell in which the host is loaded.

Following on from the previous point, the second change is to introduce a
separate build of the PSES host which is compiled against the Nano Server
SDK.  This enables the PSES host to be loaded in the edition of PowerShell
that is shipped with Nano Server.

Both of these changes bring us much closer to enabling a remote editing
experience which would allow someone to use VS Code on a Linux or OSX
machine and connect to an Editor Services host running in a Windows or
Nano Server VM.  We will need to finish the WebSockets channel support
before this is fully possible, though.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/256)


rkeithhill:
That's quite the PR.  :-)  What would be really cool is if this module could get loaded into a PowerShell process that could be used by VSCode as the "terminal".  

I see that VSCode does define some env vars in the PowerShell process it creates for the terminal:
```
VSCODE_CWD                     C:\Program Files (x86)\Microsoft VS Code
VSCODE_IPC_HOOK                \\.\pipe\Code-61fe0e-sock
VSCODE_NLS_CONFIG              {"locale":"en-us","availableLanguages":{}}
VSCODE_PID                     13132
VSCODE_SHARED_IPC_HOOK         \\.\pipe\Code-61fe0e-shared-sock
```
I wonder if any of that could help with this?

gerane:
If you imported the module in integrated terminal, would it be able to
communicate with the editor?

Very interesting idea though Keith. I like where this is headed.

Great job David!

daviwil:
@rkeithhill yeah, I wasn't crazy about that one huge commit but it would have taken a while to break up the changes once I finally got everything working.

Regarding the use of this to have VS Code's terminal connect to it...  I've been thinking about that.  It's possible that once PSES has been loaded by the extension, I could invoke the terminal and have it run powershell.exe but launch a script which connects it to the language service's session so that you'd have a terminal which shares the same session.  Not sure what the details of that will be yet, but my vague sense is that it's possible.

PR #261, opened 2016-07-17
Added OS info to log file.

rkeithhill:
It is handy to know which version of the OS we are running on and what architecture for both the extension hosts and the OS.  Also, converted from \r\n to Environment.NewLine just in case this code runs on Linux/MacOSX someday.  :-)

This outputs to the log file:
```
    PowerShell Editor Services Host v0.0.0.0 starting (pid 2864)...
    
      Host application details:
    
        Name:      Visual Studio Code Host
        ProfileId: Microsoft.VSCode
        Version:   0.6.1
        Arch:      64-bit
    
      Operating system details:
    
        Version: Microsoft Windows NT 10.0.10586.0
        Arch:    64-bit
```

BTW this does use the C# 6 string interpolation feature but I think we already agreed that using C# 6 features was fair game, right?  There are some places where the `null conditional` operator would be real handy - like `session?.Dispose();`

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/261)

### DIFF
--- a/src/PowerShellEditorServices.Host/EditorServicesHost.cs
+++ b/src/PowerShellEditorServices.Host/EditorServicesHost.cs
@@ -86,17 +86,22 @@ namespace Microsoft.PowerShell.EditorServices.Host
                 FileVersionInfo.GetVersionInfo(this.GetType().Assembly.Location);
 #endif
 
+            var newLine = Environment.NewLine;
+
             Logger.Write(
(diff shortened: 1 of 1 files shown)


daviwil:
Yep, C# 6 features are definitely fair game.  Good idea adding the OS version and architecture, should have thought of that before!

PR #271, opened 2016-08-18
Release 0.7.0

daviwil:
This changes brings over the Linux and macOS support that I was working on privately before today's launch.  Since there's a lot of changes in the integration commit, I'll list the highlights:

- Introduced a new TCP channel to provide a commonly-available communication channel
  across multiple editors and platforms
- PowerShell Script Analyzer integration has been shifted from direct use via DLL to
  consuming the PowerShell module and cmdlets
- Updated code to account for platform differences across Windows and Linux/macOS
- Improved stability of the language service when being used in Sublime Text

I fully expect the CI tests to crash and burn after these changes, so I imagine I'll have a little more work to do :)

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershelleditorservices/271)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ lock
 .corext/gen
 registered_data.ini
 .vs/
+module/
 
 docs/_site/
 docs/_repo/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # PowerShell Editor Services Release History
 
+## 0.7.0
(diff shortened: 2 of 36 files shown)


daviwil:
Alright, still the same old recurring test failures (not getting the 'terminated' event).  Everything else looks solid.  I'll go ahead and merge this so that I can move on to getting the build system moved over to compiling the CoreCLR version of PSES.

PR #283, opened 2016-08-31
Add support for debug protocol setVariableRequest.

rkeithhill:


rkeithhill:
The build failed on two two DebugAdapter* unit tests that I don't "think" my code should have impacted.  FWIW all the unit tests pass on my dev machine (except for a few that are skipped).

daviwil:
Same old test failures, definitely not related to your changes.  The code looks awesome!  Merge when ready :)

Issue #298, opened 2016-11-01
Add support for breakpoint hit count


rkeithhill:
Just saw this in the 1.7 insider's release notes.  Sounds like debugger extensions must be modified to support this.  I suspect this should be easy to support.

daviwil:
Sounds good!  That will be a nice addition.

rkeithhill:
Investigated and ran into a snag.  Enh request for PowerShell here: https://github.com/PowerShell/PowerShell/issues/2594

Initial work is committed to my fork: https://github.com/rkeithhill/PowerShellEditorServices/commit/f56a68367494646c4806d2c7011bad8aabd33fa6

Might be able to track the `conditional break` count ourselves and get this done.  Need to think on it some.

daviwil:
Even if this improves in PowerShell 6.0+, we'd still have to make it work for older versions.  I think we'll have to do it ourselves for now.  Maybe we can augment the user's condition code if it helps?

rkeithhill:
True enough.  I'll experiment with maintaining the hit count in a global variable.  Do we have a naming convention for globals created by the debugger?  Maybe something like `__psEditorServices_hitCount`?

rkeithhill:
OK, this seems to work although it does inject a global into the debug session.  I think that should be OK.
```powershell
if (<user expression>) { if (++$global:__psEditorServices_BreakHitCounter_1 -eq <user hit count>) { break } }
```

daviwil:
This is done right?  Should we close it?

rkeithhill:
Yes, it has been implemented.

PR #300, opened 2016-11-05
Add support for Hit Count breakpoints

rkeithhill:
Add support plain Hit Count breakpoints as well as combination of Condition and Hit Count.


rkeithhill:
Fix #298

rkeithhill:
All the unit tests pass on my dev PC under both the Release and Debug configs.

daviwil:
This is awesome man, looks great! Merging it

PR #302, opened 2016-11-18
Remove PSScriptAnalyzer submodule

kapilmb:
EditorServices does not use PSScriptAnalyzer `C#` APIs. Instead it calls `Invoke-ScriptAnalyzer` to get diagnostic markers.


daviwil:
Looks great, thanks Kapil!

PR #304, opened 2016-11-21
Provide quickfix functionality

kapilmb:


src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -31,15 +31,19 @@ public class AnalysisService : IDisposable
...
             "PSUseApprovedVerbs",
+            "PSAvoidUsingPlainTextForPassword",
             "PSReservedCmdletChar",
             "PSReservedParams",
             "PSShouldProcess",
             "PSMissingModuleManifestField",
             "PSAvoidDefaultValueSwitchParameter",
-            "PSUseDeclaredVarsMoreThanAssigments",
-            "PSMisleadingBacktick",
+            "PSUseDeclaredVarsMoreThanAssigments"
```
rkeithhill: We should test the new rule additions.  We got a lot of blow-back from the community early on because of lots of green squiggles in folks' scripts.  Also, we should add a note in the comments for this section, that any updates here should be also made in the https://github.com/PowerShell/vscode-powershell/blob/master/examples/PSScriptAnalyzerSettings.psd1 file as well.  

BTW it is more palatable for folks when the green squiggles underline just the bare minimum of the script as opposed to say entire functions.  :-)   BTW that also helps with not having "overlapping" script extents for multiple rule violations.

kapilmb: I was testing `PSUseDeclaredVarsMoreThanAssigments` so I added it there and forgot to remove it. The rule is still a bit noisy and I would rather remove it.  Thanks for catching it.

rkeithhill:
@kapilmb Sorry, I wasn't reffering to `PSUseDeclaredVarsMoreThanAssigments` as that rule has been in there for a while. It was the other rules you added.  I'm not saying they need to come out - just that we need to verify they don't cause mass green squiggles.  :-)  Specifically referring to `PSAvoidUsingPlainTextForPassword` and `PSUseToExportFieldsInManifest`.  If the green squiggle impact isn't too bad then I say leave them in.

kapilmb:
@rkeithhill Yes, you are right. `PSUseDeclaredVarsMoreThanAssignments` has been there for a while. 

The extents of `PSAvoidUsingPlainTextForPassword` and `PSUseToExportFieldsInManifest` do not cause mass green squiggles. However, the extent of `PSAvoidUsingCmdletAliases` spills over to the entire command. I will fix the extent in the next PSSA release. So, I guess I will just revert the previous commit and after reverting the rule list will look like this:
```powershell
            "PSUseToExportFieldsInManifest", # provides quickfix
            "PSMisleadingBacktick", # provides quickfix
            "PSAvoidUsingCmdletAliases", # provides quickfix
            "PSUseApprovedVerbs",
            "PSAvoidUsingPlainTextForPassword", # provides quickfix
            "PSReservedCmdletChar",
            "PSReservedParams",
            "PSShouldProcess",
            "PSMissingModuleManifestField",
            "PSAvoidDefaultValueSwitchParameter",
            "PSUseDeclaredVarsMoreThanAssigments"
```

One thing I would like to note is that the quickfix scenario will not work until the next release of PSSA (which should be soon.)

kapilmb:
@rkeithhill I have reverted the "rule removal" changes.

rkeithhill:
I didn't notice the addition of `PSAvoidUsingCmdletAliases` before - sorry.  You might want to run that by @daviwil.  That rule is likely to light up a lot in scripts. :-)

kapilmb:
@daviwil rebased on develop branch.

daviwil:
Thanks a lot Kapil!

PR #306, opened 2016-11-29
Fix issue with debugger output not being written for short scripts

### LINKED ISSUE #138: Debugger doesn't flush output reliably
I've noticed while debugging that output from a line that has executed is not shown even though the script has hit a breakpoint on a subsequent line.  I can step once and that is usually enough to get the output to appear.  It would be nice though to have any pending output flushed on debugger stop.

daviwil:
This change fixes an issue where script output for very small scripts is
not being written out reliably before the debug adapter terminates.  The
fix is to add an extra output flush before sending the TerminatedEvent
back to the client.  This fix is temporary until we get the REPL
integration online.

Resolves #138.
Resolves PowerShell/vscode-powershell#284.


daviwil:
Fixing this in a cheeky way just to solve the issue in the short term.

/cc @pcgeek86

daviwil:
Waiting to see if people verify the fix from the preview build I sent out.

PR #307, opened 2016-12-01
Improve PowerShell version and architecture gathering

daviwil:
This change makes some slight tweaks to the PowerShell version and
architecture gathering code to improve the user experience in the host
editors which display those values.

### DIFF
--- a/src/PowerShellEditorServices/Session/PowerShellVersionDetails.cs
+++ b/src/PowerShellEditorServices/Session/PowerShellVersionDetails.cs
@@ -8,6 +8,27 @@ using System;
 namespace Microsoft.PowerShell.EditorServices.Session
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Session/PowerShellVersionDetails.cs
```diff
@@ -8,6 +8,27 @@
...
+        /// <summary>
+        /// The processor architecture is 32-bit.
+        /// </summary>
+        X86,
+
+        /// <summary>
+        /// The processor architecture is 64-bit.
+        /// </summary>
+        X64
+    }
```
daviwil: Need to try running it on Windows 10 IoT ;)

dfinke: Definitely want try this on a Raspberry PI 3

PR #322, opened 2016-12-17
Clear diagnostic markers after file is closed

kapilmb:
When a user closes a file in a workspace, the extension does not clear the diagnostic markers corresponding to the closed file from the Problems window in vscode.

Fixes PowerShell/vscode-powershell#314


kapilmb:
Also fixes PowerShell/vscode-powershell#329

daviwil:
Theoretically we shouldn't have to do this, but I don't see any harm in doing this.  Thanks!

PR #326, opened 2016-12-19
Fix #325: Module dependency loading should pick latest module version

### LINKED ISSUE #325: PSScriptAnalyzer module loading needs to pick the latest installed module
After changing the "bundled modules" behavior in the VS Code extension, the door was opened for an older version of PSScriptAnalyzer to be loaded by accident since the user- and system-wide module paths now take priority.  We need to make sure we only load the latest version of this module no matter where it comes from (bundled or installed).

/cc @gerane

daviwil:
This change fixes an issue with the loading of both the PSScriptAnalyzer
and Plaster modules.  We were simply using Import-Module without
specifying a version to be loaded.  Apparently PowerShell prioritizes the
PSModulePath order over the module version order so there are instances
where an older version of one of these module will be loaded even though a
newer version is available at a different location on the machine.

This change adds extra logic to sort the list of returned modules so that
the latest version of the module is always loaded.

### DIFF
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -179,10 +179,18 @@ namespace Microsoft.PowerShell.EditorServices
             {
                 ps.Runspace = this.analysisRunspace;
 
-                var modules = ps.AddCommand("Get-Module")
-                    .AddParameter("List")
-                    .AddParameter("Name", "PSScriptAnalyzer")
-                    .Invoke();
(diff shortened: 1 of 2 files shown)


daviwil:
@rkeithhill any comments or concerns about this manner of loading the latest version of our dependency modules?  Within the next couple of "major" releases I'm going to add a "ModuleService" that will allow us to go a little further and ask for a version of the module within a specified range so that we won't pick up major version bumps.

rkeithhill:
LGTM

daviwil:
Thanks!

PR #327, opened 2016-12-19
Provide symbol outline for psd1 files

kapilmb:
This update will show the list of keys present in a psd1 file when a user invokes `Go to Symbols in File... (Ctrl+Shift+O)` on an psd1 file.

### DIFF
--- a/src/PowerShellEditorServices/Language/FindSymbolsVisitor.cs
+++ b/src/PowerShellEditorServices/Language/FindSymbolsVisitor.cs
@@ -89,4 +89,48 @@ namespace Microsoft.PowerShell.EditorServices
             return false;
         }
     }
+
+    internal class FindHashtabeSymbolsVisitor : AstVisitor
+    {
+        public List<SymbolReference> SymbolReferences { get; private set; }
+
+        public FindHashtabeSymbolsVisitor()
+        {
+            SymbolReferences = new List<SymbolReference>();
+        }
+
+        public override AstVisitAction VisitHashtable(HashtableAst hashtableAst)
+        {
+            if (hashtableAst.KeyValuePairs == null)
+            {
+                return AstVisitAction.Continue;
+            }
+
+            foreach (var kvp in hashtableAst.KeyValuePairs)
+            {
+                var keyStrConstExprAst = kvp.Item1 as StringConstantExpressionAst;
+                if (keyStrConstExprAst != null)
+                {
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Language/FindSymbolsVisitor.cs
```diff
@@ -89,4 +89,48 @@ private bool IsAssignedAtScriptScope(VariableExpressionAst variableExpressionAst
             return false;
         }
     }
+
+    internal class FindHashtabeSymbolsVisitor : AstVisitor
```
daviwil: Slight typo, **FindHashtabeSymbolsVisitor**

kapilmb:
@daviwil rebased on develop, fixed the type and updated inline docs.

daviwil:
This is pretty awesome!  Really handy to jump down to a specific field in the file, even in nested structures.  It'd be really nice if we could also jump to commented-out fields too but based on some brief research there seems to be no Ast class or visitor for comment lines.  Any thoughts on how we might be able to do that?

daviwil:
Hmmm, we might be able to use the list of tokens returned by the parser to find them.  There's a `TokenKind.Comment` here: https://msdn.microsoft.com/en-us/library/system.management.automation.language.tokenkind%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

rkeithhill:
> there seems to be no Ast class or visitor for comment lines

I've asked Jason about this. This is one area where I wish PowerShell were more like Roslyn which retains all source code trivia (like whitespace between tokens and comments).

daviwil:
I've considered creating a higher-level abstraction that can (literally) fill in the gaps of the Ast for things like that.  I think we're going to need it for making code formatting rules in PSScriptAnalyzer.

kapilmb:
I think we can tackle some basic scenarios without needing any help from Asts. For example, 
```powershell
@{
  ModuleVersion = 0.1
  Author = "XYZ"
 # RootModule = ".\test.psm1"
 # FunctionsToExport = @('get-foo')
}
```
Here, we just extract out the commented fields with some regular expression help.

daviwil:
@kapilmb I think this PR is done, right?  Can I merge it?

kapilmb:
Yes, this done.

daviwil:
Cool!  Merging it.

PR #329, opened 2016-12-20
Prevent concurrent commands in analysis services

kapilmb:


kapilmb:
@daviwil Can you please verify this on your side too. I am not able to reproduce PowerShell/vscode-powershell#400 consistently on my PC. By the time I open the PSSA rules menu PSES is already initialized, but sometimes I can get lucky to get the session termination error.

daviwil:
Just tried this out, works fine for me!  I added a 5 second thread sleep after the diagnostic records were returned so that I can emulate a long delay.  While it was sleeping, I tried to get the PSSA rules using the menu.  The menu didn't appear until after the diagnostics got sent back to the editor.  Looks good.

kapilmb:
Thanks @daviwil

PR #330, opened 2016-12-30
Enable language and debugging features to use remote and attached runspaces

daviwil:
Still finishing this up, needs some polish and minor refactoring here and there.  Also need to enable remote file editing, but that should be pretty easy now.

@rkeithhill, take a look when you have some time, I'd certainly appreciate a second pair of eyes on the debugging changes!


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -186,15 +187,45 @@ protected override void Shutdown()
...
+                // TODO: Version check!
+
+                await this.editorSession.PowerShellContext.ExecuteScriptString(
+                    $"Enter-PSHostProcess -Id {attachParams.ProcessId}");
+
+                if (attachParams.RunspaceId > 0)
+                {
+                    // Execute the Debug-Runspace command but don't await it because it
+                    // will block the debug adapter initialization process.
+                    Task nonAwaitedTask =
```
rkeithhill: Do you need to create this variable? It doesn't seem to be used later.

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -82,10 +89,31 @@ public DebugService(PowerShellContext powerShellContext)
 
             if (breakpoints.Length > 0)
             {
+                // Make sure we're using the remote script path
+                string scriptPath = scriptFile.FilePath;
+                if (this.powerShellContext.CurrentRunspace.Locale == RunspaceLocale.Remote)
```
rkeithhill: Locale might be confused with i18n "locale".  Or perhaps you don't have control over that name.

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -612,16 +640,16 @@ public VariableScope[] GetVariableScopes(int stackFrameId)
                 new VariableContainerDetails(this.nextVariableId++, "Scope: " + scope);
             this.variables.Add(scopeVariableContainer);
 
-            var results = await this.powerShellContext.ExecuteCommand<PSVariable>(psCommand, sendErrorToHost: false);
+            var results = await this.powerShellContext.ExecuteCommand<PSObject>(psCommand, sendErrorToHost: false);
```
rkeithhill: So when querying variables in a remote runspace, I take it we don't get back PSVariable types but PSObject instead?

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -686,9 +735,14 @@ private bool AddToAutoVariables(PSVariable psvariable, string scope)
...
             PSCommand psCommand = new PSCommand();
-            psCommand.AddCommand("Get-PSCallStack");
 
-            var results = await this.powerShellContext.ExecuteCommand<CallStackFrame>(psCommand);
+            // This glorious hack ensures that Get-PSCallStack returns a list of CallStackFrame
+            // objects (or "deserialized" CallStackFrames) when attached to a runspace in another
+            // process.  Without the intermediate variable Get-PSCallStack inexplicably returns
+            // an array of strings containing the formatted output of the CallStackFrame list.
+            psCommand.AddScript("$global:__psEditorServices_CallStack = Get-PSCallStack; $global:__psEditorServices_CallStack");
+
```
rkeithhill: I think I create variables like this for the hit count impl. We should probably create a string const somewhere like `PsesGlobalVariableNamePrefix = "__psEditorServices_";`

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -686,9 +735,14 @@ private bool AddToAutoVariables(PSVariable psvariable, string scope)
         private async Task FetchStackFrames()
         {
             PSCommand psCommand = new PSCommand();
-            psCommand.AddCommand("Get-PSCallStack");
 
-            var results = await this.powerShellContext.ExecuteCommand<CallStackFrame>(psCommand);
+            // This glorious hack ensures that Get-PSCallStack returns a list of CallStackFrame
+            // objects (or "deserialized" CallStackFrames) when attached to a runspace in another
```
rkeithhill: +1 At least you know where the responsible parties sit.  :-)

src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
```diff
@@ -58,19 +74,21 @@ public class StackFrameDetails
...
             VariableContainerDetails autoVariables,
             VariableContainerDetails localVariables)
         {
             return new StackFrameDetails
             {
-                ScriptPath = callStackFrame.ScriptName ?? "<No File>",
-                FunctionName = callStackFrame.FunctionName,
-                LineNumber = callStackFrame.Position.StartLineNumber,
-                ColumnNumber = callStackFrame.Position.StartColumnNumber,
+                ScriptPath = (callStackFrameObject.Properties["ScriptName"].Value as string) ?? NoFileScriptPath,
```
rkeithhill: I wonder if over time we will want two classes: `LocalStackFrameDetails` and `RemoteStackFrameDetails`.  This class isn't that complicated now so it probably isn't worth it but if it creeps in complexity over time, maybe we reconsider?

src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -532,22 +532,17 @@ public FindOccurrencesResult FindSymbolsInFile(ScriptFile scriptFile)
             return new List<ScriptFile>().ToArray();
         }
 
-        private async Task<SymbolReference> FindDeclarationForBuiltinCommand(
-            CommandInfo cmdInfo, 
+        private SymbolReference FindDeclarationForBuiltinCommand(
+            CommandInfo commandInfo,
```
rkeithhill: Spelled out is better. :+1:

src/PowerShellEditorServices/Session/PowerShellVersionDetails.cs
```diff
@@ -72,5 +81,83 @@ public class PowerShellVersionDetails
...
+        #endregion
+
+        #region Public Methods
+
+        /// <summary>
+        /// Gets the PowerShell version details for the given runspace.
+        /// </summary>
+        /// <param name="runspace">The runspace for which version details will be gathered.</param>
+        /// <returns>A new PowerShellVersionDetails instance.</returns>
+        public static PowerShellVersionDetails GetVersionDetails(Runspace runspace)
```
rkeithhill: Nice!

src/PowerShellEditorServices/Session/RunspaceDetails.cs
```diff
@@ -0,0 +1,182 @@
...
+using Microsoft.CSharp.RuntimeBinder;
+using Microsoft.PowerShell.EditorServices.Utility;
+using System.Management.Automation.Runspaces;
+
+namespace Microsoft.PowerShell.EditorServices.Session
+{
+    /// <summary>
+    /// Specifies the possible types of a runspace.
+    /// </summary>
+    public enum RunspaceLocale
```
rkeithhill: Aha, you do have control over this name.  How about `RunspaceLocation`?

src/PowerShellEditorServices/Utility/ThreadSynchronizationContext.cs
```diff
@@ -25,6 +25,11 @@ public class ThreadSynchronizationContext : SynchronizationContext
 
         #region Constructors
 
+        public ThreadSynchronizationContext()
+        {
+            var hash = this.GetHashCode();
```
rkeithhill: What does this do?

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -202,6 +213,35 @@ public string GetRelativePath(string filePath)
...
+        internal static string MapRemotePathToLocal(string remotePath, string connectionString)
+        {
+            string sessionTempPath =
+                Path.Combine(
+                    Path.GetTempPath(),
+                    "PSES-" + Process.GetCurrentProcess().Id,
+                    "RemoteFiles");
+
+            // The path generated by this code will look something like
+            // %TEMP%\PSES-[PID]\RemoteFiles\1205823508\computer-name\MyFile.ps1
```
rkeithhill: Do you think we need the PID?  One advantage of using a hash in the name is that if the file doesn't change, debugging it a second time doesn't create a new copy of the same file on my file system.  Then again, PowerShell script files usually aren't very big and folks are probably used to cleaning their temp folder from time to time.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
daviwil: Yeah, the C# compiler gives a warning if you call an async method without doing something with the task.  I've been using this trick to silence the message in cases where I don't actually want to await the task.

src/PowerShellEditorServices/Debugging/DebugService.cs
daviwil: Yep, that's one of the things I'm going to change, will say more in your other comment.

src/PowerShellEditorServices/Debugging/DebugService.cs
daviwil: Remote/attached runspaces return some concrete type that looks like a PSVariable but isn't the same actual type, thus I can't pull them back out as PSVariable.  Same thing happened with the CallStackFrames.  Not sure why it was done this way, but probably because remote sessions were added later and the types weren't designed to be remotable in the beginning.

src/PowerShellEditorServices/Debugging/DebugService.cs
daviwil: Yep, I stole your naming pattern :)  I'll add a const string for that.

src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
daviwil: Definitely, if there turns out to be a good reason for it we can make the change.

src/PowerShellEditorServices/Session/PowerShellVersionDetails.cs
daviwil: I've finally started breaking down the code in the PowerShellContext class to put it where it belongs!

src/PowerShellEditorServices/Session/RunspaceDetails.cs
daviwil: Yeah, I was trying to find a good name for it when I thought I needed a separate designation between "location" and whether the runspace was created or attached to.  Turns out that `Debug-Runspace` merely just blocks the current runspace's pipeline and hooks into the other runspace's debugger, so no runspace change occurs.

I think I'm just going to collapse those two enums back into one which has 3 values, `Local`, `LocalProcess`, and `Remote` to reflect the 3 states that I can discern via runspace changes.  I think `Create` vs `Attach` should be a debugger distinction since it really only matters for breakpoint management (cleaning up breakpoints when detaching, etc).

src/PowerShellEditorServices/Utility/ThreadSynchronizationContext.cs
daviwil: Ahhh, I forgot to take that out.  I was debugging a gnarly threading issue when I first added the "attach" debug mode; the old ThreadSynchronizationContext wasn't being cleared out properly so it broke future debugging sessions.

src/PowerShellEditorServices/Workspace/Workspace.cs
daviwil: That's a good point, though the one thing I worry about is the locally cached file being out of date.  For example, the user connects to a remote machine to debug a script running there, finds a bug, fixes it and re-deploys the script.  They attempt to debug the same script again remotely but since we already have the cached file we show the old version.

Maybe I could send the local copy's timestamp to the server to determine whether we need to sync, but that might be a little overkill.  I'm certainly interested to go that direction in the future if the need arises.

I guess this problem will exist with the current approach if PSES somehow ends up with the same PID after X launches, or even if they change the remote file while in the same editing session.  Maybe I should proactively delete that remote files folder once the remote runspace session ends?

daviwil:
I'm happy with how it's turned out so far.  There's a few rough edges to smooth out but the experience is pretty solid at the moment.  Fingers crossed that it works correctly on Linux/macOS ;)

src/PowerShellEditorServices/Debugging/DebugService.cs
daviwil: Haha, yep!  I'll be asking some questions next week ;)

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: But if you're including the hash of the file's contents in the filename, any change would result in a new hash and hence a new file.  Or maybe I'm misunderstanding the "hash slug".

daviwil: This is a hash of the file's folder path (sans filename) on the remote machine so a file will always end up with the same "slug" (I should use a better name) when loaded in the future so long as its path remains the same.

daviwil:
Finally got the changes finished up!  Take a look at the latest commit if you're interested, I'll merge it tomorrow at some point.  I'm feeling pretty good about the stability and experience here.  Tomorrow I'll also send a PR for the extension changes to add the new debug configurations, etc.

daviwil:
Weird test failures, not sure what I broke in some cases.  I'll take a look tomorrow.

src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -784,7 +849,7 @@ private bool AddToAutoVariables(PSVariable psvariable, string scope)
                         if (hitCount.HasValue)
                         {
                             string globalHitCountVarName =
-                                $"$global:__psEditorServices_BreakHitCounter_{breakpointHitCounter++}";
+                                $"$global:{PsesGlobalVariableNamePrefix}BreakHitCounter_{breakpointHitCounter++}";
```
rkeithhill: 👍

src/PowerShellEditorServices/Session/RunspaceDetails.cs
```diff
@@ -0,0 +1,202 @@
...
+using Microsoft.CSharp.RuntimeBinder;
+using Microsoft.PowerShell.EditorServices.Utility;
+using System.Management.Automation.Runspaces;
+
+namespace Microsoft.PowerShell.EditorServices.Session
+{
+    /// <summary>
+    /// Specifies the possible types of a runspace.
+    /// </summary>
+    public enum RunspaceLocation
```
rkeithhill: 👍

daviwil:
Managed to fix the test issues, now I just need to verify that the changes work correctly across the range of supported PS versions and I'll get it merged!

PR #331, opened 2017-01-11
Fetch script listings for debugger stop events with no ScriptName

daviwil:
This change adds new behavior to the DebugService which enables it to
gather the listing of any source code that is being executed outside of a
script file when the debugger hits a breakpoint.  This is useful for
allowing the user to step through code in this type of scenario.

Currently this is being used for an unreleased feature in PowerShell 6.0
which allows you to step into ScriptBlocks run against remote sessions
with Invoke-Command.


rkeithhill:
Cool!

PR #332, opened 2017-01-12
Add feature to format PowerShell source files

kapilmb:


daviwil:
Oh, I just realized there's a build failure here in the PowerShell v3 and v4 API tests:

```
Language\LanguageService.cs(480,58): error CS1061: 'HashtableAst' does not contain a definition for 'SafeGetValue' and no extension method 'SafeGetValue' accepting a first argument of type 'HashtableAst' could be found
```

kapilmb:
Yes, even I noticed it only a couple of minutes ago. PSv3 and PSv4 do not have the `ast.safegetvalue()` api. For PSSA we wrote our own hashtable parser to support PSv3 and PSv4. But I do not think that is a very good idea. Another approach is to transmit an object instead of hashtable literal and convert it to hashtable at the server.

daviwil:
Damn, I think I had written a review comment about this in the past but I might have deleted it.  Here's what you can do:

- Change `ScriptFileMarkerRequestParams.settings` to a `JObject` instead of `string`
- When you want to convert to a `Hashtable`, use the `JObject` instance method `ToObject` to convert to that type

JObject comes from Newtonsoft.Json.  On the client side, instead of serializing your settings object into a string just pass it directly in the `settings` property when you make the request.

daviwil:
Seems like we're still getting the build failures from the use of `HashtableAst.SafeGetValue`, is this code needed anymore: 

https://github.com/PowerShell/PowerShellEditorServices/blob/bfcec76bd465deb476b49a6ae4abe5077c2d4111/src/PowerShellEditorServices/Language/LanguageService.cs#L480

kapilmb:
I forgot to remove the redundant method. The tests now don't fail on compilation for v3 and v4.

daviwil:
Awesome, thanks a lot!  Merging it.

kapilmb:
@daviwil Thanks!

PR #334, opened 2017-01-14
Remove submodule instructions from README.md.

rkeithhill:
... Or we could just remove the whole section on cloning since it is just a simple clone operation now.  What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,10 @@ how to use this project. You can also read our plans for future feature developm
 
 ## Cloning the Code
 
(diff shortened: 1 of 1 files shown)


rkeithhill:
Hmm, I branched this from `develop`.  I can merge this to develop and assume this change gets propagated when you merge `develop` to `master` for 0.9.0, right?

daviwil:
Yep, merge it to develop and I'll take it to master at release time.

PR #337, opened 2017-01-16
Implement psedit command for remote file loading

daviwil:
This change adds new behavior to RemoteFileManager so that a 'psedit'
command will be added to all sessions for the purpose of loading local or
remote files in the debugging experience.  For now the remotely opened
files do not have saved contents propagated to the remote machine but that
will be added shortly.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -344,7 +344,6 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             RequestContext<SetBreakpointsResponseBody> requestContext)
         {
             ScriptFile scriptFile = null;
-            Exception notFoundException = null;
 
             // Fix for issue #195 - user can change name of file outside of VSCode in which case
             // VSCode sends breakpoint requests with the original filename that doesn't exist anymore.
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -322,15 +322,25 @@ protected override void Shutdown()
...
+            Exception notFoundException = null;
 
             // Fix for issue #195 - user can change name of file outside of VSCode in which case
             // VSCode sends breakpoint requests with the original filename that doesn't exist anymore.
             try
             {
                 scriptFile = editorSession.Workspace.GetFile(setBreakpointsParams.Source.Path);
             }
-            catch (FileNotFoundException)
+            catch (DirectoryNotFoundException e)
```
rkeithhill: You could use:
```csharp
catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
{
    notFoundException = ex;
}
```

daviwil: Hah!  Yep, haven't had a chance to use that one yet.

daviwil:
Doesn't work in all cases yet unfortunately.  Seems that either PowerShell isn't forwarding the event all the way from nested sessions (debugging a runspace in a local or remote process) back to the local session.  I'll have to bug somebody about that today.

PR #340, opened 2017-01-18
Make analysis service methods async

kapilmb:
Also improves the code formatting experience as the diagnostics now potentially run in a separate runspace and hence do not block code formatting request.


daviwil:
Interesting idea using a RunspacePool for this.  Let's give it a shot and see how it goes this release.  Does it seem pretty stable at this point?

kapilmb:
I think so, and it definitely makes the formatting experience much smoother.

daviwil:
Awesome!  Merging it.

PR #354, opened 2017-01-25
Fix crash caused when formatting document

kapilmb:
The crash was very likely due to deferred execution when enumerating through the diagnostic records returned by PowerShell invocation.


kapilmb:
I haven't observed any crash after making this fix yet.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -102,7 +110,10 @@ public AnalysisService(IConsoleHost consoleHost, string settingsPath = null)
             try
             {
                 this.SettingsPath = settingsPath;
+
+                scriptAnalyzerModuleInfo = FindPSScriptAnalyzerModule();
                 var sessionState = InitialSessionState.CreateDefault2();
+                sessionState.ImportPSModulesFromPath(scriptAnalyzerModuleInfo.ModuleBase);
```
daviwil: Thanks for changing that back!  I felt dirty adding the other fix :)

PR #360, opened 2017-01-31
Update projects and build script to .NET SDK 1.0.0-rc3

daviwil:


daviwil:
Alright, this seems to be working despite the test failures.  Those failing tests need to be fixed, they've gone beyond the realm of annoyance.  Will work on that tonight.

src/PowerShellEditorServices.Channel.WebSocket/PowerShellEditorServices.Channel.WebSocket.csproj
```diff
@@ -1,20 +1,13 @@
-﻿<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), PowerShellEditorServices.Common.props))\PowerShellEditorServices.Common.props" />
 
   <PropertyGroup>
     <AssemblyTitle>PowerShell Editor Services WebSocket Protocol Channel</AssemblyTitle>
-    <TargetFrameworks>net451;netstandard1.6</TargetFrameworks>
+    <TargetFramework>net451</TargetFramework>
```
rkeithhill: So how does this compile for .NET Core now?  Via the PackageTargetFallback property below?

daviwil: It wasn't even compiling for netstandard1.6 before, I used a condition to take out the list of compiled files for that TargetFramework.  That trick didn't seem to work for the new csproj changes so I decided to take the WebSocket projects out of the solution for now.  The NuGet dependencies that it uses aren't supported yet on .NET Core unfortunately.

Based on some feedback from some other folks, I think we're going to switch to an HTTP long-polling model instead of WebSockets for a web-based transport.

rkeithhill: Interesting, thanks. BTW what issues did folks have with WebSockets?

daviwil: We want to use it in Azure to provide a better PowerShell editing experience, due to the complexity of Azure Resource Manager's architecture, there's no way to maintain a persistent WebSocket connection between a browser and a backend server.  Others have been using long-polling for "connected" sessions so I figured it might be better to make that the primary approach.

daviwil: One other issue is that AFAIK there is no WebSockets implementation for CoreCLR yet, but maybe that's changing soon.

rkeithhill: Yeah, I thought David Fowler and crew were working on a SignalR implementation for ASP.NET Core due late this year.  Also this site claims WebSockets is already in CoreFX:
https://apisof.net/catalog/System.Net.WebSockets

PR #361, opened 2017-01-31
Clean up CI tests and build output

daviwil:
These changes help the CI build fail far less often and prevent it from spewing compiler warnings.


daviwil:
I'm a dubious about this fix, but I suppose it's possible that this could be the issue.

daviwil:
There are still a couple of tests failing randomly here and there, I'll investigate them at a later date since I think they fail much less frequently than the DebugAdapterTests did.

PR #364, opened 2017-02-03
Add CompletionItemKind.Folder.

rkeithhill:
They finally added this in 1.9.0. Noticed as I was reading through the release notes.


rkeithhill:
I'm not sure how VSCode < 1.9 will handle an enum value it doesn't understand.  If it ignores it then we're good.  If not, then I guess we'd have to set the min version to 1.9.  Here is what it looks like:

image

daviwil:
Damn, they finally added it!  I would guess that it won't crash at the very least.  If it does we'll just tell people to upgrade.

PR #366, opened 2017-02-06
Fix crashes caused by multiple analysis runspaces

kapilmb:
Having multiple runspaces crashed scriptanalyzer invocation because of the following two issues:
* PSCmdlet.WriteOutput being called from another thread than Begin/Process
* CompositionContainer.ComposeParts complaining that "...Only one batch can be composed at a time"

We used multiple runspaces in the first place because PSSA was slow and having multiple runspaces sped things up by executing PSSA in a parallel manner. But having multiple runspaces causes a lot of issues which are hard to debug. Also, PSSA has become significantly fast compared to the previous version and thereby having just one runspace should not deteriorate the performance significantly. And on the positive side we do not have to deal with the aforementioned crashes.

### DIFF
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -333,13 +333,6 @@ namespace Microsoft.PowerShell.EditorServices
 
         private PSObject[] InvokePowerShell(string command, IDictionary<string, object> paramArgMap)
         {
-            var task = InvokePowerShellAsync(command, paramArgMap);
-            task.Wait();
-            return task.Result;
-        }
-
-        private async Task<PSObject[]> InvokePowerShellAsync(string command, IDictionary<string, object> paramArgMap)
-        {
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -24,7 +24,7 @@ public class AnalysisService : IDisposable
     {
         #region Private Fields
 
-        private const int NumRunspaces = 2;
+        private const int NumRunspaces = 1;
```
daviwil: Should we just get rid of the RunspacePool now?

kapilmb: Runspacepool does the queuing of commands for us. If we do not use runspacepool then we will need to back to using a lock to prevent concurrent PS invocation.

daviwil: Hmmm, that's interesting...  I hadn't considered that before, maybe it would be useful in a broader context in PSES.  Thanks for the tip!

PR #367, opened 2017-02-07
Improve detection of runspace changes

daviwil:
This change adds new behavior that makes the detection of sudden runspace
changes much more reliable.  These changes were particularly necessary for
detecting when we start debugging a runspace in a remote session since we
don't get a PushRunspace notifications when that happens.

This change also improves how we gather the prompt string just after a
command is executed.


daviwil:
Gonna go ahead and merge this because I need it for another change.  Feel free to leave feedback and I'll address it later this week.

PR #368, opened 2017-02-08
Add integrated console support via custom host implementation

daviwil:
This change enables a true integrated terminal experience for editors like
VS Code which have an embedded terminal UI.  The change centers primarily
around the new ConsoleReadLine class which provides a "good enough"
terminal editing and tab completion experience until we can find a way to
make PSReadLine work inside of our custom host.

The debugging experience has also been updated to share the language
service's runspace.  In a later change an independent debugging mode will
be added.


daviwil:
This change is pretty massive so don't be in a hurry to review it.  I've got a pretty good feeling that many tests will fail, I'll look into those later today.  I might have to disable some tests temporarily until I can come up with a new testing strategy that leverages the new console experience.  My goal is to get this merged today so that people can try out a preview build.

I'll be sending a corresponding PR to the vscode-powershell repo after this gets merged.

rkeithhill:
+4,659 −2,686  Now that is one heck of a PR!

daviwil:
It's too big...  And a couple files got their line endings messed up so it might be reporting incorrectly.  Still hacking through some final issues, the rabbit hole goes deeper and deeper...

src/PowerShellEditorServices/Console/ConsoleReadLine.cs
```diff
@@ -19,8 +19,9 @@ namespace Microsoft.PowerShell.EditorServices.Console
 
     internal class ConsoleReadLine
     {
-        #region Private Fields
+        #region Private Field
 
+        private bool isUnix = false;
```
rkeithhill: C# fields are automatically initialized to their default values which is false for a bool.

daviwil: Yeah, not sure why I initialized it.  I was trying to scratch out some code in VS Code on Mac OSX and I think the lack of IntelliSense was screwing up my senses.  Something about the net451 TargetFramework was making OmniSharp fail to load the projects correctly.

Julianzs:
Very excited about this change, I was told it would meet the exact requirement I was looking for that exists today in the PowerShell ISE: http://stackoverflow.com/questions/42607282/using-visual-studio-code-with-interactive-powershell-window-like-powershell-is/42607437#42607437. Thanks, David!

PR #369, opened 2017-02-10
Fix debug hang when watch is defined

### LINKED ISSUE #357: Debugger stops responding after a bit of debugging with watches defined
VS Code 1.8.1, PowerShell ext 0.9.0, Windows 10 AU.

I'm finding that if I have any watch defined, the debugger becomes unresponsive after hitting a few breakpoints.  I didn't see anything obvious in the log file.

[1485651267-212b9958-570a-4602-8228-df639d9e79441485647668037.zip](https://github.com/PowerShell/PowerShellEditorServices/files/737234/1485651267-212b9958-570a-4602-8228-df639d9e79441485647668037.zip)

kapilmb:
Resolves #357

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -678,23 +678,11 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
             }
             else
             {
-                VariableDetailsBase result;
-                if (evaluateParams.Context.Equals("watch", StringComparison.OrdinalIgnoreCase))
-                {
-                    // If we evaluate an expression from a watch context then the server goes into a deadlock.
-                    // So instead, we retrieve the value from already evaluated variables.
-                    result = editorSession.DebugService.GetVariableFromExpression(
+                VariableDetails result =
+                    await editorSession.DebugService.EvaluateExpression(
                         evaluateParams.Expression,
-                        evaluateParams.FrameId);
-                }
-                else
-                {
-                    result =
-                        await editorSession.DebugService.EvaluateExpression(
-                            evaluateParams.Expression,
-                            evaluateParams.FrameId,
-                            isFromRepl);
-                }
+                        evaluateParams.FrameId,
+                        isFromRepl);
 
                 if (result != null)
                 {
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -678,11 +678,23 @@ protected override void Shutdown()
...
             else
             {
-                VariableDetails result =
-                    await editorSession.DebugService.EvaluateExpression(
+                VariableDetailsBase result;
+                if (evaluateParams.Context.Equals("watch", StringComparison.OrdinalIgnoreCase))
+                {
+                    // If we evaluate an expression from a watch context then the server goes into a deadlock.
+                    // So instead, we retrieve the value from already evaluated variables.
+                    result = editorSession.DebugService.GetVariableFromExpression(
```
rkeithhill: Keep in mind that the user can enter arbitrary expressions - not just variable names e.g. `1+1`, `Get-Date`, `$count+10`

daviwil: Yep.  I think it's better if we can figure out a way to deal with the root cause here, I'll chat with Kapil about it when I get to the office later.

kapilmb: Good point. I didn't think about evaluating an arbitrary expression when I found the underlying issue. I would like to think of this as a temporary fix till we figure out the right way to do it. As I have mentioned in `powershellcontext.cs`, the hang is the result of the server trying to acquire a handle to an already occupied runspace and that runspace won't be release till the script finishes execution. But the bigger question to ask is why does vscode send a `watch` evaluation request when the user `continues` and not when the debugger hits a breakpoint, unlike variables, scopes, etc? Something doesn't seem right here.

kapilmb: Made changes to allow arbitrary expression evaluation as before.

daviwil:
I've got a simpler fix for this now, and VS Code team has fixed it on their side as well.  I'll close this PR in favor of the fix that's coming in the interactive console PR.

Thanks for working on this Kapil!

PR #370, opened 2017-02-14
Enable formatOnType feature

kapilmb:


src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -569,6 +591,55 @@ public FindOccurrencesResult FindSymbolsInFile(ScriptFile scriptFile)
...
+
+                if (scriptExtent.EndLineNumber == lineNumber)
+                {
+                    return scriptExtent.EndColumnNumber >= columnNumber;
+                }
+
+                return true;
+            }, true);
+
+            if (asts == null || asts.Count() == 0)
```
daviwil: IEnumerable<T>.Any() is way faster than IEnumerable<T>.Count() because it checks if there's at least one item rather than iterating over the whole collection to get the full count.  Doesn't matter at all in this case, just a helpful tip to pass along :)

src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -569,6 +591,55 @@ public FindOccurrencesResult FindSymbolsInFile(ScriptFile scriptFile)
...
+            {
+                return null;
+            }
+
+            Func<IScriptExtent, int> getExtentWitdh = extent => extent.EndOffset - extent.StartOffset;
```
daviwil: Since this is such a simple calculation that is used on 2 lines I'd just inline it instead of trying to make a reusable func, faster that way.  Also not important to change for this PR

src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -569,6 +591,55 @@ public FindOccurrencesResult FindSymbolsInFile(ScriptFile scriptFile)
...
+            var minDiff = getExtentWitdh(scriptFile.ScriptAst.Extent);
+            Ast minAst = scriptFile.ScriptAst;
+            foreach (var ast in asts)
```
daviwil: Can you add a comment explaining what this loop is doing?  Isn't perfectly clear why it's necessary.

daviwil:
👍

kapilmb:
@daviwil Thanks for the tips. I have updated the code to take into account all your suggestions.

daviwil:
If it's ready to merge, go for it!

kapilmb:
Thanks @daviwil!

PR #371, opened 2017-02-17
Improve detection of session changes when debugging ScriptBlocks

daviwil:


daviwil:
@rkeithhill @kapilmb Note that I've updated to .NET SDK RC4, you'll want to update to the latest VS 2017 RC build (check the Updates UI in VS) so that it can be built in Visual Studio

rkeithhill:
Thanks. At this point, I'll probably wait for the RTM bits on March 7th. I've already hosed one machine with a VS 2017 RC (can't update to a newer version and can't fully uninstall the old version). :-(

daviwil:
I just realized that the RC4 bump won't have any effect on VS, if you're using the VS 2017 that supports RC3 everything will still build as it should.

PR #372, opened 2017-02-18
Modify ExecuteScriptAtPath to handle arb script.

rkeithhill:
Renamed to ExecuteScriptWithArgs and check if "script" is a filename before attempting to escape wildcard chars in path.


daviwil:
Sweet, thanks!

PR #375, opened 2017-02-19
Enable column breakpoints.

rkeithhill:
In response to https://github.com/Microsoft/vscode/issues/14784


daviwil:
Awesome, looks like you already had that wired up so this was pretty easy :)

rkeithhill:
Almost wired up.  Was just missing a few bits.  :-)

PR #398, opened 2017-03-21
Preserve PowerShell's CommandCompletion result sort order in VS Code

daviwil:
This change causes VS Code to preserve PowerShell's original
CommandCompletion result sort order by creating SortText entries for every
completion result, not just parameters.

Resolves PowerShell/vscode-powershell#579.


daviwil:
@rkeithhill any objection to this?  Seems the most expedient way to force VS Code to preserve PowerShell's sort order.  I get sorted IntelliSense results identical to the ISE after this change.

rkeithhill:
No objections.  LGTM.

daviwil:
Sweet, thanks!

PR #404, opened 2017-03-23
Add command to get pertinent version info.

rkeithhill:
I'm not sure about the noun on this command.  Open to suggestions.  But if we take this then the VSCode-PowerShell issue template simplifies to having folks run the command from the IC and paste results into issue.  Another option would be to have them open the logs and copy the version info from there.

Also, I don't have a Mac so the command to get OS version `lsb_release -d` needs to be tested and potentially replaced with the correct command.

### DIFF
--- a/module/PowerShellEditorServices/PowerShellEditorServices.psd1
+++ b/module/PowerShellEditorServices/PowerShellEditorServices.psd1
@@ -66,7 +66,7 @@ Copyright = '(c) 2016 Microsoft. All rights reserved.'
 # NestedModules = @()
 
 # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
-FunctionsToExport = @('Start-EditorServicesHost')
+FunctionsToExport = @('Start-EditorServicesHost', 'Get-PowerShellEditorServicesVersion')
 
 # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
 CmdletsToExport = @()
--- a/module/PowerShellEditorServices/PowerShellEditorServices.psd1
+++ b/module/PowerShellEditorServices/PowerShellEditorServices.psd1
@@ -66,7 +66,7 @@ Copyright = '(c) 2016 Microsoft. All rights reserved.'
 # NestedModules = @()
 
 # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
-FunctionsToExport = @('Start-EditorServicesHost', 'Get-PowerShellEditorServicesVersion')
+    FunctionsToExport = @('Start-EditorServicesHost', 'Get-PowerShellEditorServicesVersion', 'Compress-LogDir')
 
 # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
 CmdletsToExport = @()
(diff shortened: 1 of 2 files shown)


rkeithhill:
Hmm, given that PSES is supposed to be cross editor perhaps having this command check VSCode version & extensions isn't the right way to go.  Maybe this would be better as a VSCode command that copied version to the clipboard?  It would have the advantage of being able to copy to the clipboard on Linux and macOS.  It would just need a way to get the PSVersionTable back from the PowerShell session hosting PSES (as well as OS info).  Presumably we can get the VSCode version from within VSCode.

daviwil:
I think your general approach is a good one, but we may be able to take it one step further:

- `Get-PowerShellEditorServicesVersion` returns the editor-agnostic version information about PSES, PowerShell version, and OS details and merely prints to output, no clipboard behavior
- PowerShell extension registers a new built-in command "File GitHub issue for PowerShell extension" which invokes the `powershell.developer.powerShellExePath` with a simple script that runs `Get-PowerShellEditorServicesVersion` and the other `code` commands, then scrapes that info into a string
- This command then invokes the browser using GitHub's [issue creation URL syntax](http://stackoverflow.com/a/34147022).  The fields are pre-populated and the user can now type in what they were experiencing
- Maybe provide another command, "Upload logs to GitHub issue" which asks for a GitHub issue number (maybe providing a quick-pick list of the last 10 issues on the repo) and then zips/uploads the logs to that issue using GitHub APIs.

It's all a lot of work, but would probably make issue filing much easier.  For now we could just start with the editor-agnostic command in the PSES module.

Probably need to think a little bit on the name, but we can go ahead and merge it while thinking about it.

rkeithhill:
Sounds like a great plan to me.  The Upload logs would be especially handy as in theory we could use PowerShell >= 5 to zip the logs and upload the zip.  Not sure what to do about v3/v4.  Perhaps use the COM Shell object to compress the files?  Seems like there is a compress function in there somewhere.

daviwil:
Yeah, using a COM object seems fine to me if it works on Win7 and up.  I've been wanting to do this kind of automated issue posting thing for a while now.

rkeithhill:
With the latest change, the following string is returned:
```
70:71ms> Get-PowerShellEditorServicesVersion
PSES module version: 0.11.0
PSVersion:           5.1.14393.953
PSEdition:           Desktop
PSBuildVersion:      10.0.14393.953
CLRVersion:          4.0.30319.42000
Operating system:    Windows 64-bit 10.0.14393
```
There are other `$PSVersionTable` fields but I'm not sure we need them.

I'm assuming the code to actually submit an issue will be in the VSCode extension. However it may be easier to ZIP the log directory from PowerShell.  Should I add a command that takes the Log directory path and zip it using the appropriate method on the OS?  I would likely put the ZIP file beside the original log dir.  Thoughts?

rkeithhill:
I added a `Compress-LogDir` command but I don't like exposing such a "general purpose" command from the module.  Maybe we have just one command called something like `New-IssueReport` that zips the folder and returns an object with the zipPath and version info?

daviwil:
Yeah, `Compress-LogDir` is potentially too generic to export.  Since `New-IssueReport` is also pretty generic, I wonder how we can do this in a way that could work for all editors that use PSES to post the issue to the appropriate repo of that editor extension.  It's possible that we could just include this as a utility script with the PSES module rather than exporting it as a command.  I'm planning to ship and use the `Start-EditorServices.ps1` script that way.

rkeithhill:
If the idea is to invoke it from the editor which, I believe, is your plan. Then it doesn't have to be discoverable like a command.  In fact, we don't want it to be discoverable in that case.  So yeah, a script would work.  I can rework this into a script.  Any suggestions on a name?  GetPsesVersionInfo.ps1?

daviwil:
I'm gonna go ahead and merge this so that I can roll it into an internal command module for PSES.

PR #405, opened 2017-03-25
Report column number in stack frame when debugger stops

daviwil:
This change causes the column number of the debugger's currently stopped
position to be passed through with the top stack frame.  This allows VS
Code's debugger UI to show an indicator that points to where the debugger
stopped when stepping through code.

Resolve PowerShell/vscode-powershell#616


rkeithhill:
Still a bummer we have to resort reflection in places.

daviwil:
Yeah, apparently there's a DisplayScriptPosition property on InvocationInfo but it returns null :(  Had to find this approach by looking at the ISE code.

PR #406, opened 2017-03-26
Fix issue w/Auto variables empty in script scope.

rkeithhill:
This doesn't fix the issue with all the global variables appearing in `Auto`.

Fixes PowerShell/vscode-powershell#614.


daviwil:
Since we're going to keep the current globals-in-auto behavior when the script is dot-sourced, can we consider this PR done?  We can do the launch config option in a separate PR.

rkeithhill:
Yes. I still can't believe I had the bit | vs bit & error.  Sheesh.

daviwil:
Hah!  It happens :)

PR #424, opened 2017-04-06
Remove extra prompt string added to interactive debugging session

### LINKED ISSUE #422: "Interactive Session" debugging mode is adding extra prompt to console

daviwil:
This change removes some code that was causing an extra command prompt
string to be printed to the integrated console when running an
interactive debugging session.  It also makes the termination of such a
session more reliable.

Resolves #422.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -26,6 +26,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
         private OutputDebouncer outputDebouncer;
 
         private bool noDebug;
+        private string arguments;
         private bool isRemoteAttach;
         private bool isAttachSession;
         private bool waitingForAttach;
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -303,13 +304,9 @@ protected override void Shutdown()
 
             await requestContext.SendResult(null);
 
-            // If no script is being launched, execute an empty script to
-            // cause the prompt string to be evaluated and displayed
-            if (string.IsNullOrEmpty(this.scriptToLaunch))
-            {
-                await this.editorSession.PowerShellContext.ExecuteScriptString(
-                    "", false, true);
-            }
```
rkeithhill: So what do we miss out on during an interactive session debug by not having the above code? An initial prompt string?

daviwil: The prompt has already been written by this point, so this line is totally pointless now.  It was necessary before the integrated console came along because a prompt was never displayed at first when the debugger starts, the assumption was that we were launching a script.  For an "interactive session" I needed to invoke a prompt to display, so this was a hacky way to do that :)

PR #425, opened 2017-04-06
How about popping UI to attach the debugger?

rkeithhill:
Any chance we can go back to this? This is nicer because it pops the attach a debugger UI.  You don't have to track down the PID.  You only have to select which instance of VS you want to use to debug.


daviwil:
Have you tried it with PowerShell Core?  I think that's the reason I went back to this approach.  I'll give it a shot later to check.

rkeithhill:
I built PSES in VSCode using the build task provided by vscode-powershell.  But since I develop on Windows, I suppose that just uses the desktop framework?  

Well if it doesn't work on .NET Core (although the API is doc'd for .NET Standard 1.6 - or are we on 1.3?) then I can `#if CoreCLR` in the old code and use the new code only for Windows.

PR #430, opened 2017-04-07
Temporarily mitigate intermittent IntelliSense slowness

daviwil:
This change mitigates seemingly intermittent IntelliSense slowness by
preventing Get-Command and Get-Help from being called on cmdlets from
the PowerShellGet and PackageManagement modules.  Something in the
PackageManagement module is causing completions to be returned at 10x
delay when that module is loaded.  If the module is removed from the
session, completions go back to their normal duration.

By avoiding calling Get-Command on these cmdlets, we avoid loading
PackageManagement when the user types a string like `Get-P` which
incidentally happens to make VS Code see the `Get-Package` cmdlet and
try to resolve its details.

This fix is temporary and will be removed in the future once
https://github.com/OneGet/oneget/issues/275 is fixed.

Related to PowerShell/vscode-powershell#647

### DIFF
--- a/src/PowerShellEditorServices/Language/CommandHelpers.cs
+++ b/src/PowerShellEditorServices/Language/CommandHelpers.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using Microsoft.PowerShell.EditorServices.Utility;
+using System.Collections.Generic;
 using System.Linq;
 using System.Management.Automation;
 using System.Threading.Tasks;
@@ -14,6 +16,20 @@ namespace Microsoft.PowerShell.EditorServices
     /// </summary>
     public class CommandHelpers
     {
+        private static HashSet<string> NounBlackList =
+            new HashSet<string>
+            {
+                "Module",
+                "Script",
+                "Package",
+                "PackageProvider",
+                "PackageSource",
+                "InstalledModule",
+                "InstalledScript",
+                "ScriptFileInfo",
+                "PSRepository"
+            };
+
         /// <summary>
         /// Gets the CommandInfo instance for a command with a particular name.
         /// </summary>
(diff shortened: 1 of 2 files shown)


daviwil:
Keith, do you think there's any risk of confusion to users by making this change?  I think the potential benefit of not having IntelliSense perf issues outweighs any possible discomfort from missing the cmdlet signatures for PowerShellGet and PackageManagement.

rkeithhill:
I don't notice much of a slowdown on my work PC.  Maybe 3-5 secs for the first time I try to complete Install-Package.  But this does seem to be affecting folks so we should try to address it.  I wonder if perhaps this black list should be an extension setting that gets passed into PSES though?  That way, folks could customize it.  Maybe I use PowerShellGet a lot and am willing to put up with the delay?  

The downside to the setting though is if this is a temporary glitch, it might become unnecessary in the future.

daviwil:
Yeah, I don't think that it should be configurable since it's temporary.  The problem is that once PackageManagement is loaded, IntelliSense is permanently slower until you remove that module.  Same issue in powershell.exe and PowerShell ISE.

The only degradation for PowerShellGet and PackageManagement is the loss of extra info in completion results, they'll still get completions for those commands though.

src/PowerShellEditorServices/Language/CommandHelpers.cs
```diff
@@ -14,6 +16,20 @@ namespace Microsoft.PowerShell.EditorServices
     /// </summary>
     public class CommandHelpers
     {
+        private static HashSet<string> NounBlackList =
+            new HashSet<string>
+            {
+                "Module",
```
rkeithhill: Does this mean we won't get info for Import-Module?  That is a pretty commonly used command.  The rest, less so.  But if it is temporary, then let's try it and see if folks complain about missing info for these nouns.  

I wonder if it would be worth returning some static text for these nouns like "Intellisense for this noun disabled due to perf issue, will be fixed in the future." or something like that?

daviwil: Probably not worth the extra effort at the moment, I imagine many people won't even notice it.  If people do notice and complain, I'll definitely find a way to make it a bit more obvious what's going on.

PR #436, opened 2017-04-11
Enable reference look up in entire workspace

kapilmb:
This PR adds the following two features:
* Enables looking up references in the entire workspace
* Enables looking up definitions in the entire workspace

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -679,7 +679,8 @@ function __Expand-Alias {
             FindReferencesResult referencesResult =
                 await editorSession.LanguageService.FindReferencesOfSymbol(
                     foundSymbol,
-                    editorSession.Workspace.ExpandScriptReferences(scriptFile));
+                    editorSession.Workspace.ExpandScriptReferences(scriptFile),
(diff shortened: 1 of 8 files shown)


daviwil:
Awesome!  I might not get to review this for a couple days, but I'm looking forward to it.

daviwil:
Just looked at this more thoroughly, looks great!  Not sure why the CI build seems to be hanging though.  I'm re-running it to see if I can figure out what's happening.

kapilmb:
I guess some tests were failing initially but then I think I fixed them. But I am not sure why I didn't remove the `WIP`. Before you merge it, let me see if I am not missing anything.

daviwil:
Something about the change is causing the Host assembly tests to hang, probably due to one of the existing find references tests hitting an issue with the new logic.  Have you tried running the host tests on your machine?  Make sure you run the LayoutModule task in the build script before running the host tests otherwise the module won't have the latest binaries (I need to fix that...)

kapilmb:
The newly added EnumeratePSFiles method was not checking for the validity of workspace root path. I have fixed it and all the tests pass on my computer. Hopefully it should be the same here too.

daviwil:
Excellent, it passed!  Thanks a lot for looking into that.

kapilmb:
@daviwil Thanks!

PR #445, opened 2017-04-27
Upgrade language server protocol to v3.2.0

kapilmb:

PR #447, opened 2017-05-03
Fix definition and reference request parameters

kapilmb:
These changes got left out of #445


daviwil:
Looks good, thanks for the fix!

PR #448, opened 2017-05-08
Move to netstandard2.0 to support PowerShell 6.0.0-beta1

daviwil:
This change moves the project to netstandard2.0 so that our module loads
and runs inside of PowerShell 6.0.0-beta1.  We will not maintain a
netstandard1.6 build at this time because PowerShell 6.0.0-alpha builds
will no longer be supported.


daviwil:
That was unbelievably painless...

daviwil:
Turns out that this may not even be necessary, double-checking...

daviwil:
Skipping this for now, seems like it's not needed yet!

PR #457, opened 2017-05-18
Release 1.1.0

daviwil:


daviwil:
Yeah, I keep putting off making a real solution for that :)

rkeithhill:
I'm guessing the tests file could probably use reflection to get the version number?  If so, perhaps I can implement that.

daviwil:
I think the best solution would be for it to read it from the .psd1 file since we're loading based on the module version.  Shouldn't be too hard to grab it out of the .psd1 contents with a regex

rkeithhill:
That'll work too.

PR #458, opened 2017-05-19
Fix finding variable definitions in workspace

kapilmb:
Resolves PowerShell/vscode-powershell#710

### DIFF
--- a/src/PowerShellEditorServices/Language/FindDeclartionVisitor.cs
+++ b/src/PowerShellEditorServices/Language/FindDeclartionVisitor.cs
@@ -14,25 +14,31 @@ namespace Microsoft.PowerShell.EditorServices
     internal class FindDeclartionVisitor : AstVisitor
     {
         private SymbolReference symbolRef;
+        private string variableName;
 
         public SymbolReference FoundDeclartion{ get; private set; }
 
         public FindDeclartionVisitor(SymbolReference symbolRef)
         {
             this.symbolRef = symbolRef;
+            if (this.symbolRef.SymbolType == SymbolType.Variable)
+            {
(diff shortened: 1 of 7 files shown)


src/PowerShellEditorServices/Language/FindDeclartionVisitor.cs
```diff
@@ -14,25 +14,31 @@ namespace Microsoft.PowerShell.EditorServices
     internal class FindDeclartionVisitor : AstVisitor
```
daviwil: Mind fixing the typo in the name of this class? Find*Declaration*Visitor.  That's bugged me for a while :)

kapilmb:
To take `scope` into account we have two options - 
1. implement our own ad-hoc logic OR
2. use the [*variable analysis*](https://github.com/PowerShell/PSScriptAnalyzer/blob/development/Engine/VariableAnalysis.cs) part in PSSA. 

The advantage with **2** is that we will have a robust solution and the techniques used there can then also be used in *rename* logic.

daviwil:
I'm all for using the existing variable analysis logic, maybe it's better to wait to do that until we can centralize our analysis and workspace logic in a reusable module

kapilmb:
The existing *variable analysis* logic is still somewhat buggy and needs some work. But I agree that we should centralize our analysis logic then use it from there.

src/PowerShellEditorServices/Language/FindDeclartionVisitor.cs
kapilmb: fixed!

daviwil:
Thanks dude!  Feel free to merge it once the tests pass

PR #484, opened 2017-06-02
Establish new channel connection model using server listeners

daviwil:
This change refactors our existing channel model to move all connection
logic outside of the ChannelBase implementations so that the
language and debug client/service pairs can be simplified.  This change
also allows us to remove a long-standing hack in our Host unit tests
which added an artifical delay to give the channel and MessageDispatcher
time to get established.

### DIFF
--- a/src/PowerShellEditorServices.Host/EditorServicesHost.cs
+++ b/src/PowerShellEditorServices.Host/EditorServicesHost.cs
@@ -3,7 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using Microsoft.PowerShell.EditorServices.Console;
+using Microsoft.PowerShell.EditorServices.Extensions;
+using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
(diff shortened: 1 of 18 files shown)


daviwil:
The basic idea here is that connections should be established before creating an instance of a ChannelBase implementation.  This simplifies client and server implementation since they don't have to deal with waiting for a connection before proceeding.  The assumption is that the ChannelBase implementation is connected and fully established before being passed to a ProtocolEndpoint implementation (like LanguageServiceClient, etc).

/cc @mrward @kamilkosek since it affects the LanguageClient class and how you establish a TcpSocketServerClient.  Check out this code for example on how to set up a LanguageServiceClient with the altered model:

https://github.com/daviwil/PowerShellEditorServices/blob/90afbeb689c7ddb8be360de8bbf14c7fd694725d/test/PowerShellEditorServices.Test.Host/LanguageServerTests.cs#L53

daviwil:
Not in a rush to merge this, I'll leave it open until later tomorrow in case anyone has questions or feedback!

PR #487, opened 2017-06-03
Add new Commands submodule to expose useful commands inside editor

### LINKED ISSUE #403: Stop loading extension commands script resource as a .ps1
This is causing trouble for users on machines with stricter group policy or AppLocker settings turned on.  Since we haven't signed this dynamically-generated script it throws errors on load.

/cc @dotps1

daviwil:
This change adds a new PowerShellEditorServices.Commands module that
gets loaded inside of the editor session's runspace to provide functions
and cmdlets which are useful when working with the $psEditor APIs.  This
will also be the place where built-in "editor commands" are registered
when the session initializes.

Resolves #403.
Resolves PowerShell/vscode-powershell#784.


daviwil:
Naming of the module, it's path, etc are not set in stone, we can tweak things later if people have ideas for improvement :)  Merging this now to get @SeeminglyScience going!

PR #488, opened 2017-06-03
Remove static Logger API in favor of ILogger instance usage

daviwil:
This is a pretty big change because it touches every usage of `Logger.Write` in the codebase and adds new parameters to pass around an ILogger instance to individual components.  This is another incremental change toward the new PSES-as-a-framework model so that third-party extension modules will be able to add logs to the EditorServices.log file.

### DIFF
--- a/src/PowerShellEditorServices.Host/EditorServicesHost.cs
+++ b/src/PowerShellEditorServices.Host/EditorServicesHost.cs
@@ -34,6 +34,7 @@ namespace Microsoft.PowerShell.EditorServices.Host
     {
         #region Private Fields
 
+        private ILogger logger;
         private bool enableConsoleRepl;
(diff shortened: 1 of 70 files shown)


daviwil:
Going to forge ahead on this because I've got some work to do tomorrow that depends on it!  Let me know if you have any questions or concerns.

rkeithhill:
I like this "DI" style approach.  It is what I use in my code at work where I often have a NullLogger when I don't care about log output.

daviwil:
Yep, the refactoring that I'm doing is gradually moving everything to a DI model, you can see more of that in my latest PR 😄  I like being able to throw a dummy implementation in where something isn't needed.  It'll definitely make some of the code easier to test.

PR #490, opened 2017-06-05
Provide pester functions (it, context, describe) in document symbols

kapilmb:
This provides pester symbols (it, context and describe) for a pester file whenever the server receives a `textDocument/documentSymbol` request. As such, it allows easier navigation of a pester test file.


daviwil:
Heh, oh man, perfect timing, I've got an awesome use for this ;)

src/PowerShellEditorServices/Language/LanguageService.cs
```diff
@@ -54,6 +55,12 @@ public LanguageService(PowerShellContext powerShellContext)
 
             this.CmdletToAliasDictionary = new Dictionary<String, List<String>>(StringComparer.OrdinalIgnoreCase);
             this.AliasToCmdletDictionary = new Dictionary<String, String>(StringComparer.OrdinalIgnoreCase);
+            this.documentSymbolProviders = new DocumentSymbolProvider[]
```
daviwil: Are you reading my mind?  I was just about to implement a generic "feature provider" model in certain parts of PSES :)

nightroman:
Is it something that others can use, too? I am thinking of Invoke-Build and psake `task` symbols, for example.

daviwil:
@nightroman it will be soon, I'm going to expose functionality like this through the $psEditor model.  Possibly even for this week's release :)

daviwil:
@nightroman by the way, having a built-in task symbols provider would be really nice, would you be interested in contributing one similarly to how Kapil added the Pester symbol provider?  What did you have in mind for how you'd use this?

nightroman:
@daviwil I am not completely sure yet what a symbol provider is. I would like to see it in action at first.

daviwil:
Ahhh, sorry, it's basically the concept that Kapil added to PSES in this PR: for a given ScriptFile, return SymbolReference instances which mark any special symbols in the file, like `Task` definitions.

daviwil:
You'd basically be implementing another one class like this that looks for `Task` instead of `Describe`, etc:

https://github.com/kapilmb/PowerShellEditorServices/blob/c6271653dce470319384974430ccc36f34cb1e22/src/PowerShellEditorServices/Language/PesterDocumentSymbolProvider.cs

In fact, it'd be mostly a copy/paste job ;)

nightroman:
> In fact, it'd be mostly a copy/paste job ;)

Oh, in that case, count me in!

I am still interested in the API for not built-in stuff, looking forward to 1.3.0.

kapilmb:
@nighitroman  When I wrote this I had `InvokeBuild`in mind but  wanted to test the waters first with `Pester` and if things go well then we could've proceeded with `InvokeBuild`. An approach similar to that of `TabExpansionPlusPlus`, such that any module can register their symbols providers would be ideal. But I couldn't think of any modules other than `Pester`, `InvokeBuild` and `psake` that use DSL-like syntax - I am sure there probably are more but I am not aware of them. I wanted an efficient way to navigate Pester test files and the current approach seemed to achieve that with least possible effort. However, with the extensibility model that David will soon be pushing, I think we can move towards the `ideal` approach sooner rather than later. But as @daviwil has already pointed out, please feel free to add `InvokeBuild` symbol providers.

daviwil:
I think just a general task symbol provider might be good enough to support both psake and Invoke-Build since they both use the `Task` command syntax, but Invoke-Build should also be able to provide its own for anything outside of that.

PR #492, opened 2017-06-05
Establish new host interface abstraction model, remove ConsoleService

daviwil:
This change is a fairly large refactoring of our PSHost implementation
to remove the ConsoleService and IConsoleHost types and instead use two
different PSHostUserInterface implementations.  Here's a detailed
breakdown of the changes:

- Centralized host UI behavior in EditorServicesPSHostUserInterface
  abstract class.  This class now exposes overridable abstract methods
  that simplify PSHostUserInterface implementations that will be used
  with the EditorServicesPSHost.

- Removed the ConsoleService class and put its behavior and abstractions
  into EditorServicesPSHostUserInterface.

- Created TerminalPSHostUserInterface for the integrated terminal host
  experience.

- Created ProtocolPSHostUserInterface for the protocol-based host
  experience.

- Removed the concept of a "prompt handler context" because each host
  implementation will have a single way to deal with input and choice
  prompts.

- Lifted direct management of the console interface out of the
  LanguageServer and DebugAdapter classes, now managed in the
  EditorServicesHost.

- Disabled the ConsoleServiceTests until we decide how to test for
  this behavior at the level of the EditorServicesPSHostUserInterface.


daviwil:
### Work remaining

- [ ] Make sure that debug variables and watches still get evaluated correctly
- [ ] Investigate issue with pipeline exception when debugging a remote script

PR #496, opened 2017-06-08
Add position function set to Commands

SeeminglyScience:
This provides a set of functions to assist with finding and manipulating positions in script files.  This also includes the `Import-EditorCommand` which will register any function from a module that has the `EditorCommandAttribute` attached to it.

This is the full list of added commands:

- Find-Ast
- Get-Token
- ConvertFrom-ScriptExtent
- ConvertTo-ScriptExtent
- Set-ScriptExtent
- Join-ScriptExtent
- Test-ScriptExtent
- Import-EditorCommand

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ PowerShellEditorServices.sln.ide/edb.chk
 PowerShellEditorServices.sln.ide/edbres00001.jrs
 PowerShellEditorServices.sln.ide/storage.ide
 *.jrs
+
+# Don't include PlatyPS generated MAML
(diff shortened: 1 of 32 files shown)


daviwil:
Awesome!  Thanks a lot for sending this, I'll give it a full review later.  Looks like AppVeyor needs to have PlatyPS installed, you can add the Install-Module line for that to appveyor.yml.

daviwil:
Looks like the Travis init script also needs to be updated, sorry I forgot to mention that!

https://github.com/PowerShell/PowerShellEditorServices/blob/master/scripts/travis.ps1

PR #499, opened 2017-06-08
Fix before switch parameter for Find-Ast

SeeminglyScience:
This fixes an issue where the before switch wasn't searching in the correct order.


daviwil:
Sweet, thanks man!

PR #504, opened 2017-06-10
Fix variable's 'add to watch' feature.

rkeithhill:
Fix #838 in vscode-powershell

### DIFF
--- a/src/PowerShellEditorServices/Debugging/DebugService.cs
+++ b/src/PowerShellEditorServices/Debugging/DebugService.cs
@@ -603,7 +603,7 @@ namespace Microsoft.PowerShell.EditorServices
             // of command executions into string output.  However, if null is returned
             // then return null so that no output gets displayed.
             string outputString =
-                results != null ?
+                results != null && results.Count() > 0 ?
                     string.Join(Environment.NewLine, results) :
                     null;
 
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Debugging/DebugService.cs
```diff
@@ -603,7 +603,7 @@ public VariableDetailsBase GetVariableFromExpression(string variableExpression,
             // of command executions into string output.  However, if null is returned
             // then return null so that no output gets displayed.
             string outputString =
-                results != null ?
+                results != null && results.Count() > 0 ?
```
daviwil: Possibly more efficient to use results.Any() here?

rkeithhill: Um ... yeah, good call!  Much quicker than to count a 10000 element sequence.

rkeithhill:
Before your merge a PR do you assign it a milestone or do you do that just for the related issue?

daviwil:
I've been assigning milestones for both issues and PRs lately so I can keep track of what work got done.  In this case since we don't have a new version number milestone yet we can just assign it to the June milestone.  Once I'm ready to ship an update I scan through the closed items for the current month milestone and assign them to the specific version milestone.

I should write up something about the milestone model since it's gone through several variations before I settled on this ;)

PR #508, opened 2017-06-11
Fix for vscode-powershell #837 callstack missing line info. 

rkeithhill:
This PR addresses two bugs.  First the missing line info was due to use passing back null when we didn't have end line/col info.  The protocol claims that should work.  But when I changed the end line/col to reflect the same values as the start line/col (assuming the end values were null - not set) then the callstack started to display the line info.  It is  a bit weird but the very first frame in some cases contains a col other than 0.

The second bug was that the call stack was getting duplicated.  I implemented the new stackTraceRequest for paging stack frames back to VSCode and now that bug is fixed.

I tried to implement StackFrame.PresentationHint == "subtle" for script outside the initial working directory.  Do you have a better way to get the workspaceRoot path in the debugger?  I had to cache the working dir when SetWorkingDirectory() gets called.  And when you attach to a process, there isn't a way to set the "cwd".

Finally, I ran into one bug I'll file later that has to do with the base of the call stack sometimes showing "<ScriptBlock> <No File> 1".  If you click on that you get an error telling you <No File> can't be opened.

### DIFF
--- a/src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
+++ b/src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
@@ -3,10 +3,30 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
 using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices
 {
+    public enum StackFramePresentationHint
+    {
+        /// <summary>
+        /// Dispays the stack frame as a normal stack frame.
+        /// </summary>
+        Normal,
+
+        /// <summary>
+        /// Used to label an entry in the call stack that doesn't actually correspond to a stack frame.
+        /// This is typically used to label transitions to/from "external" code.
+        /// </summary>
+        Label,
+
+        /// <summary>
+        /// Displays the stack frame in a subtle way, typically used from loctaions outside of the current project or workspace.
+        /// </summary>
+        Subtle
+    }
+
     /// <summary>
     /// Contains details pertaining to a single stack frame in
     /// the current debugging session.
@@ -43,7 +63,7 @@ namespace Microsoft.PowerShell.EditorServices
         /// <summary>
         /// Gets the line number of the script where the stack frame occurred.
         /// </summary>
-        public int EndLineNumber { get; internal set; }
+        public int? EndLineNumber { get; internal set; }
 
         /// <summary>
         /// Gets the start column number of the line where the stack frame occurred.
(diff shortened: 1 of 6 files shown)


src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
```diff
@@ -43,7 +63,7 @@ public class StackFrameDetails
         /// <summary>
         /// Gets the line number of the script where the stack frame occurred.
         /// </summary>
-        public int EndLineNumber { get; internal set; }
+        public int? EndLineNumber { get; internal set; }
```
rkeithhill: Considering that Get-PSCallStack doesn't return any info about end line/col it seems like these values should be nullable (even though I'm setting them to startLine / 0 respectively).

daviwil:
Looks like there are a couple missing XML comments causing the build to fail:

```
Debugging\StackFrameDetails.cs(11,17): error CS1591: Missing XML comment for publicly visible type or member 'StackFramePresentationHint' [C:\projects\powershelleditorservices\src\PowerShellEditorServices\PowerShellEditorServices.csproj]
Debugging\StackFrameDetails.cs(115,20): error CS1573: Parameter 'workspaceRootPath' has no matching param tag in the XML comment for 'StackFrameDetails.Create(PSObject, VariableContainerDetails, VariableContainerDetails, string)' (but other parameters do) [C:\projects\powershelleditorservices\src\PowerShellEditorServices\PowerShellEditorServices.csproj]
```

src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
daviwil: Yep, it's clearer if the contract is that those values may not be populated.

src/PowerShellEditorServices/Debugging/StackFrameDetails.cs
```diff
@@ -86,16 +111,34 @@ public class StackFrameDetails
...
             PSObject callStackFrameObject,
             VariableContainerDetails autoVariables,
-            VariableContainerDetails localVariables)
+            VariableContainerDetails localVariables,
+            string workspaceRootPath = null)
         {
+            string moduleId = string.Empty;
+            var presentationHint = StackFramePresentationHint.Normal;
+
+            var invocationInfo = callStackFrameObject.Properties["InvocationInfo"].Value as InvocationInfo;
```
daviwil: Have you tested this with remote debugging?  Sometimes these objects come back as a weird serialized type that will cause issues.

src/PowerShellEditorServices/Session/PowerShellContext.cs
```diff
@@ -103,6 +103,12 @@ public RunspaceDetails CurrentRunspace
             private set;
         }
 
+        /// <summary>
+        /// Gets the working directory path the PowerShell context was inititially set when the debugger launches.
+        /// This path is used to determine whether a script in the call stack is an "external" script.
+        /// </summary>
+        public string InitialWorkingDirectory { get; private set; }
```
daviwil: This is fine for now, we'll come up with a better way to get the workspace path soon.

daviwil:
Thanks for implementing the new stack frame behavior too!  Feel free to merge after the XML comment issues are fixed.

daviwil:
Thanks man!

daviwil:
Fixes https://github.com/PowerShell/vscode-powershell/issues/837

PR #516, opened 2017-06-13
Add request handlers for document formatting

kapilmb:


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -200,7 +204,8 @@ protected Task Stop()
                         SignatureHelpProvider = new SignatureHelpOptions
                         {
                             TriggerCharacters = new string[] { " " } // TODO: Other characters here?
-                        }
+                        },
+                        DocumentFormattingProvider = false
```
daviwil: Should this be `true`?  Or are you doing this as a workaround to the FormattingOptions issue?

daviwil: I saw the answer to my question in the other PR :)

daviwil:
Missing a couple of XML comments, otherwise looks good!

```
Analysis\AnalysisService.cs(256,23): error CS1573: Parameter 'settings' has no matching param tag in the XML comment for 'AnalysisService.Format(string, Hashtable, int[])' (but other parameters do) [C:\projects\powershelleditorservices\src\PowerShellEditorServices\PowerShellEditorServices.csproj]
Analysis\AnalysisService.cs(257,19): error CS1573: Parameter 'rangeList' has no matching param tag in the XML comment for 'AnalysisService.Format(string, Hashtable, int[])' (but other parameters do) [C:\projects\powershelleditorservices\src\PowerShellEditorServices\PowerShellEditorServices.csproj]
```

daviwil:
Could you also remove the old formatting message type handlers you were using if they aren't being used anymore?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
kapilmb: Yes, this is part of the workaround - when I started working on this I had set it to `true`, only to run into the `FormattingOptions` issue later. So I set to to false, which is redundant as the it would be `false` by default.

kapilmb:
Fixed both the issues!

daviwil:
Something weird is going on with the tests...  They're failing in ways that should be irrelevant to your changes.  I'll clone your code and see if I can reproduce the issue locally but I may not get to it until tomorrow morning.  In the meantime you could try running the Host tests on your machine to see if you see an issue and then check the logs in `.\PowerShellEditorServices\test\PowerShellEditorServices.Test.Host\bin\Debug\net452\logs\LanguageServerTests`

daviwil:
Great, thanks for fixing the tests!  Didn't think that this would be the problem, good to know there was an explanation.

PR #517, opened 2017-06-14
Added NewFile() for $psEditor and new functions

dfinke:


daviwil:
Awesome, thanks for sending this!

I'd say leave out the Get-GitLog function since it is a little too special case for this module.  Works well as an example but not something I'd include as part of the cmdlet API.  Also, you'll need to add an XML comment for the IEditorOperations.NewFile method to make the compiler happy :)  Other than that, looks good!

PR #522, opened 2017-06-17
Add Stroustrup formatting style.

rkeithhill:
I think folks are going to like this preset option.  Pretty slick @kapilmb !!


daviwil:
Yep, it'll be a major improvement.  No more tweaking individual settings!

PR #523, opened 2017-06-20
Add custom HTML views to VSCode module

daviwil:
This change adds support for custom HTML views in the VSCode module
which forms the basis for UI extensions in the PowerShell extension in
Visual Studio Code.  This change includes both .NET PowerShell
cmdlet APIs for interacting with custom HTML UI in the editor.

### DIFF
--- a/module/PowerShellEditorServices.VSCode/PowerShellEditorServices.VSCode.psd1
+++ b/module/PowerShellEditorServices.VSCode/PowerShellEditorServices.VSCode.psd1
@@ -69,7 +69,11 @@ Description = 'Provides added functionality to PowerShell Editor Services for th
 # NestedModules = @()
 
 # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
-FunctionsToExport = @()
+FunctionsToExport = @('New-VSCodeHtmlContentView',
+                      'Show-VSCodeHtmlContentView',
+                      'Close-VSCodeHtmlContentView',
+                      'Set-VSCodeHtmlContentView',
+                      'Write-VSCodeHtmlContentView')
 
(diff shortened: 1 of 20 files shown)


daviwil:
I'm not fully done here yet, I think I need to rethink the HtmlContentView/HtmlOutputView split because there's a lot of unnecessary API duplication there.  Any thoughts?

I'll send the vscode-powershell PR after I've firmed this one up.

daviwil:
A few firsts in this change:

- This is the first time a new feature component has been added from an extension module
- This is the first time that new message types have been added from an extension module
- This is also the first attempt at making a cmdlet wrapper on top of the .NET APIs instead of exposing the .NET APIs directly for PowerShell scripting

Would love some feedback on any of these since I'm still trying to formalize the pattern.

dfinke:
I like the sound of a message type being added from an extension module

SeeminglyScience:
Great job with the wrappers, these will be a lot easier to discover from PowerShell.

daviwil:
What are your thoughts on using a `DefaultCommandPrefix` of `VSCode`?  This means that all cmdlets like `New-HtmlContentView` will turn into `New-VSCodeHtmlContentView` when imported.  Even though prefixing is considered somewhat of a best practice, I feel like it makes the cmdlet names a lot clunkier and harder to type.  Thoughts?

dfinke:
Yeah, clunkier. I do like the option of being able to do `gcm *-vsc*` and seeing all the cmdlets I can use.

daviwil:
The ISE only exposes 3 cmdlets but they're all prefixed "Ise", so maybe prefixing is generally the right thing to do.  I'll hardcode the prefixes for now instead of using `DefaultCommandPrefix` so that I have the opportunity to change the name later and use aliases for the old prefixed versions.  Thanks Doug!

PR #533, opened 2017-06-23
Implement GetCommands() on $psEditor

kamilkosek:
With this method you are able to receive all registered EditorCommand´s, and then unregister them or expose all commands through the HTML-ContentView and and be able to execute them once [this](https://github.com/PowerShell/vscode-powershell/issues/910) is implemented.


daviwil:
Awesome, thanks Kamil!

PR #534, opened 2017-06-26
Fix vscode-powershell issue 872 - watch expandable variables can't be…

rkeithhill:
… expanded.

I can't remember if this used to work or not.  But the `GetVariableFromExpression` had no callers which is suspicious.  Anyway, I think this should fix the issue.  You'll see my note that the way 'GetVariableFromExpression` works isn't perhaps 100% foolproof but is probably good enough.


daviwil:
Looks good to me, thanks man!

PR #535, opened 2017-06-26
Add Type field for Variables.

rkeithhill:
This implements a debug protocol feature that was added a while ago (but after we did the majority of work on the Variables view).  This will display the variables type if you hover over it e.g.:

image

The screen shot didn't capture my cursor.  I'm hovering over the `$InformationPreference` variable.


daviwil:
Awesome, thanks!

PR #519, opened 2017-06-15
Add Diagnostic logging level for protocol message JSON

daviwil:
This change adds a new "Diagnostic" logging level which is used for
logging the JSON contents of protocol messages that are read or written.
This will provide two benefits:

1. Increase the privacy of user logs that are transmitted over GitHub
2. Increase the performance of writing Verbose logs

We still log the high-level message details like message type, method,
and ID so that we can keep track of the message sequence.

PR #536, opened 2017-06-29
Add "Diagnostic" to ValidateSet for LogLevel

kamilkosek:
Since the _Diagnostic_ LogLevel was introduced [here](https://github.com/PowerShell/PowerShellEditorServices/pull/519) it wasn´t possible to use it due to the ValidateSet in the scripts.


daviwil:
Thanks Kamil!  I just noticed last night it was missing :)

PR #537, opened 2017-07-06
Fix stack trace so it shows external code as "deemphasized".

rkeithhill:


src/PowerShellEditorServices.Protocol/DebugAdapter/StackFrame.cs
```diff
@@ -66,13 +69,35 @@ public class StackFrame
...
+        /// <summary>
+        /// Used to label an entry in the call stack that doesn't actually correspond to a stack frame.
+        /// This is typically used to label transitions to/from "external" code.
+        /// </summary>
+        Label,
+
+        /// <summary>
+        /// Displays the stack frame in a subtle way, typically used from loctaions outside of the current project or workspace.
+        /// </summary>
+        Subtle
```
daviwil: Should this be Deemphasize too or are they inconsistent?

rkeithhill: Weird as it seems, the PresentationHint on StackFrame has different values than the PresentationHint on Source.  https://github.com/Microsoft/vscode-debugadapter-node/blob/master/protocol/src/debugProtocol.ts#L1131

That accounts for part of the confusion I was running into.  And then there was also the typo in the VSCode impl.  But this works with 1.13.1 so yay!

PR #540, opened 2017-07-11
Add pipeline input to Write-VSCodeHtmlContentView cmdlet

daviwil:
This change adds the ValueFromPipeline setting to the
AppendedHtmlBodyContent parameter of the Write-VSCodeHtmlContentView
cmdlet so that pipeline output can be written out to an HtmlContentView
with ease.

Resolves PowerShell/vscode-powershell#909.


stefanstranger:
Thanks for implementing this.

PR #541, opened 2017-07-15
Fix ConvertTo-ScriptExtent swapping start line and column

SeeminglyScience:
- Fixed an issue where `ConvertTo-ScriptExtent` could reverse starting line and column in some situations.


daviwil:
Whoops :)  Thanks man!

PR #547, opened 2017-08-22
Update to .NET SDK 2.0

daviwil:
This change updates the build script to use the newly released .NET SDK
2.0.  dotnet CLI 2.0 now restores package dependencies implicitly so all
restore logic has been taken out of the build script.  The test projects
have also been updated to use xUnit 2.3 and their new `dotnet xunit`
runner.


daviwil:
Though I should mention that this doesn't change our assemblies to build against netstandard2.0, I tried that and had some trouble building against multiple PowerShell versions.  I'll try it again soon, though.

PR #560, opened 2017-10-26
Fix issues with breakpoint synchronization and transcribed sessions

daviwil:


daviwil:
Strangely there's some issues with the breakpoint tests inside the Test.Host assembly.  I think that the DebugAdapterClient isn't initializing the debug session in the right order, causing the breakpoints to be cleared out before script execution starts.  I'll have to look into this before I can do a release build...

PR #561, opened 2017-10-27
Don't attempt to import DSC module on PowerShell core

rkeithhill:
Also, use env var for ProgramFiles location.


daviwil:
Thanks Keith!

PR #568, opened 2017-11-04
Fix DirectoryNotFoundException user is seeing in deep node_modules dir

rkeithhill:
Enumerating files in a dir shouldn't crash PSES for security, path length and errors from dir not found.


daviwil:
Nice fix, thanks man!

Issue #567, opened 2017-11-02
Microsoft.PowerShell.EditorServices.ScriptFile.ValidatePosition out of range


codykonior:
Hi,

I'm using VS Code 1.17.2, VS Code PowerShell 1.5.0, which bundles 1.5.0 of this module. When I trigger debugging, my debug sessions often terminate unexpectedly along with my PowerShell terminals. This is on Windows Server 2012 R2.

```
Name                           Value
----                           -----
PSVersion                      5.0.10586.117
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.10586.117
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

Here's the relevant part of the logs (there's some lead up but it's at the end):

```
2/11/2017 3:03:14 PM [VERBOSE] - Method "ExecuteCommand" at line 491 of C:\projects\powershelleditorservices\src\PowerShellEditorServices\Session\PowerShellContext.cs

    Attempting to execute command(s):
    
        Microsoft.PowerShell.Utility\Set-PSBreakpoint
    

2/11/2017 3:03:14 PM [VERBOSE] - Method "OnSessionStateChanged" at line 1138 of C:\projects\powershelleditorservices\src\PowerShellEditorServices\Session\PowerShellContext.cs

    Session state changed --
    
        Old state: Ready
        New state: Running
        Result: NotFinished

2/11/2017 3:03:14 PM [VERBOSE] - Method "OnSessionStateChanged" at line 1138 of C:\projects\powershelleditorservices\src\PowerShellEditorServices\Session\PowerShellContext.cs

    Session state changed --
    
        Old state: Running
        New state: Ready
        Result: Completed

2/11/2017 3:03:14 PM [VERBOSE] - Method "ExecuteCommand" at line 551 of C:\projects\powershelleditorservices\src\PowerShellEditorServices\Session\PowerShellContext.cs

    Execution completed successfully.

2/11/2017 3:03:14 PM [VERBOSE] - Method "WriteMessage" at line 61 of C:\projects\powershelleditorservices\src\PowerShellEditorServices.Protocol\MessageProtocol\MessageWriter.cs

    Writing Response 'setBreakpoints' with id 9

[...]

PR #573, opened 2017-11-05
Fix off-by-one on maxlLine calc

rkeithhill:
The old code allowed the line param to be *equal* to FileLines.Count + 1.
That's not a valid index even after subtracting 1 to get to a 0-based index.
Also added more data to log messages to help in the future.

@daviwil I pretty sure about the off-by-one but not 100% sure.  Can you double check?  At the very least, I'd like to enhance the log messages.

This is related to #567.  I'm not sure if this fixes it (kind of doubt it) but I'm hopeful the enhanced log messages might help if someone repro's this.

### DIFF
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -279,12 +279,13 @@ namespace Microsoft.PowerShell.EditorServices
         /// <param name="column">The 1-based column to be validated.</param>
         public void ValidatePosition(int line, int column)
         {
-            if (line < 1 || line > this.FileLines.Count + 1)
+            int maxLine = this.FileLines.Count;
+            if (line < 1 || line > maxLine)
             {
-                throw new ArgumentOutOfRangeException("Position is outside of file line range.");
+                throw new ArgumentOutOfRangeException($"Position {line}:{column} is outside of the line range of 1 to {maxLine}.");
             }
 
-            // The maximum column is either one past the length of the string
+            // The maximum column is either **one past** the length of the string
             // or 1 if the string is empty.
             string lineString = this.FileLines[line - 1];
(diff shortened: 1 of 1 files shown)


rkeithhill:
So how are you running the unit tests these days?  I can only build PSES by building the extension from VSCode.  When I try to run the unit tests in VS, it tries to build and the build fails (even though it builds from VSCode).

daviwil:
@rkeithhill I use `Invoke-Build Test[Host|Server|Protocol]?` to run them, the project build is a little more complex and is driven by the build script these days.

daviwil:
You may be right with this fix, actually.  I've seen some instances recently where out of range exceptions get thrown when working with file positions, maybe this was it.  Can you still get IntelliSense in the last line of a file with this fix?

rkeithhill:
That's better than what I've tried - installing xunit console runner from NuGet and invoking it directly.  :-)

RE - Intellisense on the last line - I'll check that tonight.

rkeithhill:
OK, yeah auto-completion works on the last line, last column. BTW I see this in the VS output window while I'm debugging.
```
Dispatching event "$/cancelRequest"
Dispatching message "$/cancelRequest" HAS NO HANDLER TO AWAIT
```
Is that harmless or ...?

rkeithhill:
OK I think I fixed the issue the test found.  Stupid mistake on my part.  The builds appear to be hung. Is there a way you can kick or restart them?

rkeithhill:
@daviwil Can you kick the Travis build?  It passed on Linux but failed to clone the repo on macOS.  I think this PR is good to go but it would be nice to get a clean pass through the tests.

daviwil:
Just restarted the build!

daviwil:
Travis CI is super slow and unreliable today for macOS builds...

daviwil:
It's building fine on Linux, I'd say go ahead and merge it even though the macOS side isn't running (their fault).

PR #579, opened 2017-11-17
Fix Get-Help -ShowWindows on Windows to appear on top of VSCode

rkeithhill:
BTW it seems the VSCode extension host returns a bogus VSCODE_PID env var.  :-(  However, when I plugged in the real PID in the debugger - it works.


rkeithhill:
Hold off on merge of this.  Need to do some more testing.

rkeithhill:
Aw crap, the [code](https://github.com/PowerShell/PowerShell/blob/52569380c7d2bcf4060a6618e3cf81dcd320041c/src/Microsoft.PowerShell.GraphicalHost/commandHelpers/ShowCommandHelper.cs#L720-L736) that asks for the PrivateData["Window"] property is expecting a WPF Window object - not a handle.  This looks to be very specific to ISE:
```
Window                                    : Microsoft.Windows.PowerShell.Gui.Internal.MainWindow
```

PR #580, opened 2017-11-24
Fix crash of PSES on startup when workspace folder has [] in path

rkeithhill:
This fixes https://github.com/PowerShell/vscode-powershell/issues/1014


daviwil:
Is there a place this method is being called where we would pass true for that parameter? If not, maybe just make path escaping be the default behavior?

rkeithhill:
@daviwil  Not in our code base.  I almost made the path escaping the default but the API is public and that would have been a breaking change.  That said, I'm not sure we have any external users yet, other than maybe @adamdriscoll.  What do you think? Should I escape by default and leave the second overload for anyone who doesn't want the path escaped (i.e. it's already escaped).

daviwil:
I don't think anyone external would be using it, but I agree that maybe we should retain the current behavior by default until we make a 2.0 of the .NET APIs. I'd say leave it how you've got it for now.

adbertram:
Is there a way to implement this into 1.5.1 rather than waiting for the release? This is a big thing for me.

rkeithhill:
You could build your own version of PSES and copy it over the one that comes with the extension.

PR #607, opened 2018-01-09
Found GetStringForPSCommand - moved parameter dump code into it

rkeithhill:
I thought I had looked for an existing impl of this but somehow missed `GetStringForPSCommand()`.  Modified it to render parameter info and use it from `ExecuteCommand` method.  

Actually,  you'd think this ability to render the complete command to a string would be built into the PSCommand type - oh well.


daviwil:
Sorry I missed this earlier this week, looks great!

rkeithhill:
No problem.  Everybody's busy.  I've been real busy with Keith Dahlby getting a 1.0.0-beta1 version of posh-git out to coincide with the PS Core 6 GA release yesterday.   Since have a bit of work to do on the README for that project.

PR #647, opened 2018-03-29
Change PSES to be buildable as a standalone

rjmholt:
When PSES builds, it should be responsible for its own modules. So I've added a build rule that populates its `/module/` directory. This moves the `modules.json` file over from the VSCode extension as well.

So with this, PSES should be generally self contained and ready to run after building.

### DIFF
--- a/PowerShellEditorServices.build.ps1
+++ b/PowerShellEditorServices.build.ps1
@@ -5,7 +5,13 @@
 
 param(
     [ValidateSet("Debug", "Release")]
-    [string]$Configuration = "Debug"
+    [string]$Configuration = "Debug",
+
+    [string]$PsesSubmodulePath = "$PSScriptRoot/module",
+
+    [string]$ModulesJsonPath = "$PSScriptRoot/modules.json",
+
+    [string]$DefaultModuleRepository = "PSGallery"
 )
 
 #Requires -Modules @{ModuleName="InvokeBuild";ModuleVersion="3.2.1"}
@@ -13,6 +19,7 @@ param(
 $script:IsCIBuild = $env:APPVEYOR -ne $null
 $script:IsUnix = $PSVersionTable.PSEdition -and $PSVersionTable.PSEdition -eq "Core" -and !$IsWindows
 $script:TargetFrameworksParam = "/p:TargetFrameworks=\`"$(if (!$script:IsUnix) { "net451;" })netstandard1.6\`""
+$script:SaveModuleSupportsAllowPrerelease = (Get-Command Save-Module).Parameters.ContainsKey("AllowPrerelease")
 
 if ($PSVersionTable.PSEdition -ne "Core") {
     Add-Type -Assembly System.IO.Compression.FileSystem
@@ -179,7 +186,7 @@ task TestProtocol -If { !$script:IsUnix} {
 task TestHost -If { !$script:IsUnix} {
     Set-Location .\test\PowerShellEditorServices.Test.Host\
     exec { & $script:dotnetExe build -c $Configuration -f net452 }
-    exec { & $script:dotnetExe xunit -configuration $Configuration -framework net452 -verbose -nobuild -x86 }
+    exec { & $script:dotnetExe xunit -configuration $Configuration -framework net452 -verbose -nobuild }
 }
 
 task CITest ?Test, {
@@ -218,6 +225,67 @@ task LayoutModule -After Build {
     if (!$script:IsUnix) {
         Copy-Item -Force -Path $PSScriptRoot\src\PowerShellEditorServices.VSCode\bin\$Configuration\net451\* -Filter Microsoft.PowerShell.EditorServices.VSCode*.dll -Destination $PSScriptRoot\module\PowerShellEditorServices.VSCode\bin\Desktop\
     }
+}
+
+task RestorePsesModules -After Build {
+    $submodulePath = (Resolve-Path $PsesSubmodulePath).Path + [IO.Path]::DirectorySeparatorChar
+    Write-Host "`nRestoring EditorServices modules..."
+
+    # Read in the modules.json file as a hashtable so it can be splatted
+    $moduleInfos = @{}
+
+    (Get-Content -Raw $ModulesJsonPath | ConvertFrom-Json).PSObject.Properties | ForEach-Object {
+        $name = $_.Name
+        $body = @{
+            Name = $name
+            MinimumVersion = $_.Value.MinimumVersion
+            MaximumVersion = $_.Value.MaximumVersion
+            Repository = if ($_.Value.Repository) { $_.Value.Repository } else { $DefaultModuleRepository }
+            Path = $submodulePath
+        }
+
+        if (-not $name)
+        {
+            throw "EditorServices module listed without name in '$ModulesJsonPath'"
+        }
+
+        if ($script:SaveModuleSupportsAllowPrerelease)
+        {
+            $body += @{ AllowPrerelease = $_.Value.AllowPrerelease }
+        }
+
+        $moduleInfos.Add($name, $body)
+    }
+
+    # Save each module in the modules.json file
+    foreach ($moduleName in $moduleInfos.Keys)
+    {
+        if (Test-Path -Path (Join-Path -Path $submodulePath -ChildPath $moduleName))
+        {
+            Write-Host "`tModule '${moduleName}' already detected. Skipping"
+            continue
+        }
+
+        $moduleInstallDetails = $moduleInfos[$moduleName]
+
+        $splatParameters = @{
+           Name = $moduleName
+           MinimumVersion = $moduleInstallDetails.MinimumVersion
+           MaximumVersion = $moduleInstallDetails.MaximumVersion
+           Repository = if ($moduleInstallDetails.Repository) { $moduleInstallDetails.Repository } else { $DefaultModuleRepository }
+           Path = $submodulePath
+        }
+
+        if ($script:SaveModuleSupportsAllowPrerelease)
+        {
+            $splatParameters += @{ AllowPrerelease = $moduleInstallDetails.AllowPrerelease }
+        }
+
+        Write-Host "`tInstalling module: ${moduleName}"
+
+        Save-Module @splatParameters
+    }
+    Write-Host "`n"
 }
 
 task BuildCmdletHelp {
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -13,6 +13,7 @@ using System.Management.Automation;
 using System.Collections.Generic;
 using System.Text;
 using System.Collections;
+using System.IO;
 
 namespace Microsoft.PowerShell.EditorServices
 {
@@ -26,17 +27,12 @@ namespace Microsoft.PowerShell.EditorServices
 
         private const int NumRunspaces = 1;
 
-        private ILogger logger;
-        private RunspacePool analysisRunspacePool;
-        private PSModuleInfo scriptAnalyzerModuleInfo;
+        private const string PSSA_MODULE_NAME = "PSScriptAnalyzer";
 
-        private bool hasScriptAnalyzerModule
-        {
-            get
-            {
-                return scriptAnalyzerModuleInfo != null;
-            }
-        }
+        private ILogger _logger;
+        private RunspacePool _analysisRunspacePool;
+
+        private bool _hasScriptAnalyzerModule;
 
         private string[] activeRules;
         private string settingsPath;
@@ -108,25 +104,30 @@ namespace Microsoft.PowerShell.EditorServices
         /// <param name="logger">An ILogger implementation used for writing log messages.</param>
         public AnalysisService(string settingsPath, ILogger logger)
         {
-            this.logger = logger;
+            this._logger = logger;
 
             try
             {
                 this.SettingsPath = settingsPath;
 
-                scriptAnalyzerModuleInfo = FindPSScriptAnalyzerModule(logger);
-                var sessionState = InitialSessionState.CreateDefault2();
-                sessionState.ImportPSModulesFromPath(scriptAnalyzerModuleInfo.ModuleBase);
+                if (!(_hasScriptAnalyzerModule = VerifyPSScriptAnalyzerAvailable()))
+                {
+                    throw new Exception("PSScriptAnalyzer module not available");
+                }
+
+                // Create a base session state with PSScriptAnalyzer loaded
+                InitialSessionState sessionState = InitialSessionState.CreateDefault2();
+                sessionState.ImportPSModule(new [] { PSSA_MODULE_NAME });
 
                 // runspacepool takes care of queuing commands for us so we do not
                 // need to worry about executing concurrent commands
-                this.analysisRunspacePool = RunspaceFactory.CreateRunspacePool(sessionState);
+                this._analysisRunspacePool = RunspaceFactory.CreateRunspacePool(sessionState);
 
                 // having more than one runspace doesn't block code formatting if one
                 // runspace is occupied for diagnostics
-                this.analysisRunspacePool.SetMaxRunspaces(NumRunspaces);
-                this.analysisRunspacePool.ThreadOptions = PSThreadOptions.ReuseThread;
-                this.analysisRunspacePool.Open();
+                this._analysisRunspacePool.SetMaxRunspaces(NumRunspaces);
+                this._analysisRunspacePool.ThreadOptions = PSThreadOptions.ReuseThread;
+                this._analysisRunspacePool.Open();
 
                 ActiveRules = IncludedRules.ToArray();
                 EnumeratePSScriptAnalyzerCmdlets();
@@ -137,7 +138,7 @@ namespace Microsoft.PowerShell.EditorServices
                 var sb = new StringBuilder();
                 sb.AppendLine("PSScriptAnalyzer cannot be imported, AnalysisService will be disabled.");
                 sb.AppendLine(e.Message);
-                this.logger.Write(LogLevel.Warning, sb.ToString());
+                this._logger.Write(LogLevel.Warning, sb.ToString());
             }
         }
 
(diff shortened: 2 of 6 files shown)


PowerShellEditorServices.build.ps1
```diff
@@ -5,19 +5,48 @@
...
 $script:IsCIBuild = $env:APPVEYOR -ne $null
 $script:IsUnix = $PSVersionTable.PSEdition -and $PSVersionTable.PSEdition -eq "Core" -and !$IsWindows
 $script:TargetFrameworksParam = "/p:TargetFrameworks=\`"$(if (!$script:IsUnix) { "net451;" })netstandard1.6\`""
+$script:SaveModuleSupportsAllowPrerelease = (Get-Command Save-Module).Parameters.ContainsKey("AllowPrerelease")
 
 if ($PSVersionTable.PSEdition -ne "Core") {
     Add-Type -Assembly System.IO.Compression.FileSystem
 }
 
+function Import-SubmoduleConfigurations
```
tylerl0706: Is this function needed?

I would think that in RestorePsesModules you could just do:

```powershell
    $modules = Get-Content -Raw "$PSScriptRoot/modules.json" | ConvertFrom-Json
    $modules.PSObject.Properties | ForEach-Object {
        $params = @{
            Name = $_.Name
            MinimumVersion = $_.Value.MinimumVersion
            MaximumVersion = $_.Value.MaximumVersion
            Path = "$PSScriptRoot/modules/"
        }
        if ($script:SaveModuleSupportsAllowPrerelease)        {
            $params += @{ AllowPrerelease = 
            $moduleInstallDetails.AllowPrerelease }
        }
        Save-Module @params
    }
```

tylerl0706:
AppVeyor so drunk

rjmholt:
No, those tests were failing on my machine too. I need to work out why that is.

rjmholt:
Ok, I've now run this against [vscode-powershell #1252](https://github.com/PowerShell/vscode-powershell/pull/1252) and verified that it works.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -115,8 +116,8 @@ public AnalysisService(string settingsPath, ILogger logger)
                 this.SettingsPath = settingsPath;
 
                 scriptAnalyzerModuleInfo = FindPSScriptAnalyzerModule(logger);
-                var sessionState = InitialSessionState.CreateDefault2();
-                sessionState.ImportPSModulesFromPath(scriptAnalyzerModuleInfo.ModuleBase);
+                InitialSessionState sessionState = InitialSessionState.CreateDefault2();
+                sessionState.ImportPSModulesFromPath(GetModulePathFromModuleInfo(scriptAnalyzerModuleInfo));
```
tylerl0706: Since we're just grabbing PSSA from the module path, I wonder if there's a helper variable somewhere that already has the Module path in it. Then we don't need  to do this fancy path logic with PSSA just to get the module path.

rjmholt: Yeah that's a good idea. It may even be a matter of querying the variable.

test/PowerShellEditorServices.Test.Host/ServerTestsBase.cs
```diff
@@ -111,12 +116,23 @@ public class ServerTestsBase
             // Start the process
             this.serviceProcess.Start();
 
+            /*
+            System.Console.WriteLine("Test Process PID: " + Process.GetCurrentProcess().Id);
+            System.Console.WriteLine("Service Process PID: " + this.serviceProcess.Id);
+            System.Console.ReadKey();
```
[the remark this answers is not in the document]
rjmholt: Ideally I'd like to wrap it in a conditional, unless you think that will make the code too ugly. I can just imagine wanting to use this again.

PowerShellEditorServices.build.ps1
```diff
@@ -179,7 +186,7 @@ task TestProtocol -If { !$script:IsUnix} {
 task TestHost -If { !$script:IsUnix} {
     Set-Location .\test\PowerShellEditorServices.Test.Host\
     exec { & $script:dotnetExe build -c $Configuration -f net452 }
-    exec { & $script:dotnetExe xunit -configuration $Configuration -framework net452 -verbose -nobuild -x86 }
```
rjmholt: Also, I have no idea what the consequences of this change could be. Because I'm not sure why that was there originally.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: I've been thinking about this. It looks like the way to access this directly is [kept internal](https://github.com/PowerShell/PowerShell/blob/2a45b48ae2f05c007636d716e41a9c60df6488ec/src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs#L1009). It's present as a property of `ExecutionContext`, but getting that out of the current runspace would be tricky and possibly error prone.

So another option is to get our powershell session to evaluate `$env:PSModulePath`, but I don't think that's very performant. My bigger concern is that, even if we don't consider the possibility that the module path could be changed in some way we don't expect, we still would be trying to import every module in the PSES module directory. Which we don't need in the analysis part and could potentially be quite slow.

But it's also true that I've overengineered the module path resolution bit already. Since we sort of know specifically whether PSSA will be using a versioned path or not. Except that this broke before. And ideally we could eventually move this method out and make it a static utility function.

PowerShellEditorServices.build.ps1
```diff
@@ -220,6 +227,62 @@ task LayoutModule -After Build {
...
+        if ($script:SaveModuleSupportsAllowPrerelease)
+        {
+            $body += @{ AllowPrerelease = $_.Value.AllowPrerelease }
+        }
+
+        $moduleInfos.Add($name, $body)
+    }
+
+    # Save each module in the modules.json file
+    foreach ($moduleName in $moduleInfos.Keys)
```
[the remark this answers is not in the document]
rjmholt: The only problem is that if there's a bad field in the json file, we'll install half the things and then abort.

Ideally, we could have an explicit failure line after the first foreach, but I'm not sure what the best way to detect a failure is or what the best way to fail in a `task` is.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
SeeminglyScience: Keep in mind PSSA can come from anywhere, we definitely don't want to lock it to the bundled version.

rjmholt: Ah, that's a very good point (which admittedly I had forgotten)!

But I'm pretty sure the new code still does that. It still retrieves the module given by `Get-Module PSScriptAnalyzer` and gets the path. The problem is that `PSModuleInfo` doesn't have a good way of representing its base path; `PSModuleInfo.ModuleBase` gives you what could be a versioned path that doesn't work with `Import-Module` (or the C# API equivalent used here).

So the only bit I added was to remove the versioned directory if it's the leaf directory.

SeeminglyScience: Oh for sure, I should have clarified that I was commenting on the thread and not the actual changes.  Side note, I usually just target the `psd1` in these situations.  Something like this

```csharp
Path.ChangeExtension(
    Path.Combine(
        moduleInfo.ModuleBase,
        moduleInfo.Name),
    ".psd1"))
```

But your approach accounts for modules without a manifest so it's more flexible in general.

rjmholt: Well actually I wanted to do that (or use whatever `PSModuleInfo.Path` gave me. Or there's also an overload of `InitialSessionState.ImportPSModulesFromPath` that takes an `IEnumerable<ModuleSpecification>`, and there's a constructor for `ModuleSpecification` that takes a `PSModuleInfo`.

**But**, it turns out that `ImportPSModulesFromPath` calls `ModuleUtils.GetDefaultAvailableModuleFiles` internally, [which only accepts paths to directories](https://github.com/PowerShell/PowerShell/blob/2a45b48ae2f05c007636d716e41a9c60df6488ec/src/System.Management.Automation/engine/Modules/ModuleUtils.cs#L200) as arguments.

(And the second method doesn't seem to populate the `ModuleSpecification` fields other than `Name`.)

I just wish PowerShell exposed a more convenient way to manipulate its modules and module loading.

SeeminglyScience: Ah right, I always forget the differences between those methods. I typically use `InitialSessionState.ImportPSModule`, which does work with `psd1` files.

> I just wish PowerShell exposed a more convenient way to manipulate its modules and module loading.

👍 👍 👍

rjmholt: In a quest to work out what might be best, I noticed that `InitialSessionState.ImportPSModulesFromPath()` in turn just converts the modules it finds into `ModuleSpecification`s and feeds them to `InitialSessionState.ImportPSModule`.

Anyway, I went to look at why this particular command can take a `ModuleSpecification` with everything but the `Name` field set to `null`, and have the `Name` be just a name or a path to a specification, or a whole heap of other things, and just work.

Well, turns out no matter what you do, PowerShell itself just calls `Import-Module` (see [here](https://github.com/PowerShell/PowerShell/blob/38c53c50a623f7c54707d348dd42d68a583b8932/src/System.Management.Automation/engine/InitialSessionState.cs#L3015)).

So I might actually just change our code to do that directly (even though I don't really like the idea of that)...

PowerShellEditorServices.build.ps1
tylerl0706: I suspect it's because VSCode use to only be 32bit but now it has a 64bit version so we can probably safely remove this.

PowerShellEditorServices.build.ps1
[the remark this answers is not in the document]
tylerl0706: eh. Not worth implementing. What you have is good :)

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -108,25 +109,25 @@ public string SettingsPath
...
             try
             {
                 this.SettingsPath = settingsPath;
 
-                scriptAnalyzerModuleInfo = FindPSScriptAnalyzerModule(logger);
-                var sessionState = InitialSessionState.CreateDefault2();
-                sessionState.ImportPSModulesFromPath(scriptAnalyzerModuleInfo.ModuleBase);
+                this._scriptAnalyzerModuleInfo = FindPSScriptAnalyzerModule(logger);
+                InitialSessionState sessionState = InitialSessionState.CreateDefault2();
+                sessionState.ImportPSModule(new [] { this._scriptAnalyzerModuleInfo.Path });
```
tylerl0706: We talked about this offline but I think it would be easier just to put `"PSScriptAnalyzer"` here instead of the path received from calling the `FindPSScriptAnalyzerModule(logger);` function. This also means that the `FindPSScriptAnalyzerModule` function can be totally removed from the code which is also a huge plus in my opinion.

I'd say give it a shot, if it doesn't work then this is fine.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -333,37 +333,6 @@ public void Dispose()
...
-                    return psModuleInfo;
-                }
-
-                logger.Write(
-                    LogLevel.Normal,
-                    "PSScriptAnalyzer module was not found.");
-                return null;
-            }
-        }
-
```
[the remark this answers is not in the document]
rkeithhill: Maybe I'm missing something but the deleted function seems to be replicating the Import-Module logic for loading a module.  That is, ipmo always loads the latest version in the PSModulePath anyway.  Unless the caller specifically needs the PSModuleInfo for some reason other than Import-Module.

rjmholt:
I've also noticed that there's a few lines in the `Start-EditorServices.ps1` script that want to install the right version of EditorServices. But I'm thinking we should remove those...

tylerl0706:
looks like a zip was added:

https://github.com/PowerShell/PowerShellEditorServices/pull/647/files#diff-837b6c39bb63ded0123f1da825124299

can you add something to the gitignore so that we don't have this anymore? I'm guessing it's the "artifact" from doing a full `Invoke-Build`

tylerl0706:
@rjmholt 

>I've also noticed that there's a few lines in the `Start-EditorServices.ps1` script that want to install the right version of EditorServices. But I'm thinking we should remove those...

are you talking about:
https://github.com/PowerShell/PowerShellEditorServices/blob/master/module/Start-EditorServices.ps1#L250-L261

We can delete that since that module doesn't actually exist. I think David wanted some way to bring in PowerShell Editor Services if it didn't exist but due to some folks being in environments without internet, bundling everything together was the only option.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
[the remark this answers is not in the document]
tylerl0706: @rkeithhill Actually, ipmo does not load the latest version, it loads the first version it finds. This could be a bug in PowerShell Core, but here's what I did:

```powershell
PS > $env:PSModulePath
/Users/tylerleonhardt/.local/share/powershell/Modules:/usr/local/share/powershell/Modules:/usr/local/microsoft/powershell/6.0.2/Modules

## ORDER:
# 1:    /Users/tylerleonhardt/.local/share/powershell/Modules
# 2:   /usr/local/share/powershell/Modules
# 3:   /usr/local/microsoft/powershell/6.0.2/Modules

# What's in #1's Pester
PS > gci /Users/tylerleonhardt/.local/share/powershell/Modules/Pester


    Directory: /Users/tylerleonhardt/.local/share/powershell/Modules/Pester


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----           2/9/18   4:14 PM                4.1.1

# What's in #2's Pester
PS > gci /usr/local/share/powershell/Modules/Pester/


    Directory: /usr/local/share/powershell/Modules/Pester


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----          4/13/18  12:20 PM                4.3.1


# Load Pester - see it's 4.1.1
PS > ipmo Pester
WARNING: OS Information retrieval is not possible, reports will contain only partial system data
/Users/tylerleonhardt/Desktop/Tech/PowerShell/vscode/PowerShellEditorServices [master ≡]> Get-Module Pester

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     4.1.1      Pester                              {Add-AssertionOperator, AfterAll, AfterEach, AfterEachFeature...}
```

rjmholt:
Ok, needed to put a PowerShell call back in to verify that we actually have PSScriptAnalyzer available and set all the `_hasScriptAnalyzer` checks to true (although I don't think they're necessary because the constructor fails if it isn't....).

Anyway, this class I'm touching is ripe for some refactoring (for starters because I don't think we should be calling PowerShell in the constructor at all, and should make a static factory method).

src/PowerShellEditorServices/Analysis/AnalysisService.cs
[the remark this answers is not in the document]
rkeithhill: Well that is somewhat disappointing behavior.  But it would explain why we had that method.  Another option would be to import with `-MinimumVersion` but that might be a pain to keep current.  Anytime we used a new PSSA feature we'd have to update the `MinimumVersion` parameter.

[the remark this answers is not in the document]
tylerl0706: Agreed - AFAIK that's how "PATH" works too.

> Another option would be to import with -MinimumVersion but that might be a pain to keep current

I like this idea. I don't see crazy PSSA changes just for PSES coming our way in the near future and I don't think they'll happen often. It's a small change to make us that much more resilient against someone who still has PSSA 1.0 or something that doesn't work with PSES.

@rjmholt sorry we keep giving you more work!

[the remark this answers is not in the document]
tylerl0706: I just remembered we're not actually doing `Import-Module` it's the `sessionState.ImportPSModule` and there doesn't really seem to be a good way to specify `MinimumVersion` in its overloads so I've fine with doing what we have in this PR.

rjmholt:
So my current question is: the intent of this PR is to make PSES a standalone, so it doesn't need to be build by the VSCode client, and as a first step to being used with other clients, and ideally being released onto the PowerShell Gallery.

In that case, in `Start-EditorServices.ps1`, we run `Import-Module PowerShellEditorServices` ([here](https://github.com/rjmholt/PowerShellEditorServices/blob/81fb727031f385e76c3f607dc3523d0464d9c8f2/scripts/Start-EditorServices.ps1#L247)), which might end up using a different EditorServices module if one exists further up the path.

Should we change it to ensure it only loads our particular EditorServices module?

tylerl0706:
So, the thing is, `Start-EditorServices` uses a path passed into it (called `BundledModulePath`) which gets tacked on to the PSModulePath for only that session.

So this means that, if someone was using PSES for another editor, say vim, they would still put PSES somewhere not on the PSModulePath, run `Start-EditorServices` and pass in that location. The location would be tacked on to the PSModulePath in that particular session and so `Import-Module PowerShellEditorServices` would work as intended because it'd be the only one in their PSModulePath.

All that said, I agree we eventually want to put PSES in the Gallery to be installable from `Install-Module` but to do that, I think we need to figure out what to do with `Start-EditorServices.ps1` - for example, do we want to add that to the PowerShellEditorServices module, or have it in its own module, or something like that.

tylerl0706:
that last paragraph could be a separate PR IMO.

rjmholt:
Agreed about the last paragraph. For now I just want to be able to build PSES by itself -- running it might be complicated, but at least we don't need the VSCode client codebase to do it.

tylerl0706:
This PR also allows the tests to be run locally which is great :)

tylerl0706:
Looks like this is good to go but I wanna merge in #629 first because I'd rather you have merge conflicts then them (not that I'm expecting any).

rjmholt:
Yes agreed

PR #649, opened 2018-04-06
Get rid of the unneeded exception variable causing a compile warning

rjmholt:
We've been sitting on a compile warning for a while, so I've removed the offending variable.


rjmholt:
Looks like one of those server tests is failing...

rkeithhill:
Random test failures are annoying.  :-(

PR #675, opened 2018-06-04
Add symbols to modules built in Debug configuration

rjmholt:
This copies over the .pdb files into the module so that it can be debugged alongside the/a client. Only copies the files in the Debug configuration.


rjmholt:
BTW, at some point I'd love to clean all this copy logic up into a big list of key/value pairs and just run a for loop over it.

Issue #657, opened 2018-04-23
Refactor Analysis code


rjmholt:
The PSES [analysis code](https://github.com/PowerShell/PowerShellEditorServices/tree/master/src/PowerShellEditorServices/Analysis) has grown a bit and recent forays into it indicate it might be due for a clean up. The script analysis part of PSES is one of its main jobs and where a lot of performance issues seem to lie. We should refactor the code here as much as possible to simplify how analysis works, do as little work as required, and ideally not call PowerShell unless we really need to.

rjmholt:
Looks like I should do this anyway since I need to fix https://github.com/PowerShell/vscode-powershell/issues/1298.

PR #677, opened 2018-06-05
Make AnalysisService use the latest version of PSScriptAnalyzer

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/1298 and https://github.com/PowerShell/PowerShellEditorServices/issues/657.

## Changes are:
* Load the latest version of PSScriptAnalyzer we can find on the path.
* Eliminate the `VerifyPSScriptAnalyzerAvailable`, since `ImportPSModule()` executes the same code internally.
* Do all the runspace manipulation in a factory method rather than in a constructor...
* Get rid of a bunch of needless private variables that just backed public ones.
* Get rid of `_hasScriptAnalyzerModule`, since it only got checked after the object was constructed (meaning it was always true...).
* Added some null checks on `AnalysisService` usage, since it can definitely be null.

### DIFF
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -37,12 +37,6 @@ namespace Microsoft.PowerShell.EditorServices
         private const string PSSA_MODULE_NAME = "PSScriptAnalyzer";
 
         /// <summary>
-        /// The minimum version of the PSScriptAnalyzer module with which we maintain
-        /// compatibility.
-        /// </summary>
-        private static readonly Version s_pssaMinimumVersion = new Version(1, 16);
-
-        /// <summary>
         /// Provides logging.
         /// </summary>
         private ILogger _logger;
@@ -512,21 +506,18 @@ namespace Microsoft.PowerShell.EditorServices
         {
             using (var ps = System.Management.Automation.PowerShell.Create())
             {
-                // Encode an equivalent of the PowerShell pipeline:
-                //   Get-Module -ListAvailable -Name "PSScriptAnalyzer" | Where-Object -Property "Version" -ge "1.16"
+                // Run `Get-Module -ListAvailable -Name "PSScriptAnalyzer"`
                 ps.AddCommand("Get-Module")
-                        .AddParameter("ListAvailable")
-                        .AddParameter("Name", PSSA_MODULE_NAME)
-                    .AddCommand("Where-Object")
-                        .AddParameter("Property", "Version")
-                        .AddParameter("ge")
-                        .AddParameter("Value", s_pssaMinimumVersion);
+                    .AddParameter("ListAvailable")
+                    .AddParameter("Name", PSSA_MODULE_NAME);
 
                 try
                 {
+                    // Get the latest version of PSScriptAnalyzer we can find
                     pssaModuleInfo = ps.Invoke()?
                         .Select(psObj => psObj.BaseObject)
                         .OfType<PSModuleInfo>()
+                        .OrderBy(moduleInfo => moduleInfo.Version)
                         .FirstOrDefault();
                 }
                 catch (Exception e)
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -26,14 +26,38 @@ namespace Microsoft.PowerShell.EditorServices
     {
         #region Private Fields
 
+        /// <summary>
+        /// Maximum number of runspaces we allow to be in use for script analysis.
+        /// </summary>
         private const int NumRunspaces = 1;
 
+        /// <summary>
+        /// Name of the PSScriptAnalyzer module, to be used for PowerShell module interactions.
+        /// </summary>
         private const string PSSA_MODULE_NAME = "PSScriptAnalyzer";
 
+        /// <summary>
+        /// The minimum version of the PSScriptAnalyzer module with which we maintain
+        /// compatibility.
+        /// </summary>
         private static readonly Version s_pssaMinimumVersion = new Version(1, 16);
 
+        /// <summary>
+        /// Provides logging.
+        /// </summary>
         private ILogger _logger;
+
+        /// <summary>
+        /// Runspace pool to generate runspaces for script analysis and handle
+        /// ansynchronous analysis requests.
+        /// </summary>
         private RunspacePool _analysisRunspacePool;
+
+        /// <summary>
+        /// Info object describing the PSScriptAnalyzer module that has been loaded in
+        /// to provide analysis services.
+        /// </summary>
+        private PSModuleInfo _pssaModuleInfo;
 
         /// <summary>
         /// Defines the list of Script Analyzer rules to include by default if
@@ -74,12 +98,18 @@ namespace Microsoft.PowerShell.EditorServices
 
         #region Constructors
 
-        private AnalysisService(RunspacePool analysisRunspacePool, string pssaSettingsPath, IEnumerable<string> activeRules, ILogger logger)
+        private AnalysisService(
+            RunspacePool analysisRunspacePool,
+            string pssaSettingsPath,
+            IEnumerable<string> activeRules,
+            ILogger logger,
+            PSModuleInfo pssaModuleInfo = null)
         {
             _analysisRunspacePool = analysisRunspacePool;
             SettingsPath = pssaSettingsPath;
             ActiveRules = activeRules.ToArray();
             _logger = logger;
+            _pssaModuleInfo = pssaModuleInfo;
         }
 
         #endregion // constructors
@@ -98,12 +128,13 @@ namespace Microsoft.PowerShell.EditorServices
             try
             {
                 RunspacePool analysisRunspacePool;
+                PSModuleInfo pssaModuleInfo;
                 try
                 {
                     // Try and load a PSScriptAnalyzer module with the required version
                     // by looking on the script path. Deep down, this internally runs Get-Module -ListAvailable,
                     // so we'll use this to check whether such a module exists
-                    analysisRunspacePool = CreatePssaRunspacePool();
+                    analysisRunspacePool = CreatePssaRunspacePool(out pssaModuleInfo);
 
                 }
                 catch (Exception e)
@@ -122,7 +153,12 @@ namespace Microsoft.PowerShell.EditorServices
                 analysisRunspacePool.ThreadOptions = PSThreadOptions.ReuseThread;
                 analysisRunspacePool.Open();
 
-                var analysisService = new AnalysisService(analysisRunspacePool, settingsPath, s_includedRules, logger);
+                var analysisService = new AnalysisService(
+                    analysisRunspacePool,
+                    settingsPath,
+                    s_includedRules,
+                    logger,
+                    pssaModuleInfo);
 
                 // Log what features are available in PSSA here
                 analysisService.LogAvailablePssaFeatures();
@@ -326,16 +362,20 @@ namespace Microsoft.PowerShell.EditorServices
                 return;
             }
 
-            PSObject[] modules = InvokePowerShell(
-                "Get-Module",
-                new Dictionary<string, object>{ {"Name", PSSA_MODULE_NAME} });
+            // If we already know the module that was imported, save some work
+            if (_pssaModuleInfo == null)
+            {
+                PSObject[] modules = InvokePowerShell(
+                    "Get-Module",
+                    new Dictionary<string, object>{ {"Name", PSSA_MODULE_NAME} });
 
-            PSModuleInfo pssaModuleInfo = modules
-                .Select(m => m.BaseObject)
-                .OfType<PSModuleInfo>()
-                .FirstOrDefault();
+                _pssaModuleInfo = modules
+                    .Select(m => m.BaseObject)
+                    .OfType<PSModuleInfo>()
+                    .FirstOrDefault();
+            }
 
-            if (pssaModuleInfo == null)
+            if (_pssaModuleInfo == null)
             {
                 throw new Exception("Unable to find loaded PSScriptAnalyzer module for logging");
             }
@@ -344,11 +384,11 @@ namespace Microsoft.PowerShell.EditorServices
             sb.AppendLine("PSScriptAnalyzer successfully imported:");
 
             // Log version
-            sb.AppendLine($"    Version: {pssaModuleInfo.Version}");
+            sb.AppendLine($"    Version: {_pssaModuleInfo.Version}");
 
             // Log exported cmdlets
             sb.AppendLine("    Exported Cmdlets:");
-            foreach (string cmdletName in pssaModuleInfo.ExportedCmdlets.Keys.OrderBy(name => name))
+            foreach (string cmdletName in _pssaModuleInfo.ExportedCmdlets.Keys.OrderBy(name => name))
             {
                 sb.AppendLine("        " + cmdletName);
             }
@@ -462,26 +502,7 @@ namespace Microsoft.PowerShell.EditorServices
         /// session state.
         /// </summary>
         /// <returns>A runspace pool with PSScriptAnalyzer loaded for running script analysis tasks.</returns>
-#if PowerShellv5
-        private static RunspacePool CreatePssaRunspacePool()
-        {
-            // Use public but unfriendly APIs to import a PSScriptAnalyzer module with the needed minimum version
-            var pssaModuleSpec = new ModuleSpecification(new Hashtable()
-            {
-                { "ModuleName", PSSA_MODULE_NAME },
-                { "ModuleVersion", s_pssaMinimumVersion }
-            });
-
-            // Create a base session state with PSScriptAnalyzer loaded
-            InitialSessionState sessionState = InitialSessionState.CreateDefault2();
-            sessionState.ImportPSModule(new [] { pssaModuleSpec });
-
-            // RunspacePool takes care of queuing commands for us so we do not
-            // need to worry about executing concurrent commands
-            return RunspaceFactory.CreateRunspacePool(sessionState);
-        }
-#else
-        private static RunspacePool CreatePssaRunspacePool()
+        private static RunspacePool CreatePssaRunspacePool(out PSModuleInfo pssaModuleInfo)
         {
             using (var ps = System.Management.Automation.PowerShell.Create())
             {
@@ -493,8 +514,6 @@ namespace Microsoft.PowerShell.EditorServices
                         .AddParameter("ge")
                         .AddParameter("Value", s_pssaMinimumVersion);
 
-
-                PSModuleInfo pssaModuleInfo;
                 try
                 {
                     pssaModuleInfo = ps.Invoke()?
@@ -521,7 +540,6 @@ namespace Microsoft.PowerShell.EditorServices
                 return RunspaceFactory.CreateRunspacePool(sessionState);
             }
         }
-#endif
 
         #endregion //private methods
     }
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -486,8 +486,13 @@ namespace Microsoft.PowerShell.EditorServices
             using (var ps = System.Management.Automation.PowerShell.Create())
             {
                 ps.AddCommand("Get-Module")
-                    .AddParameter("ListAvailable")
-                    .AddParameter("Name", PSSA_MODULE_NAME);
+                        .AddParameter("ListAvailable")
+                        .AddParameter("Name", PSSA_MODULE_NAME)
+                    .AddCommand("Where-Object")
+                        .AddParameter("Property", "Version")
+                        .AddParameter("ge")
+                        .AddParameter("Value", s_pssaMinimumVersion);
+
 
                 PSModuleInfo pssaModuleInfo;
                 try
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -555,6 +555,10 @@ namespace Microsoft.PowerShell.EditorServices
 
         private bool disposedValue = false; // To detect redundant calls
 
+        /// <summary>
+        /// Dispose of this object.
+        /// </summary>
+        /// <param name="disposing">True if the method is called by the Dispose method, false if called by the finalizer.</param>
         protected virtual void Dispose(bool disposing)
         {
             if (!disposedValue)
--- a/src/PowerShellEditorServices/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Analysis/AnalysisService.cs
@@ -125,8 +125,7 @@ namespace Microsoft.PowerShell.EditorServices
                 var analysisService = new AnalysisService(analysisRunspacePool, settingsPath, s_includedRules, logger);
 
                 // Log what features are available in PSSA here
-                analysisService.LogAvailablePssaCmdlets();
-                analysisService.LogAvailablePssaRules();
+                analysisService.LogAvailablePssaFeatures();
 
                 return analysisService;
             }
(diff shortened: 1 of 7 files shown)


src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -29,19 +30,16 @@ public class AnalysisService : IDisposable
 
         private const string PSSA_MODULE_NAME = "PSScriptAnalyzer";
 
+        private static readonly Version s_pssaMinimumVersion = new Version(1, 16);
```
rkeithhill: Considering 1.17 is shipping now, should we move to that version?

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -62,90 +60,83 @@ public class AnalysisService : IDisposable
...
-            {
-                return activeRules;
-            }
-
-            set
-            {
-                activeRules = value;
-            }
-        }
+        public string[] ActiveRules { get; set; }
```
rkeithhill: Ah, soo much nicer.  Thx.  :-)

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -333,43 +314,44 @@ public void Dispose()
...
+        /// <summary>
+        /// Write a list of all PSSA cmdlets that are available to the logs.
+        /// </summary>
+        private void LogAvailablePssaCmdlets()
         {
-            if (_hasScriptAnalyzerModule)
-            {
-                var sb = new StringBuilder();
-                var commands = InvokePowerShell(
+            var commands = InvokePowerShell(
```
rkeithhill: In a separate PR I'd like to change this from Get-Command to Get-Module -Name PSScriptAnalyzer.  From the resulting PSModuleInfo, we can get the Version of the module as well as the `ExportedCmdlets`.  Unless you're feeling spunky and want to make this change.  :-)

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -333,43 +314,44 @@ public void Dispose()
...
                 "Get-Command",
                 new Dictionary<string, object>
                 {
                     {"Module", "PSScriptAnalyzer"}
```
rkeithhill: I wonder if we should use `PSSA_MODULE_NAME` here?

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: Yes that crossed my mind. On one hand I feel like minimum version should encode what our lowest minimum supported version of PSSA is. On the other, we don't necessarily have a good answer for what that is, and I can't think of any downside to keeping it fresh...

rjmholt:
Uhoh, my last commit broke it! Will fix.

rjmholt:
Rats! The `ImportPSModule(IEnumerable<ModuleSpecification>)` API is only in PowerShell Core!!!

SeeminglyScience:
```
### Testing API usage for PowerShell v3...
Analysis\AnalysisService.cs(476,41): error CS1503: Argument 1: cannot convert from 'Microsoft.PowerShell.Commands.ModuleSpecification[]' to 'string[]' 
 [C:\projects\powershelleditorservices\src\PowerShellEditorServices\PowerShellEditorServices.csproj]
```

Might have to switch back to the method of loading a specific path until we can drop PSv3 :(

rjmholt:
@SeeminglyScience Yes I just saw that :face_with_head_bandage:. The annoying part is that there's no documentation about that fact :angry:

rkeithhill:
I feel your pain!

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,24 +455,73 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
 
-        private bool VerifyPSScriptAnalyzerAvailable()
+        /// <summary>
+        /// Create a new runspace pool around a PSScriptAnalyzer module for asynchronous script analysis tasks.
+        /// This uses the PowerShell module API to load the first PSScriptAnalyzer module on the module path with
+        /// with the required minimum version and then creates a runspace pool with that module loaded in the initial
+        /// session state.
+        /// </summary>
+        /// <returns>A runspace pool with PSScriptAnalyzer loaded for running script analysis tasks.</returns>
+#if PowerShellv5
```
SeeminglyScience: Not sure if that'll work like you expect unfortunately.  We don't actually build separate assemblies for each version (except as a test), the compiler directives just to allow it to compile.

The way we get this to work in the debugger is by having an interface (`IVersionSpecificOperations`) and several implementations for each PowerShell version required.  Then at runtime the implementation for that version is picked.   That way no missing member exceptions get thrown.

Here's the version specific interface stuff

[IVersionSpecificOperations](https://github.com/PowerShell/PowerShellEditorServices/blob/4bb71bd93320ba9b2105e3b5d2007f09f0f4a959/src/PowerShellEditorServices/Session/IVersionSpecificOperations.cs)
[PowerShell3Operations](https://github.com/PowerShell/PowerShellEditorServices/blob/4bb71bd93320ba9b2105e3b5d2007f09f0f4a959/src/PowerShellEditorServices/Session/PowerShell3Operations.cs)
[PowerShell4Operations](https://github.com/PowerShell/PowerShellEditorServices/blob/4bb71bd93320ba9b2105e3b5d2007f09f0f4a959/src/PowerShellEditorServices/Session/PowerShell4Operations.cs)
[PowerShell5Operations](https://github.com/PowerShell/PowerShellEditorServices/blob/4bb71bd93320ba9b2105e3b5d2007f09f0f4a959/src/PowerShellEditorServices/Session/PowerShell5Operations.cs)

And here's where that's used in [PowerShellContext](https://github.com/PowerShell/PowerShellEditorServices/blob/4bb71bd93320ba9b2105e3b5d2007f09f0f4a959/src/PowerShellEditorServices/Session/PowerShellContext.cs#L227-#L245).

At least, I'm pretty sure that's how it works.  It's possible that was done for readability though.  If y'all have a PSv3 machine there I'd give it a go real quick to confirm.  If it crashes immediately, that's probably why.

rjmholt: Bugger! I was just basing it on the defined constants in the csproj. I'll see if I can test it on a v3 machine. And then ideally we should save the better solution for later somewhere (not that I'm necessarily suggesting leaving it in as a comment)

rjmholt:
Ok, I've just reverted to the old way now, but include a cache for logging at least.

The better way for when we drop v3/4 support is here: https://gist.github.com/rjmholt/9820861c669610941a6d62b36d121fa7

rjmholt:
Just noticed this is also trying to log a `CmdletInfo.Version`

rjmholt:
This probably merits some interactive testing, but I'm happy to merge when others are. I tried to test it on a machine running PowerShell v4 but there was some problem with reference assembly linking...

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -283,11 +316,11 @@ public IEnumerable<string> GetPSScriptAnalyzerRules()
...
         {
-            if (this._analysisRunspacePool != null)
+            if (_analysisRunspacePool != null)
             {
-                this._analysisRunspacePool.Close();
-                this._analysisRunspacePool.Dispose();
-                this._analysisRunspacePool = null;
+                _analysisRunspacePool.Close();
+                _analysisRunspacePool.Dispose();
+                _analysisRunspacePool = null;
```
tylerl0706: is the nulling needed here? I'm not sure.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -333,43 +365,58 @@ public void Dispose()
...
+            foreach (string cmdletName in _pssaModuleInfo.ExportedCmdlets.Keys.OrderBy(name => name))
             {
-                var rules = GetPSScriptAnalyzerRules();
-                var sb = new StringBuilder();
-                sb.AppendLine("Available PSScriptAnalyzer Rules:");
-                foreach (var rule in rules)
-                {
-                    sb.AppendLine(rule);
-                }
+                sb.AppendLine("        " + cmdletName);
```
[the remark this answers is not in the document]
rjmholt: I deliberately avoided it here on the basis that we're using string builders for performance. In fact, [it looks like we should split this out into new string builder calls](https://koukia.ca/c-string-interpolation-performance-test-73ae6ab27276). I'll make the change to that.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: Well really we should implement the [Dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern). I'm guessing the nulling is to allow GC to collect, but that's protecting against the scenario where the parent is disposed and the reference still carried around for a while. It doesn't hurt I guess.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,22 +511,51 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
+        /// with the required minimum version and then creates a runspace pool with that module loaded in the initial
+        /// session state.
+        /// </summary>
+        /// <returns>A runspace pool with PSScriptAnalyzer loaded for running script analysis tasks.</returns>
+        private static RunspacePool CreatePssaRunspacePool(out PSModuleInfo pssaModuleInfo)
         {
             using (var ps = System.Management.Automation.PowerShell.Create())
             {
+                // Encode an equivalent of the PowerShell pipeline:
+                //   Get-Module -ListAvailable -Name "PSScriptAnalyzer" | Where-Object -Property "Version" -ge "1.16"
```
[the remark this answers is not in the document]
rjmholt: I feel like we should address that in an issue and have a proper discussion - just so we know what scenarios there are for version-locking vs flexible drop-in.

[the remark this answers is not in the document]
bergmeister: If I remember correctly, the extension used to install PSSA in its AppData folder but does not do that any more. Is that correct?
I don't know much about the compatibility of PSSA before 1.16.1 but the version of it should not matter too much. For example due to one of the recent PRs, PSES now has a new rule from 1.17.0 but it is no problem if one has only 1.16.1 because PSSA just ignores rules in the pssa settings that it  does not recognize.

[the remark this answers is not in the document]
rjmholt: Currently we install PSSA to a directory internal to the extension itself, and when the extension is loaded it adds that directory to the end of the module path in the internally hosted PowerShell session.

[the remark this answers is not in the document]
rjmholt: The question is really whether we always take the latest PSSA (i.e. PSSA 17.1) or we use a minimum version.

Given the recent re-release of PSSA to 17.1, I would imagine it benefits us to only use a minimum version, so that we don't block an update of PSSA unnecessarily (by block I mean, users would only get that new release when we do our release).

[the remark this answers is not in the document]
rjmholt: So by that logic, I suppose I'm suggesting we should:
* Use a minimum version
* Set the minimum version to whatever the currently released version of PSSA is at our release time

Ideally we could tool the second point so that we don't have to update the version in source here, but I could see that taking some doing.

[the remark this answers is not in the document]
tylerl0706: When I meant latest - I meant like... sort the results we get from `Get-Module` and then grab the newest on the machine like it use to be.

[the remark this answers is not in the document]
tylerl0706: Because as it is now, we don't really know what `.FirstOrDefault` a few lines down is going to give us. Could be the latest on the machine, could be the oldest on the machine, or a random one.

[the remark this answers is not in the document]
SeeminglyScience: Does `Get-Module -ListAvailable` output in the order that `Import-Module` would import from?

If it does, then I think there's value in it.  That conforms well with the original intention of the change, to give the user more control over what version is used. 

If it doesn't, then we should probably stick to the highest version installed until we can drop v3/4 and use the better `initialsessionstate.ImportModule` API.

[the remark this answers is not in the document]
rjmholt: `Get-Module -ListAvailable` looks like it goes over the module path in order (i.e. respects the module path).

Actual method that gets all the modules:
https://github.com/PowerShell/PowerShell/blob/4192463ce6119ec298459ea60cd12f5c2fa1f07c/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L1001-L1045

ModuleIntrinsics method it uses to enumerate the module path:
https://github.com/PowerShell/PowerShell/blob/4192463ce6119ec298459ea60cd12f5c2fa1f07c/src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs#L1009-L1031

Also worth noting that the newer module import API I wanted to move to before we noticed the incompatibility [also uses `Get-Module -ListAvailable` internally](https://github.com/PowerShell/PowerShell/blob/ebc7cdf847ef39c8d71fd0f69849bf59f498111b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L3969-L3984).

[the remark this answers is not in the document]
rjmholt: I'm assuming of course that older PowerShell versions also have this behaviour...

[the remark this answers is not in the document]
rjmholt: I've looked at Windows PowerShell 5.1 and I don't think it guarantees a module path ordering for `Get-Module -ListAvailable`. In fact, neither does `InitialSessionState.ImportPSModule(ModuleSpecification[])` or `InitialSessionState.ImportPSModule(string[])`. I don't there is any API for module imports that respects the module path. We would need to search the path ourselves...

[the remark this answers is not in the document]
SeeminglyScience: Welp that complicates things.

The benefit of obeying `Import-Module` ordering logic is predictability.  If we lose that it feels more like a gamble which module you're going to import.  Do we know of anyone who was specifically trying to downgrade PSSA?  It isn't super common for folks to build against PSSA as far as I know, so I don't think issues with it will be as common as say Newtonsoft.  If we haven't had any specific complaints we may want to switch back to importing the latest only.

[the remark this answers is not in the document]
bergmeister: Why not just select the version number and pipe to sort to pick the latest one?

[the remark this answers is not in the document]
rkeithhill: The original idea was to use at least the version we shipped with the extension but allow users to use a "newer" version if installed on their machine.  So, using latest sounds good to me.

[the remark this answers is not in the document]
rjmholt: Yeah I think the use-latest approach is the best way -- at least `InitialSessionState.ImportPSModulesFromPath` seems to have a sane behaviour.

rjmholt:
Ok, we're now using the latest version of PSScriptAnalyzer again 😄

src/PowerShellEditorServices/Analysis/AnalysisService.cs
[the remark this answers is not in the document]
rjmholt: I've opened an issue for our above discussion BTW:
https://github.com/PowerShell/PowerShell/issues/7020

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
+                // need to worry about executing concurrent commands
+                return RunspaceFactory.CreateRunspacePool(sessionState);
             }
         }
 
         #endregion //private methods
+
+        #region IDisposable Support
+
+        private bool disposedValue = false; // To detect redundant calls
```
SeeminglyScience: I wonder if there's a way to set this refactor to prefix this field with `_`.  I always forget to change this

rjmholt: Me too!

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -25,23 +26,38 @@ public class AnalysisService : IDisposable
...
+        /// to provide analysis services.
+        /// </summary>
+        private PSModuleInfo _pssaModuleInfo;
 
         /// <summary>
         /// Defines the list of Script Analyzer rules to include by default if
         /// no settings file is specified.
         /// </summary>
-        private static readonly string[] IncludedRules = new string[]
+        private static readonly string[] s_includedRules = new string[]
```
bergmeister: What is the `s` in there (I guess a typo, not Hungarian notation).
Just a though in general as I see that you change the style in general to be more consistent. You should make sure that those changes make it into version 1.x of the extension so that when you branch off to version 2.0, it is still easily possible to merge back commits to version 1.x

rjmholt: The `s_` is to indicate a private static variable - so I suppose it is a sort of Hungarian-notation-lite - in the same vein as the `_` before private variables. It's the convention used in all the other Microsoft code I've seen. There are places within this codebase where it's not done as much, but I think we should change it over as we touch the code for both clarity and consistency with related projects.

I don't quite follow the latter part of your comment, but if you're saying we should try and make as much as possible work with 1.x versions, I am strongly in agreement. I don't think the v2 branch will be taking in commits from master for a little while yet. I'm not sure it's worth one big commit to change the style over - I imagine opinions might differ on that - although I'd be happy to do it personally.

bergmeister: As, I see, I have never seen it this way (or probably I've just never recognized it). To re-word my other comment, which was more about the workflow in the (distant) future: Style changes should not be made any more once v2 has actually branched off (i.e. has become the mainstream branch) to allow for merge-ability of commits in v2 back to v1 (or style changes should be applied to both branches otherwise).

rjmholt: Yes agreed. My opinion there then is that we should invest in a sweeping style change PR. The style in PSES is already inconsistent but is moving over to be much closer to the conventions recommended by Microsoft and used in the other PowerShell projects (PascalCase properties, `_`-prefixed private variables, etc.). Otherwise we'll be stuck with code that is less familiar and (in my opinion) not as clear.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -333,43 +346,61 @@ public void Dispose()
...
+                    .OfType<PSModuleInfo>()
+                    .FirstOrDefault();
+            }
 
-                sb.AppendLine("The following cmdlets are available in the imported PSScriptAnalyzer module:");
-                sb.AppendLine(String.Join(Environment.NewLine, commandNames.Select(s => "    " + s)));
-                this._logger.Write(LogLevel.Verbose, sb.ToString());
+            if (_pssaModuleInfo == null)
+            {
+                throw new Exception("Unable to find loaded PSScriptAnalyzer module for logging");
```
tylerl0706: Will this fail gracefully? We don't want all of PSES to crash if it can't find PSSA.

tylerl0706: This could be a Error log instead of throwing.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
             return await task;
         }
 
-        private bool VerifyPSScriptAnalyzerAvailable()
+        /// <summary>
+        /// Create a new runspace pool around a PSScriptAnalyzer module for asynchronous script analysis tasks.
+        /// This uses the PowerShell module API to load the first PSScriptAnalyzer module on the module path with
+        /// with the required minimum version and then creates a runspace pool with that module loaded in the initial
```
tylerl0706: nit: update the comment to say we load the latest on the machine

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
+                        .Select(psObj => psObj.BaseObject)
+                        .OfType<PSModuleInfo>()
+                        .OrderBy(moduleInfo => moduleInfo.Version)
+                        .FirstOrDefault();
                 }
-                catch (Exception)
+                catch (Exception e)
                 {
-                    return false;
+                    throw new Exception("Unable to find PSScriptAnalyzer module", e);
```
tylerl0706: Same question from above about failing gracefully

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
                 }
+
+                if (pssaModuleInfo == null)
+                {
+                    throw new Exception("Unable to find PSScriptAnalyzer module");
```
tylerl0706: same

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
+        /// <param name="disposing">True if the method is called by the Dispose method, false if called by the finalizer.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    _analysisRunspacePool.Close();
+                    _analysisRunspacePool.Dispose();
+                    _analysisRunspacePool = null;
```
tylerl0706: I'm pretty sure you only need to do the Close, here. I think the Close does a Dispose. I'm also curious if the null line is needed.

src/PowerShellEditorServices/Analysis/AnalysisService.cs
```diff
@@ -467,25 +495,85 @@ private PSObject[] InvokePowerShell(string command, IDictionary<string, object>
...
+            }
+        }
+
+        /// <summary>
+        /// Clean up all internal resources and dispose of the analysis service.
+        /// </summary>
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(true);
```
tylerl0706: Love this pattern :)

src/PowerShellEditorServices/Session/EditorSession.cs
```diff
@@ -164,7 +164,7 @@ internal void InstantiateAnalysisService(string settingsPath = null)
             // Script Analyzer binaries are not included.
             try
             {
-                this.AnalysisService = new AnalysisService(settingsPath, this.logger);
+                this.AnalysisService = AnalysisService.Create(settingsPath, this.logger);
```
tylerl0706: If only .NET Core used the composer pattern with NamedPipes 🙃

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: Good catch. The invocation of `AnalysisService.Create()` is already wrapped in a `try/catch`, and I would prefer not to do both. Do you have a preference on catching all exceptions outside vs inside?

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: Had a quick look at the code and it looks like `Dispose()` calls `Close()`:
https://github.com/PowerShell/PowerShell/blob/c1c5344a8897262433141ecbc13bb06ac2c4bbef/src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs#L823-L836

I'm not sure that `Close()` calls `Dispose()`.

Dunno how many resources or how much memory a runspace pool allocates, but the comments generated with the Dispose pattern snippet indicate large resources [should be set to `null`](https://gist.github.com/traktraktrugui/7330187#file-bdisp-xml-L46-L47).

src/PowerShellEditorServices/Analysis/AnalysisService.cs
rjmholt: Ah turns out this is already handled -- `Create` catches all exceptions, logs them as exceptions and returns `null`.

rjmholt:
I've converted all the `Exception`s I threw to be `AnalysisServiceLoadException`s, so we know when we have an unexpected failure mode. `AnalysisService.Create()` catches all exceptions so I've removed the `try/catch` in the caller.

I've also updated the comment to reflect the truth of what we do now (with the latest version) and improved our use of the dispose pattern.

rjmholt:
@tylerl0706 can you have a quick look at this and if you approve we can merge it

PR #697, opened 2018-07-08
Add functionality to allow a Show-Command like panel in VS Code

corbob:
This is changes required to accommodate the functionality of [VS Code Feature Request 1405](https://github.com/PowerShell/vscode-powershell/issues/1405)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandRequest.cs
@@ -9,7 +9,6 @@ using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
-
     /// <summary>
     /// Describes the request to get the details for PowerShell Commands from the current session.
     /// </summary>
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandRequest.cs
@@ -11,8 +11,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
 
     /// <summary>
-    /// Defines a class that describes the request to get the details for a single PowerShell Command
-    /// from the current session.
+    /// Describes the request to get the details for PowerShell Commands from the current session.
     /// </summary>
         public class GetCommandRequest
     {
@@ -22,7 +21,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
     }
 
     /// <summary>
-    /// Defines a class that describes the message to get the details for a single PowerShell Command
+    /// Describes the message to get the details for a single PowerShell Command
     /// from the current session
     /// </summary>
     public class PSCommandMessage
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -1,0 +1,16 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
+
+namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
+{
+    public class GetCommandsRequest
+    {
+        public static readonly
+            RequestType<string, string, object, object> Type =
+            RequestType<string, string, object, object>.Create("powerShell/getCommands");
+    }
+}
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -9,11 +9,18 @@ using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
-    public class GetCommandsRequest
+    public class GetAllCommandsRequest
     {
         public static readonly
-            RequestType<List<PSCommandMessage>, object, object, object> Type =
-            RequestType<List<PSCommandMessage>, object, object, object>.Create("powerShell/getCommands");
+            RequestType<List<PSAllCommandsMessage>, object, object, object> Type =
+            RequestType<List<PSAllCommandsMessage>, object, object, object>.Create("powerShell/getAllCommands");
+    }
+
+        public class GetCommandRequest
+    {
+        public static readonly
+            RequestType<string, object, object, object> Type =
+            RequestType<string, object, object, object>.Create("powerShell/getCommand");
     }
 
     public class PSCommandMessage
@@ -21,8 +28,13 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         public string Name { get; set; }
         public string ModuleName { get; set; }
         public string DefaultParameterSet { get; set; }
-        public CommandTypes CommandType { get; set; }
         public Dictionary<string, ParameterMetadata> Parameters { get; set; }
         public System.Collections.ObjectModel.ReadOnlyCollection<CommandParameterSetInfo> ParameterSets { get; set; }
+    }
+
+    public class PSAllCommandsMessage
+    {
+        public string Name { get; set; }
+        public string ModuleName { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -5,6 +5,7 @@
 
 using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
 using System.Collections.Generic;
+using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
@@ -20,8 +21,8 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         public string Name { get; set; }
         public string ModuleName { get; set; }
         public string DefaultParameterSet { get; set; }
-        public System.Management.Automation.CommandTypes CommandType { get; set; }
-        public Dictionary<string, System.Management.Automation.ParameterMetadata> Parameters { get; set; }
-        public System.Collections.ObjectModel.ReadOnlyCollection<System.Management.Automation.CommandParameterSetInfo> ParameterSets { get; set; }
+        public CommandTypes CommandType { get; set; }
+        public Dictionary<string, ParameterMetadata> Parameters { get; set; }
+        public System.Collections.ObjectModel.ReadOnlyCollection<CommandParameterSetInfo> ParameterSets { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -9,6 +9,9 @@ using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
+    /// <summary>
+    /// Defines a class that describes the request to get all PowerShell Commands from the current session
+    /// </summary>
     public class GetAllCommandsRequest
     {
         public static readonly
@@ -16,6 +19,10 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
             RequestType<List<PSAllCommandsMessage>, object, object, object>.Create("powerShell/getAllCommands");
     }
 
+    /// <summary>
+    /// Defines a class that describes the request to get the details for a single PowerShell Command
+    /// from the current session.
+    /// </summary>
         public class GetCommandRequest
     {
         public static readonly
@@ -23,6 +30,10 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
             RequestType<string, object, object, object>.Create("powerShell/getCommand");
     }
 
+    /// <summary>
+    /// Defines a class that describes the message to get the details for a single PowerShell Command
+    /// from the current session
+    /// </summary>
     public class PSCommandMessage
     {
         public string Name { get; set; }
@@ -32,6 +43,9 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         public System.Collections.ObjectModel.ReadOnlyCollection<CommandParameterSetInfo> ParameterSets { get; set; }
     }
 
+    /// <summary>
+    /// Defines a class that describes the message to get all PowerShell Commands from the current session
+    /// </summary>
     public class PSAllCommandsMessage
     {
         public string Name { get; set; }
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -9,15 +9,6 @@ using System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
-    /// <summary>
-    /// Defines a class that describes the request to get all PowerShell Commands from the current session
-    /// </summary>
-    public class GetAllCommandsRequest
-    {
-        public static readonly
-            RequestType<List<PSAllCommandsMessage>, object, object, object> Type =
-            RequestType<List<PSAllCommandsMessage>, object, object, object>.Create("powerShell/getAllCommands");
-    }
 
     /// <summary>
     /// Defines a class that describes the request to get the details for a single PowerShell Command
@@ -41,14 +32,5 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         public string DefaultParameterSet { get; set; }
         public Dictionary<string, ParameterMetadata> Parameters { get; set; }
         public System.Collections.ObjectModel.ReadOnlyCollection<CommandParameterSetInfo> ParameterSets { get; set; }
-    }
-
-    /// <summary>
-    /// Defines a class that describes the message to get all PowerShell Commands from the current session
-    /// </summary>
-    public class PSAllCommandsMessage
-    {
-        public string Name { get; set; }
-        public string ModuleName { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
@@ -4,13 +4,24 @@
 //
 
 using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
+using System.Collections.Generic;
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
     public class GetCommandsRequest
     {
         public static readonly
-            RequestType<string, string, object, object> Type =
-            RequestType<string, string, object, object>.Create("powerShell/getCommands");
+            RequestType<List<PSCommandMessage>, object, object, object> Type =
+            RequestType<List<PSCommandMessage>, object, object, object>.Create("powerShell/getCommands");
+    }
+
+    public class PSCommandMessage
+    {
+        public string Name { get; set; }
+        public string ModuleName { get; set; }
+        public string DefaultParameterSet { get; set; }
+        public System.Management.Automation.CommandTypes CommandType { get; set; }
+        public Dictionary<string, System.Management.Automation.ParameterMetadata> Parameters { get; set; }
+        public System.Collections.ObjectModel.ReadOnlyCollection<System.Management.Automation.CommandParameterSetInfo> ParameterSets { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -530,15 +530,21 @@ function __Expand-Alias {
             RequestContext<object> requestContext)
         {
             PSCommand psCommand = new PSCommand();
-            if (param != "") {
+            if (!string.IsNullOrEmpty(param)) {
                 psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command").AddArgument(param);
             }
             else
             {
-                psCommand.AddScript("Microsoft.PowerShell.Core\\Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command")
+                        .AddParameter("CommandType", new[]{"Function", "Cmdlet", "ExternalScript"})
+                    .AddCommand("Select-Object")
+                        .AddParameter("Property", new[]{"Name", "ModuleName"})
+                    .AddCommand("Sort-Object")
+                        .AddParameter("Property", "Name");
+                // psCommand.AddScript("Microsoft.PowerShell.Core\\Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
             }
             IEnumerable<PSObject> result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
-            List<PSCommandMessage> commandList = new List<PSCommandMessage>();
+            var commandList = new List<PSCommandMessage>();
 
             if (result != null)
             {
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -525,17 +525,30 @@ function __Expand-Alias {
         }
 
         private async Task HandleGetCommandsRequest(
-            string content,
-            RequestContext<string> requestContext)
+            object param,
+            RequestContext<object> requestContext)
         {
-            var script = @"
-            Get-Command Get-Command | ConvertTo-Json
-";
             var psCommand = new PSCommand();
-            psCommand.AddScript("Get-Command Get-Command | Select-Object Name,Parameters,ParameterSets | ConvertTo-Json");
-            var result = await this.editorSession.PowerShellContext.ExecuteCommand<string>(psCommand);
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript");
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
 
-            await requestContext.SendResult(result.First().ToString());
+            var commandList = new List<PSCommandMessage>();
+
+            if (result != null)
+            {
+                foreach (dynamic c in result)
+                {
+                    commandList.Add(new PSCommandMessage {
+                        Name = c.Name,
+                        ModuleName = c.ModuleName,
+                        Parameters = c.Parameters,
+                        ParameterSets = c.ParameterSets,
+                        DefaultParameterSet = c.DefaultParameterSet,
+                        CommandType = c.CommandType });
+                }
+            }
+
+            await requestContext.SendResult(commandList);
         }
 
         private async Task HandleFindModuleRequest(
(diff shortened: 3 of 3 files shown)


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +467,33 @@ RequestContext<object> requestContext
             await requestContext.SendResult(result.First().ToString());
         }
 
+        private async Task HandleGetCommandsRequest(
+            object param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript");
```
tylerl0706: Did you remove the sorting here?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +467,33 @@ RequestContext<object> requestContext
...
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
+
+            var commandList = new List<PSCommandMessage>();
+
+            if (result != null)
+            {
+                foreach (dynamic c in result)
```
tylerl0706: I think these are just PSObjects. No need for the dynamic keyword, I think

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
```diff
@@ -0,0 +1,27 @@
...
+            RequestType<List<PSCommandMessage>, object, object, object> Type =
+            RequestType<List<PSCommandMessage>, object, object, object>.Create("powerShell/getCommands");
+    }
+
+    public class PSCommandMessage
+    {
+        public string Name { get; set; }
+        public string ModuleName { get; set; }
+        public string DefaultParameterSet { get; set; }
+        public System.Management.Automation.CommandTypes CommandType { get; set; }
```
tylerl0706: Can you add a `using` for SMA? That way you don't have to repeat yourself.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: It looks like I just hadn't committed the sorting. I've pushed the sorting.

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
corbob: Fixed 😃

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rjmholt: Looking through [the docs](https://docs.microsoft.com/en-au/dotnet/api/system.management.automation.psobject?view=powershellsdk-1.1.0), `PSObject` doesn't have a simple way to access its members, but it does implement [`IDynamicMetaObjectProvider`](https://docs.microsoft.com/en-au/dotnet/api/system.dynamic.idynamicmetaobjectprovider?view=netframework-4.5.1), which looks like its specifically intended for `dynamic` usage. So it looks like `dynamic` is the best (and intended) way to go here.

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
+{
+    public class GetCommandsRequest
+    {
+        public static readonly
+            RequestType<List<PSCommandMessage>, object, object, object> Type =
+            RequestType<List<PSCommandMessage>, object, object, object>.Create("powerShell/getCommands");
+    }
+
+    public class PSCommandMessage
```
rjmholt: All the public bits should ideally have comments here. The important part is that either `GetCommandsRequest` or `PSCommandMessage` documents what feature this message services and why it exists. But this is a public API, so we should aim to document it well.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +467,33 @@ RequestContext<object> requestContext
             await requestContext.SendResult(result.First().ToString());
         }
 
+        private async Task HandleGetCommandsRequest(
+            object param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Sort-Object Name");
```
rjmholt: If we think performance is a concern, we are better off sorting the result in C#. That should be possible with something like:

```csharp
var psCommand = new PSCommand()
    .AddScript("Get-Command -CommandType Function,Cmdlet,ExternalScript");

// Dunno if the <dynamic> bit works, but saves a coercion
var allCommands = await this.editorSession.PowerShellContext.ExecuteCommand<dynamic>(psCommand);

if (allCommands == null)
{
    await requestContext.SendResult(List<PSCommandMessage>.Empty);
    return;
}

var commandList = new List<PSCommandMessage>();
foreach (dynamic cmd in allCommands.OrderBy(c => c.Name, StringComparer.OrdinalIgnoreCase))
{
    commandList.Add(new PSCommandMessage()
        {
            Name = cmd.Name,
            ModuleName = cmd.ModuleName,
            Parameters = cmd.Parameters,
            ParameterSets = cmd.ParameterSets,
            DefaultParameterSet = cmd.DefaultParameterSet,
            CommandType = cmd.CommandType,
        });
}

await requestContext.SendResult(commandList);
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
SeeminglyScience: I don't know which way is best, but this is the way I typically use/see most often

```csharp
pso.Properties["PropertyName"].Value;
pso.Methods["MethodName"].Invoke(arg1, arg2);
```

rjmholt: Ah, I wondered if that was possible. I imagine using `dynamic` is a syntactic sugar for that. It's possible that there's some overhead on using the `dynamic` way, hooking up the interface, but it's probably minor enough that it comes down to preference

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+//
+
+using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
+using System.Collections.Generic;
+using System.Management.Automation;
+
+namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
+{
+    /// <summary>
+    /// Defines a class that describes the request to get all PowerShell Commands from the current session
```
rjmholt: Style nit: "Defines a class that describes" is implicit; the comment should say what the code doesn't. Ditto for other comments below

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
         }
 
+        private async Task HandleGetAllCommandsRequest(
+            object param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+            Logger.Write(LogLevel.Verbose, $"Calling {psCommand.Commands}");
+            Stopwatch stopwatch = new Stopwatch();
```
rjmholt: Note for later:

I've seen this pattern a few times in the code base. Maybe we could abuse an `IDisposable` to sugar the syntax to:
```csharp
using (logger.TimeExecution())
{
    _editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
}
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
+            stopwatch.Start();
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
+            stopwatch.Stop();
+            Logger.Write(LogLevel.Verbose, $"Returned from Get-Command with {result.Count()} results in {stopwatch.ElapsedMilliseconds}ms");
```
rjmholt: If `result` is null, we'll get an NRE here

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
+        private async Task HandleGetAllCommandsRequest(
+            object param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+            Logger.Write(LogLevel.Verbose, $"Calling {psCommand.Commands}");
+            Stopwatch stopwatch = new Stopwatch();
+            stopwatch.Start();
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
```
rjmholt: Given that we expect a type that isn't `PSObject` here, it might be clearer to put `IEnumerable<PSObject>` instead of `var`

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public class GetAllCommandsRequest
+    {
+        public static readonly
+            RequestType<List<PSAllCommandsMessage>, object, object, object> Type =
```
rjmholt: Is there a way to avoid using `object` so much? Ideally we would be able to express the API endpoint with types

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
             await requestContext.SendResult(result.First().ToString());
         }
 
+        private async Task HandleGetAllCommandsRequest(
+            object param,
+            RequestContext<object> requestContext)
```
rjmholt: If we're sending back a `List<PSCommandMessage>`, that should be the generic type of the `RequestContext` I think. Ideally we could return something where we can safely cache an empty result, like `PSCommandMessage[]`, so when `result` is `null` or empty, we can send back `Array.Empty<PSCommandResult>()` and save an allocation

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+            Logger.Write(LogLevel.Verbose, $"Calling {psCommand.Commands}");
+            Stopwatch stopwatch = new Stopwatch();
+            stopwatch.Start();
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
+            stopwatch.Stop();
+            Logger.Write(LogLevel.Verbose, $"Returned from Get-Command with {result.Count()} results in {stopwatch.ElapsedMilliseconds}ms");
+            var commandList = new List<PSAllCommandsMessage>();
```
rjmholt: If we already call count (and the result is secretly a `Collection<T>` anyway), this list can be initialised with the number of objects expected. Or it could be an array

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
+            object param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
+            psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+            Logger.Write(LogLevel.Verbose, $"Calling {psCommand.Commands}");
+            Stopwatch stopwatch = new Stopwatch();
+            stopwatch.Start();
+            var result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
+            stopwatch.Stop();
```
rjmholt: After this line, I would maybe handle the `null` case:

```csharp
if (result == null)
{
    Logger.Write(LogLevel.Verbose, /* Message about null result */);
    await requestContext.SendResult(/* Empty thing */);
    return;
}

Logger.Write(LogLevel.Verbose, /* Normal message */);

if (result.Count() == 0)
{
    await requestContext.SendResult(/* Empty thing */);
    return;
}

// Handle golden path
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -466,6 +469,71 @@ RequestContext<object> requestContext
...
+            }
+
+            await requestContext.SendResult(commandList);
+        }
+
+        private async Task HandleGetCommandRequest(
+            string param,
+            RequestContext<object> requestContext)
+        {
+            var psCommand = new PSCommand();
```
rjmholt: For both of these methods we're using a similar pattern twice:
1. Compose PowerShell command
2. Log call
3. Start stopwatch
4. Run command
5. Stop stopwatch
6. Log call and performance

I feel like steps 2-6 should be done for every PowerShell command (i.e. should be implicit in the `ExecuteCommand` call), but certainly here should be turned into their own function

rjmholt:
Sorry it took me so long to review @corbob. Everything looks really good, I just left a couple of comments

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: This was mostly put in while I was troubleshooting a performance issue where it was taking 4 - 5 seconds to come back where the command in normal powershell was under a second. Turns out the time was being consumed by sending 800,000 lines of json (hence we now get all commands only and have a separate message to get all of the command information).

To avoid refactoring it out later, I've removed this pattern in favor of adding it as a separate PR to ExecuteCommand.

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
corbob: I would love this, but I'm not sure what else I could use. It doesn't like the use of void here, and when I attach the debugger and step into the constructor this is what I see:
image

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: If I change this to use RequestContext<List<PSGetAllCommandsRequest>> it errors on line 127 where we specify the handler:

`Argument 2: cannot convert from 'method group' to 'Func<List<PSAllCommandsMessage>, RequestContext<object>, Task>' [PowerShellEditorServices.Protocol]`

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
rjmholt: I might be wrong about this, but my understanding is that .NET will instantiate the generic with the type you ask it for; `TResult` is object because that's what it's told to instantiate as (i.e. .NET doesn't have the capacity to work out a better type).

What we should do is work out what the data being sent across is and if it's already described by some type, use that type, or otherwise create a new type to describe that data. That should also solve the issue below as well. I'll give the code a proper read/compile tomorrow and hopefully we'll see if there's a nice solution 😄

corbob: Works for me. I'm just going through the vscode-powershell PR comments now and will push them up.

corbob:
I believe I've addressed most of the comments with the latest commit, but there is still the question of a few that I replied to.

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <summary>
+    /// Defines a class that describes the request to get the details for a single PowerShell Command
+    /// from the current session.
+    /// </summary>
+        public class GetCommandRequest
+    {
+        public static readonly
+            RequestType<string, object, object, object> Type =
+            RequestType<string, object, object, object>.Create("powerShell/getCommand");
```
tylerl0706: So I just thought of something and I REALLY HOPE @corbob doesn't hate me after this one...

if the difference between PSCommandMessage and PSAllCommandsMessage is just a `DefaultParameterSet` and `ParameterSets` property

and the difference between 
"powerShell/getAllCommands"
and 
"powerShell/getCommand"

is just using the `param` object... could we combine these into 1 message?

`"powerShell/getCommand"` that... if param is empty, will do `Get-Command` and if it's not empty will do `Get-Command foo`?

I know this is a refactor... but I hope that my comment makes sense that it would be nice to have 1 message and one data type rather than 2 that do similar things.

tylerl0706: pls don't hate me

tylerl0706:
@corbob is this ready for us to take a look? I have a good feeling :)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -465,6 +467,36 @@ RequestContext<object> requestContext
 
             await requestContext.SendResult(result.First().ToString());
         }
+        private async Task HandleGetCommandRequest(
```
rkeithhill: Add single blank line between method definitions.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -465,6 +467,36 @@ RequestContext<object> requestContext
...
+            string param,
+            RequestContext<object> requestContext)
+        {
+            PSCommand psCommand = new PSCommand();
+            if(param != "") {
+                psCommand.AddCommand("Get-Command").AddArgument(param);
```
rkeithhill: I wonder if we should module prefix these commands e.g. `Microsoft.PowerShell.Core\Get-Command`.  We've already had to do this to fix bugs that folks submitted because somewhere they were using a command proxy that didn't work quite right.  :-(

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -465,6 +467,36 @@ RequestContext<object> requestContext
...
+            } else {
+                psCommand.AddScript("Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
```
rkeithhill: Ditto use of command prefixes.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: Consider it done.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: Consider it done.

corbob:
I think I've got most of the stuff. Only thing I'm not 100% on is some of the comment based documentation parts.

tylerl0706:
@corbob it looks like there are a couple comments still in this PR. Can you take a look and see if you've covered them and resolve them/comment on them?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -465,6 +467,36 @@ RequestContext<object> requestContext
...
             await requestContext.SendResult(result.First().ToString());
         }
+        private async Task HandleGetCommandRequest(
+            string param,
+            RequestContext<object> requestContext)
+        {
+            PSCommand psCommand = new PSCommand();
+            if(param != "") {
+                psCommand.AddCommand("Get-Command").AddArgument(param);
+            } else {
```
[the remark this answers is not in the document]
corbob: I don't get how I missed this comment. This has been rectified.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: I don't get how I missed this comment. This has been rectified.

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandsRequest.cs
corbob: Pretty sure this is taken care of by removing GetAllCommands. Going to mark it as resolved and hope for the best 🤣

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: I'm going to side with @rkeithhill on this one. Dynamic may be syntactic sugar, but I tried to set it up with `command.Properties["PropertName"].Value` and needed a lot of `.ToString()` and then there's two I haven't resolved yet because they require explicit casting. IMHO it is harder to follow:

```C#
foreach (PSObject command in result)
                {
                    commandList.Add(new PSCommandMessage
                    {
                        Name = command.Properties["Name"].Value.ToString(),
                        ModuleName = command.Properties["ModuleName"].Value.ToString(),
                        Parameters = command.Properties["Parameters"].Value, # This line is missing an explicit cast according to vscode
                        ParameterSets = command.Properties["ParameterSets"].Value, # This line also missing explicit cast.
                        DefaultParameterSet = command.Properties["DefaultParameterSet"].Value.ToString()
                    });
                }
```

@tylerl0706 / @rjmholt are we OK to leave it with dynamic? (I've expanded it from c to command so it's a littler clearer what we're working with.

rjmholt: I'm in favour of `dynamic` usage

TylerLeonhardt: same

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -523,6 +525,39 @@ RequestContext<object> requestContext
             await requestContext.SendResult(result.First().ToString());
         }
 
+        private async Task HandleGetCommandRequest(
+            string param,
+            RequestContext<object> requestContext)
+        {
+            PSCommand psCommand = new PSCommand();
+            if (param != "") {
```
rjmholt: How about `string.IsNullOrEmpty(param)`?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -523,6 +525,39 @@ RequestContext<object> requestContext
...
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command").AddArgument(param);
+            }
+            else
+            {
+                psCommand.AddScript("Microsoft.PowerShell.Core\\Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
+            }
+            IEnumerable<PSObject> result = await this.editorSession.PowerShellContext.ExecuteCommand<PSObject>(psCommand);
+            List<PSCommandMessage> commandList = new List<PSCommandMessage>();
```
rjmholt: Can use `var` on the LHS here

src/PowerShellEditorServices.Protocol/LanguageServer/GetCommandRequest.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol;
+using System.Collections.Generic;
+using System.Management.Automation;
+
+namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
+{
+
```
TylerLeonhardt: nit: extra newline here should be removed

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -523,6 +525,39 @@ RequestContext<object> requestContext
...
+            string param,
+            RequestContext<object> requestContext)
+        {
+            PSCommand psCommand = new PSCommand();
+            if (param != "") {
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command").AddArgument(param);
+            }
+            else
+            {
+                psCommand.AddScript("Microsoft.PowerShell.Core\\Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
```
TylerLeonhardt: nit: AddScript can be evil. It's better to use AddCommand when possible

```csharp
psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command")
    .AddParameter("CommandType", new[]{ "Function", "Cmdlet", "ExternalScript" })
    .AddCommand("Select-Object")
    .AddParameter("Property", new[] { "Name", "ModuleName" })
    .AddCommand("Sort-Object")
    .AddParamater("Property", "Name");
```

but haven't tested it 😄

corbob: I've added this, it appears to be working, but I can't seem to attach the debugger at the moment to test... so I haven't fully tested this either 😜

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: Sorry... One more comment I forgot to mention...

In the v2 branch, async functions have a been renamed to have Async at the end. Can you add it to this one?

```
HandleGetCommandRequest
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -523,6 +525,45 @@ RequestContext<object> requestContext
...
+            }
+            else
+            {
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command")
+                        .AddParameter("CommandType", new[]{"Function", "Cmdlet", "ExternalScript"})
+                    .AddCommand("Select-Object")
+                        .AddParameter("Property", new[]{"Name", "ModuleName"})
+                    .AddCommand("Sort-Object")
+                        .AddParameter("Property", "Name");
+                // psCommand.AddScript("Microsoft.PowerShell.Core\\Get-Command -CommandType Function, Cmdlet, ExternalScript | Select-Object Name,ModuleName | Sort-Object Name");
```
TylerLeonhardt: Nit: delete commented out code

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -523,6 +525,45 @@ RequestContext<object> requestContext
...
+            if (!string.IsNullOrEmpty(param)) {
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command").AddArgument(param);
+            }
+            else
+            {
+                psCommand.AddCommand("Microsoft.PowerShell.Core\\Get-Command")
+                        .AddParameter("CommandType", new[]{"Function", "Cmdlet", "ExternalScript"})
+                    .AddCommand("Select-Object")
+                        .AddParameter("Property", new[]{"Name", "ModuleName"})
+                    .AddCommand("Sort-Object")
```
TylerLeonhardt: Ah. Totally my fault on this one... We should add the namespace to this command and the command two lines up.

Not sure what it is off the top of my head. Check by doing:

```powershell
Get-Command Sort-Object
```

corbob: As much as I would love to let you take the fall for this one, when I was typing it out I looked at it and thought 🤔 "These should probably be fully qualified" 🤔 
But then I though "Nah forget it! Yo homes to Bel Air!"
🎵  "I pulled up to a PR about 7 or 8..."


FWIW: both `Sort-Object` and `Select-Object` are in `Microsoft.PowerShell.Utility`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: Able to attach debugger in vscode normal. This looks like I expected it to, and works too.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: sssshhhhhhh this is not the code you're looking for 👋

PR #698, opened 2018-07-11
Fix issue with stat format on Linux

rjmholt:
The `stat` usage in the startup script currently does not work on Linux.

This changes the `%A` to `%a` to make it work.

### DIFF
--- a/module/PowerShellEditorServices/Start-EditorServices.ps1
+++ b/module/PowerShellEditorServices/Start-EditorServices.ps1
@@ -224,7 +224,7 @@ function Set-NamedPipeMode {
     )
     chmod $DEFAULT_USER_MODE $PipeFile
     if ($IsLinux) {
-        $mode = stat -c "%A" $PipeFile
+        $mode = stat -c "%a" $PipeFile
     }
     else {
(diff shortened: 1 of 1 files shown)


module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -224,7 +224,7 @@ function Set-NamedPipeMode {
     )
     chmod $DEFAULT_USER_MODE $PipeFile
     if ($IsLinux) {
-        $mode = stat -c "%A" $PipeFile
+        $mode = stat -c "%a" $PipeFile
```
rkeithhill: Out of my depth here but I assume we know that %A works on macOS?  Also does the `else` below execute on Windows?  Or maybe we don't call this function at all on Windows?

rjmholt: `stat -f "%a"` on macOS gives me something totally different

rjmholt: `stat -f "%A"` gives me e.g. `700`

PR #721, opened 2018-08-10
Change Get-Help behavior to return local help when no online help available

corbob:
Current behavior is to attempt to show online help. This returns an error when there is no online help. Would be better to return full output of online help if there is some.

Furthermore: if you try to get help for a command that doesn't exist, currently it takes 4-5 seconds and doesn't do anything. This will change that to less than a second and return an error.

Addresses some of the discussion around https://github.com/PowerShell/vscode-powershell/issues/884


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -237,14 +237,42 @@ protected Task Stop()
...
+                psCommand.AddArgument(helpParams);
+                psCommand.AddCommand("Select-Object").AddArgument("RelatedLinks");
+                IEnumerable<object> relatedLinks = await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
+                psCommand = new PSCommand();
+                psCommand.AddCommand("Get-Help");
+                psCommand.AddArgument(helpParams);
+                bool sendOutput = false;
+                // Check if the first returned item converted to string is an empty object.
+                // From experience this is the return when there are no RelatedLinks.
+                if (relatedLinks != null && relatedLinks.First().ToString() != "@{RelatedLinks=}")
```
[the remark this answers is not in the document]
tylerl0706: another option (which I'm liking better) is to convert 95% of this logic into a PowerShell script and just use `AddScript`. Especially since this function doesn't return anything, this would be a good time to do it.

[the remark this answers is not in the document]
corbob: Funny, I asked about the related links in slack and was told that there would never be a related link without a online version. Clearly this is false, so this needs to be fixed. I like the idea of the powershell script instead. I might even still have one in the commit history (or I might have removed it...)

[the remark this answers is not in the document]
SeeminglyScience: Drilling down a PSObject is kind of obtuse in C#, but it's still probably better to keep it in C#.  I'd just make it a static helper method like this

```csharp
internal static bool TryGetHelpUri(PSObject helpInfo, out string helpUri)
{
    PSObject links = helpInfo.Properties['relatedLinks']?.Value as PSObject;
    if (links == null)
    {
        return false;
    }

    PSObject[] navigation = links.Properties['navigationLink']?.Value as PSObject[];
    if (navigation == null)
    {
        return false;
    }

    helpUri = navigation.FirstOrDefault()?.Properties['uri']?.Value as string;
    return !string.IsNullOrWhiteSpace(helpUri);
}
```

```csharp
PSObject helpInfo = MethodThatInvokesGetHelp();
if (TryGetHelpUri(helpInfo, out string helpUri)
{
    // use help uri
}
```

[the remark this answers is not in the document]
tylerl0706: Yeah that's better. 👌

corbob:
I've refactored the code to use a PowerShell function to check for if the command exists instead of executing a bunch of times and then delving into a property that doesn't seem to be completely consistent.

The PowerShell function operates off of the presumption that the Get-Command and Get-Help cmdlets will error with predictable exceptions and we catch those exceptions only for our handling.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
...
+        if ($CommandExists) {
+            try {
+                Get-Help $Command -Online
+            } catch [System.Management.Automation.PSInvalidOperationException] {
+                Get-Help $Command -Full
+            }
+        }
+}";
+            PSCommand checkHelpScript = new PSCommand();
+            checkHelpScript.AddScript(script);
```
SeeminglyScience: After removing the `function` part of the script, change this to add the argument here instead and remove the second command invocation.  Also add `useLocalScope: true` to `AddScript` (so the variables don't leak into global)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
...
-            psCommand.AddCommand("Get-Help");
-            psCommand.AddArgument(helpParams);
-            psCommand.AddParameter("Online");
-
-            await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
-
+            string script = @"
+function __Check-Help {
+    param (
+        $Command
```
SeeminglyScience: Should be `[string] $Command`.  Also would prefer the name `CommandName`, since you won't be passing it a command object.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
...
+    )
+        $CommandExists = $false
+        Try {
+            Get-Command $Command -ErrorAction Stop
```
SeeminglyScience: Commands need to be module qualified (we've been burned on skipping that before) so this should be:

```powershell
Microsoft.PowerShell.Core\Get-Command $Command -ErrorAction Stop
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
...
+            $CommandExists = $true
+        } catch [System.Management.Automation.CommandNotFoundException] {
+            Write-Warning ""$Command does not exist.""
+        }
+        if ($CommandExists) {
+            try {
+                Get-Help $Command -Online
```
SeeminglyScience: echo: Module qualified, e.g. `Microsoft.PowerShell.Core\Get-Help`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
             RequestContext<object> requestContext)
         {
             if (helpParams == null) { helpParams = "get-help"; }
```
SeeminglyScience: I know this isn't part of your changes, but that casing is driving me nuts ;)

Could you make that `Get-Help` while you're in there?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -236,14 +236,32 @@ protected Task Stop()
...
-
-            var psCommand = new PSCommand();
-            psCommand.AddCommand("Get-Help");
-            psCommand.AddArgument(helpParams);
-            psCommand.AddParameter("Online");
-
-            await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
-
+            string script = @"
+function __Check-Help {
```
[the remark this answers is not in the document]
corbob: Works for me. It's another instance of copying from elsewhere (Expand-Alias). But just because it's in use elsewhere doesn't mean we should keep doing it that way 😄

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -235,15 +235,25 @@ protected Task Stop()
...
+            if (helpParams == null) { helpParams = "Get-Help"; }
+            string script = @"
+[CmdletBinding()]
+param (
+    [String]$CommandName
+)
+Try {
+    Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop | Out-Null
+} catch [System.Management.Automation.CommandNotFoundException] {
+    $PSCmdlet.ThrowTerminatingError($PSItem)
```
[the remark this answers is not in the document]
tylerl0706: in other words you might be able to ditch the try/catch all together.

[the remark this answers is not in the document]
corbob: I like it with the try catch because without it looks like this:
image

and with it looks like this:
image

But I will go with whichever you and @SeeminglyScience want.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -235,15 +235,25 @@ protected Task Stop()
...
-
-            var psCommand = new PSCommand();
-            psCommand.AddCommand("Get-Help");
-            psCommand.AddArgument(helpParams);
-            psCommand.AddParameter("Online");
-
-            await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
-
+            if (helpParams == null) { helpParams = "Get-Help"; }
+            string script = @"
```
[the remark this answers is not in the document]
corbob: I've moved this to a static private property, not sure if positioning it at the top of the file is the best spot for it, or what the convention is for this type of scenario.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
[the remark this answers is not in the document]
SeeminglyScience: @tylerl0706 the benefit of using `ThrowTerminatingError` over letting the exception propagate is that it changes the "context" of the error from the callee(?) to the caller.

Here's a little function you can use to check it out

```powershell
function Test-TerminatingError {
    [CmdletBinding()]
    param(
        [Parameter(Position = 0, Mandatory)]
        [ValidateSet('NormalThrow', 'ThrowTerminatingError')]
        [string] $ThrowType
    )
    end {
        if ($ThrowType -eq 'NormalThrow') {                j
            Get-Command DefinitelyDoesntExist -ErrorAction Stop
        }

        try {
            Get-Command DefinitelyDoesntExist -ErrorAction Stop
        } catch {
            $PSCmdlet.ThrowTerminatingError($PSItem)
        }
    }
}
```

If `NormalThrow` is specified, this is what is displayed

```raw
Get-Command : The term 'DefinitelyDoesntExist' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At line:11 char:13
+             Get-Command DefinitelyDoesntExist -ErrorAction Stop
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (DefinitelyDoesntExist:String) [Get-Command], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand
```

It shows the line in the function/script itself that failed, which is implementation detail the user doesn't care about.  Contrast that to this:

```raw
Test-TerminatingError : The term 'DefinitelyDoesntExist' is not recognized as the name of a cmdlet, function, script
file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct
and try again.
At line:1 char:1
+ Test-TerminatingError ThrowTerminatingError
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (DefinitelyDoesntExist:String) [Test-TerminatingError], CommandNotFoundE
   xception
    + FullyQualifiedErrorId : CommandNotFoundException,Test-TerminatingError
```

It also lets you fully (well mostly) customize the displayed error by creating your own `ErrorRecord`

All that said, @corbob that is definitely not how I expected that to look...  something in our host does *not* handle that right...

We should still stick with it, but we should definitely look into why that's doing that.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -42,6 +42,21 @@ public class LanguageServer
 
         private TaskCompletionSource<bool> serverCompletedTask;
 
+        private static string checkHelpScript = @"
+[CmdletBinding()]
+param (
+    [String]$CommandName
+)
+Try {
+    Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop | Out-Null
```
SeeminglyScience: `Out-Null` as well :/ I know it's a pain but I'm pretty sure the last one we ran into was something like a different `Set-Content` that actually shipped with a product. I wouldn't be shocked if there was an `Out-Null` out there with mandatory parameters or something else breaking.

Or just do `$null = Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -42,6 +42,21 @@ public class LanguageServer
...
+} catch [System.Management.Automation.CommandNotFoundException] {
+    $PSCmdlet.ThrowTerminatingError($PSItem)
+}
+try {
+    Microsoft.PowerShell.Core\Get-Help $CommandName -Online | Out-Null
```
SeeminglyScience: echo: Out-Null

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,20 +245,15 @@ protected Task Stop()
...
-            var psCommand = new PSCommand();
-            psCommand.AddCommand("Get-Help");
-            psCommand.AddArgument(helpParams);
-            psCommand.AddParameter("Online");
-
-            await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
+            if (helpParams == null) { helpParams = "Get-Help"; }
 
+            PSCommand checkHelpPSCommand = new PSCommand();
+            checkHelpPSCommand.AddScript(checkHelpScript,true).AddArgument(helpParams);
```
SeeminglyScience: Use explicit argument name for `bool` parameters.  (VSCode has a code action for this if you put your cursor on `true` and press <kbd>CTRL + .</kbd> you should see it)

e.g.
```csharp
checkHelpPSCommand.AddScript(checkHelpScript, useLocalScope: true).AddArgument(helpParams);
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,20 +245,15 @@ protected Task Stop()
...
+            await editorSession.PowerShellContext.ExecuteCommand<PSObject>(checkHelpPSCommand, true);
```
SeeminglyScience: echo: named bool arguments

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -42,6 +42,21 @@ public class LanguageServer
 
         private TaskCompletionSource<bool> serverCompletedTask;
 
+        private static string checkHelpScript = @"
```
SeeminglyScience: We want to use a constant here since `string` can be a constant and this won't change.

```csharp
private const string CheckHelpScript = @"";
```

<sup>(**Note:** There no real guidance for casing of *private* constants so I usually go with the same casing of public constants which is typically Pascal.  Unless it's an interop constant, then keep casing of the original)</sup>

For order, I typically sort fields by access first, then I do:

1. Constants
1. Static readonly
1. Static
1. Instance readonly
1. Instance

corbob: Changed casing. 
Moved to top of class.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: 👍

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: Yeah, I debated Out-Null or just assigning to $null. I think I recall reading somewhere that assignment is faster (but not by much, especially for this type of command)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: Copy pasta:
Yeah, I debated Out-Null or just assigning to $null. I think I recall reading somewhere that assignment is faster (but not by much, especially for this type of command)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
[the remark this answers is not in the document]
corbob: @SeeminglyScience should we be opening an issue for "something in our host does not handle that right...
...
we should definitely look into why that's doing that." ?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rjmholt: Casting to `[void]` or assigning to `$null` is faster in Windows PowerShell, but I recall reading that we now recoginise the pipe-to-`Out-Null` case in PS Core and have optimised it out too. Can't find a source on that though.

rjmholt: https://stackoverflow.com/questions/5260125/whats-the-better-cleaner-way-to-ignore-output-in-powershell is one source

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,20 +245,15 @@ protected Task Stop()
...
-
-            var psCommand = new PSCommand();
-            psCommand.AddCommand("Get-Help");
-            psCommand.AddArgument(helpParams);
-            psCommand.AddParameter("Online");
-
-            await editorSession.PowerShellContext.ExecuteCommand<object>(psCommand);
+            if (helpParams == null) { helpParams = "Get-Help"; }
 
+            PSCommand checkHelpPSCommand = new PSCommand();
```
rjmholt: This command instantiation could probably be streamlined a bit into:
```csharp
var checkHelpPSCommand = new PSCommand()
    .AddScript(CheckHelpScript, useLocalScope: true)
    .AddArgument(helpParams);
```

corbob: chain all the things?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -251,8 +251,9 @@ protected Task Stop()
         {
             if (helpParams == null) { helpParams = "Get-Help"; }
 
-            PSCommand checkHelpPSCommand = new PSCommand();
-            checkHelpPSCommand.AddScript(CheckHelpScript, useLocalScope: true).AddArgument(helpParams);
+            PSCommand checkHelpPSCommand = new PSCommand()
```
rjmholt: I was about to complain about the lack of `var` here, but technically the type here is the return type of the method. So all I feel is ambivalence.

corbob: I used PSCommand due to all the other comments about being explicit about the variable type 😁

src/PowerShellEditorServices.Protocol/LanguageServer/ShowHelpRequest.cs
```diff
@@ -7,10 +7,10 @@
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
-    public class ShowOnlineHelpRequest
+    public class ShowHelpRequest
     {
         public static readonly
             RequestType<string, object, object, object> Type =
-            RequestType<string, object, object, object>.Create("powerShell/showOnlineHelp");
+            RequestType<string, object, object, object>.Create("powerShell/showHelp");
```
rjmholt: Just to play Devil's advocate here:

Is the reason for the breaking change just that "showOnlineHelp" would be an inaccurate name? Or does possibly getting offline help make the change breaking too?

Just thinking that there are plenty of APIs where the message name doesn't quite mean what it used to because changing the message name is a bigger breaking change than changing what the message does.

It may be that we want to break the message compatibility anyway, but we do have other clients that might benefit from us not changing the name.

Anyway, just want to ask the question: "if we change the functionality but not the message name, is this still a breaking change?" with the followup "if the message name change is the breaking change, is it worth the breakage?"

corbob:
Last minute commit that hopefully doesn't change the approvals 😜 

I was looking at the overall changes and was reminded of a scenario where I found if nothing was sent across it was actually empty string and not null (discovered while working on the Command Explorer). I believe the only way for this to happen with the Show Help is if we have a completely empty file and try, but this at least prevents an error in that instance.

src/PowerShellEditorServices.Protocol/LanguageServer/ShowHelpRequest.cs
tylerl0706: given this is a command in the command pallet and not a settings related change or something along those lines... I think if someone complains we can consider a patch that will re-ADD the old functionality.

corbob: I think what @rjmholt is talking about is breaking on both sides with the change of the message as PSES isn't used exclusively by vscode-powershell. I'm not opposed to keeping the message name the old one (in fact I think it should be a matter of just backing out a single commit on both sides...)

rjmholt: That sounds reasonable!

tylerl0706: I'm game 👍

corbob: oh sure, now I need to first remove my big foot from my mouth and figure out how to rewrite the git history...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rjmholt: Wholly agree. In general we lose no explicit variable type constraint with `var` when the constructor is on the same line (`var x = new Banana()`), but here it's more nuanced because of a sort of Builder pattern scenario...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -249,7 +249,7 @@ protected Task Stop()
             string helpParams,
             RequestContext<object> requestContext)
         {
-            if (helpParams == null) { helpParams = "Get-Help"; }
+            if (helpParams == "") { helpParams = "Get-Help"; }
```
rjmholt: Might be worth just making this `string.IsNullOrEmpty(helpParams)`. We lose nothing really if the string is never `null`

src/PowerShellEditorServices.Protocol/LanguageServer/ShowHelpRequest.cs
corbob: the interactive rebase allowed me to delete the commits that changed the wording of the message (which is partly why I made them separate commits) while still preserving the super difficult minor bug fix commit at the end there.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: I don't know what you're talking about... it's definitely string.IsNullOrEmpty *whistles innocently*

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
SeeminglyScience: Yeah it's faster at scale, but with a single item it doesn't matter.  I typically default to `$null` assignment but I've been trying to break the habit since `Out-Null` is significantly easier to explain to beginners.

(I'm not suggesting you change it back, I'd rather `$null = ` than the fully qualified name every time 😄)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -26,6 +26,21 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 {
     public class LanguageServer
     {
+        private const string CheckHelpScript = @"
+[CmdletBinding()]
+param (
+    [String]$CommandName
+)
+Try {
+    $null = Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop
```
rkeithhill: 👍 for module prefixed commands!

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -26,6 +26,21 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 {
     public class LanguageServer
     {
+        private const string CheckHelpScript = @"
+[CmdletBinding()]
+param (
+    [String]$CommandName
+)
+Try {
```
rkeithhill: Below you use a lower case `try` (and everywhere else uses lower case keywords).  Probably should tweak that here to be consistent.  :-)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: If this script is only used in the `HandleShowOnlineHelpRequest` method, then it should be moved inside that method as a `const string`.

corbob: This seems to make sense. As was said in slack: if it's a `const string` it won't be stack alloc'd each time the methos is invoked. https://stackoverflow.com/questions/7616448/const-string-inside-a-function

Unless @tylerl0706 @SeeminglyScience or @rjmholt have any objections I'll move the string back into the function as `const`...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
corbob: I'll get this corrected after the below comment is replied to so I'm not doing a bunch of extra work :)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
tylerl0706: Reasonable to me! Let's hear what 1 other has to say

SeeminglyScience: Yep no issues with that 👍

corbob: Moved the string into the function as const

(also creeped the scope a bit by handling both commands that could come in (although not from vscode-powershell...) and marking one as deprecated.

src/PowerShellEditorServices.Protocol/LanguageServer/ShowOnlineHelpRequest.cs
```diff
@@ -7,10 +7,18 @@
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
+    // We don't expect ShowOnlineHelpRequest to come from vscode anymore, but it could come from another editor.
+    // TODO: Note that it's deprecated if it's called???
```
rjmholt: You might be able to mark this with the `[Obsolete]` attribute :)

src/PowerShellEditorServices.Protocol/LanguageServer/ShowOnlineHelpRequest.cs
```diff
@@ -7,10 +7,18 @@
...
     public class ShowOnlineHelpRequest
     {
         public static readonly
             RequestType<string, object, object, object> Type =
             RequestType<string, object, object, object>.Create("powerShell/showOnlineHelp");
     }
+        public class ShowHelpRequest
```
tylerl0706: nitpick: the indentation of this line is a bit off

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,22 +231,45 @@ protected Task Stop()
                     }
                 });
         }
-
-        protected async Task HandleShowOnlineHelpRequest(
+        protected async Task HandleShowHelpRequest(
```
tylerl0706: nitpick: newline between functions

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,22 +231,45 @@ protected Task Stop()
...
+        protected async Task HandleShowOnlineHelpRequest(
+            string helpParams,
+            RequestContext<object> requestContext
+        )
+        {
+            const string deprecatedWarning = @"
+            Write-Warning ""'powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.""";
+            PSCommand commandDeprecated = new PSCommand()
+                .AddScript(deprecatedWarning);
+            await editorSession.PowerShellContext.ExecuteCommand<PSObject>(commandDeprecated, sendOutputToHost: true);
```
rkeithhill: Instead of (or in addition to) a PSIC warning, we could log a warning to the log file.  I could go either way on this.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -257,6 +258,18 @@ protected Task Stop()
             await editorSession.PowerShellContext.ExecuteCommand<PSObject>(checkHelpPSCommand, sendOutputToHost: true);
             await requestContext.SendResult(null);
         }
+        protected async Task HandleShowOnlineHelpRequest(
+            string helpParams,
+            RequestContext<object> requestContext
+        )
+        {
+            const string deprecatedWarning = @"
+            Write-Warning ""'powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.""";
```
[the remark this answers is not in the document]
corbob: It does 😃 it's due to `sendOutputToHost: true`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rjmholt: 👍 on log file

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -230,22 +231,45 @@ protected Task Stop()
...
+            await editorSession.PowerShellContext.ExecuteCommand<PSObject>(checkHelpPSCommand, sendOutputToHost: true);
             await requestContext.SendResult(null);
         }
+        protected async Task HandleShowOnlineHelpRequest(
+            string helpParams,
+            RequestContext<object> requestContext
+        )
+        {
+            const string deprecatedWarning = @"
+            Write-Warning ""'powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.""";
```
SeeminglyScience: Instead of `AddScript` you could use `AddCommand` here instead.

```csharp
pwsh.AddCommand("Microsoft.PowerShell.Utility\\Write-Verbose")
    .AddParameter("Message", "'powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp instead.");
```

Using command instead allows it to create a command processor for that command specifically, instead of having to parse a script.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -263,10 +264,9 @@ protected Task Stop()
             RequestContext<object> requestContext
         )
         {
-            const string deprecatedWarning = @"
-            Write-Warning ""'powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.""";
             PSCommand commandDeprecated = new PSCommand()
-                .AddScript(deprecatedWarning);
+                .AddCommand("Microsoft.PowerShell.Utility\\Write-Verbose")
+                .AddParameter("Message", ";powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.");
```
SeeminglyScience: Typo here I think (from <kbd>;</kbd> to <kbd>'</kbd>)

src/PowerShellEditorServices.Protocol/LanguageServer/ShowHelpRequest.cs
```diff
@@ -4,13 +4,21 @@
...
 
 namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 {
+    [Obsolete("This class is deprecated. Use ShowHelpRequest instead.")]
     public class ShowOnlineHelpRequest
     {
         public static readonly
             RequestType<string, object, object, object> Type =
             RequestType<string, object, object, object>.Create("powerShell/showOnlineHelp");
     }
```
rjmholt: Can you add a newline here?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -122,6 +122,7 @@ public void Start()
                 this.HandleDocumentRangeFormattingRequest);
 
             this.messageHandlers.SetRequestHandler(ShowOnlineHelpRequest.Type, this.HandleShowOnlineHelpRequest);
+            this.messageHandlers.SetRequestHandler(ShowHelpRequest.Type, this.HandleShowHelpRequest);
```
rjmholt: Maybe add newlines around this line to keep the style consistent?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -231,21 +232,44 @@ protected Task Stop()
...
+                    $null = Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop
+                } catch [System.Management.Automation.CommandNotFoundException] {
+                    $PSCmdlet.ThrowTerminatingError($PSItem)
+                }
+                try {
+                    $null = Microsoft.PowerShell.Core\Get-Help $CommandName -Online
+                } catch [System.Management.Automation.PSInvalidOperationException] {
+                    Microsoft.PowerShell.Core\Get-Help $CommandName -Full
+                }";
+            if (string.IsNullOrEmpty(helpParams)) { helpParams = "Get-Help"; }
```
rjmholt: Maybe add a newline above this `if`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -231,21 +232,44 @@ protected Task Stop()
...
+
+            PSCommand checkHelpPSCommand = new PSCommand()
+                .AddScript(CheckHelpScript, useLocalScope: true)
+                .AddArgument(helpParams);
+            await editorSession.PowerShellContext.ExecuteCommand<PSObject>(checkHelpPSCommand, sendOutputToHost: true);
```
rjmholt: I would add a newline above the first `await` statement

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -231,21 +232,44 @@ protected Task Stop()
...
             await requestContext.SendResult(null);
         }
+        protected async Task HandleShowOnlineHelpRequest(
```
rjmholt: Definitely add a newline before this method definition here

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -231,21 +232,44 @@ protected Task Stop()
...
+            string helpParams,
+            RequestContext<object> requestContext
+        )
+        {
+            PSCommand commandDeprecated = new PSCommand()
+                .AddCommand("Microsoft.PowerShell.Utility\\Write-Verbose")
+                .AddParameter("Message", ";powerShell/showOnlineHelp' has been deprecated. Use 'powerShell/showHelp' instead.");
+            await editorSession.PowerShellContext.ExecuteCommand<PSObject>(commandDeprecated, sendOutputToHost: true);
```
rjmholt: Maybe add a newline above this `await`

rjmholt:
This doesn't seem to be working now that it's merged in. Going to investigate today.

rjmholt:
Can't reproduce the behaviour I saw earlier.

But it looks like `Get-Help -Online` doesn't fail if there's no internet or the help site is down.

What scenarios would trigger the offline help?

I'm trying to think of a way to identify when online help is inaccessible and fall back appropriately.

rjmholt:
Spoke to @adityapatwardhan offline.

`Get-Help -Online` won't give us any indication if getting online help failed -- it just passes a URL to the browser and forgets. It might give an error if there's no browser, but I can't imagine a common scenario where VSCode is installed but there's no browser.

There's also no simple way to get the help URL out of `Get-Help` to test it ourselves.

He suggested getting local help first and falling back to online instead.

The problem is the `Get-Help` is designed never to throw, so you get a generated help object back instead. **But**, that object won't have the `Description` property populated. So that could be a check.

Ideally what we turn this into would be:
- <kbd>Ctrl</kbd>+<kbd>F1</kbd> gets local help first.
- If description is populated, we turn the whole thing into a string (with `Out-String`) and send it in a message to VSCode to put in a pane.
- Otherwise, we run `Get-Help -Online` and return `null` (in which case VSCode knows to do nothing).

SeeminglyScience:
@rjmholt 

> There's also no simple way to get the help URL out of Get-Help to test it ourselves.

What about `Microsoft.PowerShell.Commands.GetHelpCodeMethods.GetHelpUri(PSObject)`?

**Note:** The `PSObject` required is a `CommandInfo` object (like from `Get-Command`), and it needs to be invoked from the pipeline thread.

PR #729, opened 2018-08-22
Fix a null dereference when an invalid cast exception has no inner exception

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/1450.

When a parameter causes an invalid cast exception in the console, we assumed the inner exception wasn't null. Now if it is, we pass through the outer exception.


rkeithhill:
Darn little niggly `null` bugs.  I'm looking forward to trying out C# 8's non-nullable reference types (assuming the feature makes the cut).  Maybe not in this project but somewhere.  :-)

PR #736, opened 2018-08-31
Cache the reflection call done for completions

rjmholt:
I noticed that completion queries do a reflection call every time. This code caches instead, but it's not terribly elegant. So if you have any suggestions, please let me know.

### DIFF
--- a/src/PowerShellEditorServices/Language/AstOperations.cs
+++ b/src/PowerShellEditorServices/Language/AstOperations.cs
@@ -5,19 +5,18 @@
 
 using Microsoft.PowerShell.EditorServices.Utility;
 using System;
+using System.Diagnostics;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Management.Automation.Language;
+using System.Management.Automation.Runspaces;
 
 namespace Microsoft.PowerShell.EditorServices
 {
-    using System.Diagnostics;
-    using System.Linq.Expressions;
     using System.Management.Automation;
-    using System.Management.Automation.Language;
-    using System.Management.Automation.Runspaces;
 
     /// <summary>
     /// Provides common operations for the syntax tree of a parsed script.
--- a/src/PowerShellEditorServices/Language/AstOperations.cs
+++ b/src/PowerShellEditorServices/Language/AstOperations.cs
@@ -23,28 +23,11 @@ namespace Microsoft.PowerShell.EditorServices
     /// </summary>
     internal static class AstOperations
     {
-        private static readonly MethodInfo s_extentCloneWithNewOffset;
-
-        static AstOperations()
-        {
-            // TODO: When netstandard is upgraded to 2.0, see if
-            //       Delegate.CreateDelegate can be used here instead
-            s_extentCloneWithNewOffset =
-#if CoreCLR
-                typeof(PSObject).GetTypeInfo().Assembly
-                    .GetType("System.Management.Automation.Language.InternalScriptPosition")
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Language/AstOperations.cs
```diff
@@ -337,5 +326,28 @@ static public string[] FindDotSourcedIncludes(Ast scriptAst)
 
             return dotSourcedVisitor.DotSourcedFiles.ToArray();
         }
+
+        private static MethodInfo GetExtentCloneMethod(Ast scriptAst)
```
SeeminglyScience: I think this would be better as just a static field.  If your looking for how you could do that without an instance, you could do this:

```csharp
private static MethodInfo s_extentCloneWithNewOffset =
    typeof(PSObject).Assembly.GetType("System.Management.Automation.InternalScriptExtent")
        .GetMethod(...
```

Maybe set in a static constructor because of the compiler directives.

rjmholt: Yeah I was wondering what the best way to do this was. Figured we'd find a better way in the PR. Thanks!

rjmholt:
Ok, I've updated to use a better way to get the MethodInfo.

Once we're on netstandard2.0, we might be able to improve on the `object[]` allocation with each invocation by creating a delegate. I'm not quite sure if it's possible since `InternalScriptPosition` is internal and thus hard to pass around a `Func<>` for, but it can be created. See:
- https://stackoverflow.com/questions/38111987/using-reflection-methods-that-take-array-parameters-efficiently
- https://stackoverflow.com/questions/940675/getting-a-delegate-from-methodinfo

You can model what I'm imagining doing with this PowerShell:
```powershell
$cwno = [psobject].GetTypeInfo().Assembly.GetType("System.Management.Automation.Language.InternalScriptPosition", $true).GetMethod("CloneWithNewOffset", [System.Reflection.BindingFlags]::Instance -bor [System.Reflection.BindingFlags]::NonPublic)

$t = [System.Linq.Expressions.Expression]::GetFuncType(@($cwno.ReturnType, [int], $cwno.ReturnType))

[System.Delegate]::CreateDelegate([func[object, int, object]], $null, $cwno)   
```

That should create the open delegate we want. The problem is just being able to invoke it when we can't express the type of the `Func<>` statically.

SeeminglyScience:
You could craft the delegate using LINQ expressions, e.g.

```powershell
using namespace System.Linq.Expressions
using namespace System.Management.Automation.Language

$scriptPositionType = [psobject].Assembly.
    GetType('System.Management.Automation.Language.InternalScriptPosition')

$method = $scriptPositionType.
    GetMethod(
        'CloneWithNewOffset',
        [System.Reflection.BindingFlags]'NonPublic, Instance',
        $null,
        [type[]][int],
        @(1))

$source = [Expression]::Parameter([IScriptPosition], 'source')
$offset = [Expression]::Parameter([int], 'offset')

# (IScriptPosition)((InternalScriptPosition)source).CloneWithNewOffset(source)
$delegateBody = 
    [Expression]::Convert(
        [Expression]::Call(
            [Expression]::Convert(
                $source,
                $scriptPositionType),
            $method,
            $offset),
        [IScriptPosition])

$delegate = [Expression]::Lambda(
    [Func[IScriptPosition, int, IScriptPosition]],
    $delegateBody,
    [ParameterExpression[]]($source, $offset)).
    Compile()
```

You could also write it in IL with `DynamicMethod`.  That would be a little bit faster, but then you need to add the currently unlisted `System.Reflection.Emit` package, and readability would be pretty shot.

rjmholt:
Oooh, tests have found another breakage! Will look into it tomorrow.

PR #742, opened 2018-09-05
[Ignore] Exclude integration tests

rjmholt:
I've gotten to understand the testing layout better, and to stop AppVeyor failing for now I thought we might just exclude the integration testing that's timing out.

Making testing work is the next item on my list, so it will be back in some form.


rjmholt:
Going to merge this one so that I can restart the other CI builds...

PR #743, opened 2018-09-05
Remove unnecessary LINQ calls from LanguageServer

rjmholt:
Removes a few LINQ calls from the Language Server, in an effort to chip away at performance issues.

Given how complicated and central the logic here is, I haven't made any style changes, but would be happy to make any :)

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -34,6 +34,8 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 
         private static readonly SignatureInformation[] s_emptySignatureResult = new SignatureInformation[0];
 
+        private static readonly DocumentHighlight[] s_emptyHighlightResult = new DocumentHighlight[0];
+
         private ILogger Logger;
         private bool profilesLoaded;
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -933,34 +920,29 @@ await editorSession
...
-                                    },
-                                    Name = GetDecoratedSymbolName(r)
-                                };
-                            })
-                        .ToArray();
-            }
-            else
-            {
-                symbols = new SymbolInformation[0];
+                var symbolAcc = new List<SymbolInformation>();
```
rkeithhill: This all looks great.  Just curious what the `Acc` in `symbolAcc` is short for?  Accumulator?

rjmholt: Sorry, accumulator, yes. Habit from writing tail-recursive folds in other languages. Maybe I should change it to something more obvious? Always conscious that having so many variables named essentially the same thing is a code smell, so open to suggestions :)

rkeithhill: Not a big deal. Just wanted to make sure I understood what it meant.

PR #745, opened 2018-09-06
Use PSRL release and filter out previous betas

rjmholt:
- Uses the PSReadLine beta3 release from PowerShell gallery instead of downloading the master build from AppVeyor
- Filters out the old beta versions that don't have @SeeminglyScience's private API in it (which effectively break us trying to pull PSRL2 in)

Quick question: is there anything else that needs changing for better PSRL module support here?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,10 +14,10 @@ environment:
 
 install:
   - ps: |
-      Install-Module -Name PowershellGet,PackageManagement -force -confirm:$false -verbose
+      powershell -Command { Install-Module -Name PowershellGet,PackageManagement -force -confirm:$false -verbose }
       Get-Module PowerShellGet,PackageManagement | Remove-Module -Force -Verbose
-      Install-Module -Name PowerShellGet -MinimumVersion 1.6 -PassThru -Force
-      Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -PassThru -Force
+      Install-Module -Name PowerShellGet -MinimumVersion 1.6 -Force
+      Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Force
       Install-PackageProvider -Name NuGet -Force | Out-Null
       Import-PackageProvider NuGet -Force | Out-Null
       Set-PSRepository -Name PSGallery -InstallationPolicy Trusted | Out-Null
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,10 +14,12 @@ environment:
 
 install:
   - ps: |
+      Install-Module -Name PowershellGet,PackageManagement -force -confirm:$false -verbose
+      Get-Module PowerShellGet,PackageManagement | Remove-Module -Force -Verbose
+      Install-Module -Name PowerShellGet -MinimumVersion 1.6 -PassThru -Force
+      Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -PassThru -Force
       Install-PackageProvider -Name NuGet -Force | Out-Null
       Import-PackageProvider NuGet -Force | Out-Null
-      Install-Module -Name PowerShellGet -Force
-      Import-Module PowerShellGet -Force
       Set-PSRepository -Name PSGallery -InstallationPolicy Trusted | Out-Null
       Install-Module InvokeBuild -MaximumVersion 5.1.0 -Scope CurrentUser -Force | Out-Null
       Install-Module platyPS -RequiredVersion 0.9.0 -Scope CurrentUser -Force | Out-Null
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,10 +14,11 @@ environment:
 
 install:
   - ps: |
-      powershell -Command { Install-Module -Name PowershellGet,PackageManagement -force -confirm:$false -verbose }
       Get-Module PowerShellGet,PackageManagement | Remove-Module -Force -Verbose
-      Install-Module -Name PowerShellGet -MinimumVersion 1.6 -Force
-      Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Force
+      powershell -Command { Install-Module -Name PowershellGet -MinimumVersion 1.6 -force -confirm:$false -verbose }
+      powershell -Command { Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Force -Confirm:$false -Verbose }
+      Import-Module -Name PowerShellGet -MinimumVersion 1.6 -Force
+      Import-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Force
       Install-PackageProvider -Name NuGet -Force | Out-Null
       Import-PackageProvider NuGet -Force | Out-Null
       Set-PSRepository -Name PSGallery -InstallationPolicy Trusted | Out-Null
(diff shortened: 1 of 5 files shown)


tylerl0706:
lolololol

src/PowerShellEditorServices/Session/PSReadLinePromptContext.cs
```diff
@@ -30,7 +30,11 @@ internal class PSReadLinePromptContext : IPromptContext {
             [System.Diagnostics.DebuggerStepThrough()]
             param()
             end {
-                $module = Get-Module -ListAvailable PSReadLine | Where-Object Version -ge '2.0.0' | Sort-Object -Descending Version | Select-Object -First 1
+                $module = Get-Module -ListAvailable PSReadLine `
+                    | Where-Object Version -ge '2.0.0' `
+                    | Where-Object { $psd = Import-PowerShellDataFile $_.Path; $release = $psd.PrivateData.PSData.Prerelease; -not (@('beta1', 'beta2') -contains $release) } `
```
tylerl0706: wow

tylerl0706: I can't believe we need to do this lol

rjmholt: Yeah :cry:

rkeithhill: Maybe simplify a bit with `$release -notin @('beta1','beta2')`?  If only we had a `SemanticVersion` class we could use for sorting in Win PS 5.

rjmholt: Aha! I wondered if an operator like that existed

src/PowerShellEditorServices/Session/PSReadLinePromptContext.cs
```diff
@@ -30,7 +30,11 @@ internal class PSReadLinePromptContext : IPromptContext {
             [System.Diagnostics.DebuggerStepThrough()]
             param()
             end {
-                $module = Get-Module -ListAvailable PSReadLine | Where-Object Version -ge '2.0.0' | Sort-Object -Descending Version | Select-Object -First 1
+                $module = Get-Module -ListAvailable PSReadLine `
```
SeeminglyScience: Better to use the <kbd>|</kbd> character for line continuation here.  I like to format it like this:

```powershell
$module = Get-Module -ListAvailable PSReadLine |
    Where-Object Version -ge '2.0.0' |
    Where-Object {
        $psd = Import-PowerShellDataFile $_.Path
        $release = $psd.PrivateData.PSData.Prerelease
        $release -notin (@('beta1', 'beta2'))
    } |
    Sort-Object -Descending Version |
    Select-Object -First 1

File: src/PowerShellEditorServices/Session/PSReadLinePromptContext.cs [hunk 3]
```diff
@@ -30,7 +30,11 @@ internal class PSReadLinePromptContext : IPromptContext {
...
+                    | Where-Object Version -ge '2.0.0' `
+                    | Where-Object { $psd = Import-PowerShellDataFile $_.Path; $release = $psd.PrivateData.PSData.Prerelease; $release -notin (@('beta1', 'beta2')) } `
```
SeeminglyScience: Looks like `PSModuleInfo` has this info too, so you could do this instead:
```powershell
Where-Object { $_.PrivateData.PSData.Prerelease -notin 'beta1', 'beta2' }
```

src/PowerShellEditorServices/Session/PSReadLinePromptContext.cs
rjmholt: But my aligned pipes! Ah well, I'll take one for the team

SeeminglyScience: Hah yeah I hear ya.  I don't mind back ticks personally but the pitchforks come out **fast** for that 😄

rkeithhill: Agreed, especially when you're coding in VSCode with this setting - `"files.trimTrailingWhitespace": true,`.

rjmholt:
Anyone know why it suddenly doesn't like `-AllowPrereleaseVersions`?

rjmholt:
I've tried @tylerl0706's suggestions: https://www.powershellgallery.com/packages/PowerShellGet/1.5.0.0

Works fine on my machines though

appveyor.yml
```diff
@@ -14,7 +14,12 @@ environment:
 
 install:
   - ps: |
-      Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
+      Get-Module PowerShellGet,PackageManagement | Remove-Module -Force -Verbose
+      powershell -Command { Install-Module -Name PowershellGet -MinimumVersion 1.6 -force -confirm:$false -verbose }
+      powershell -Command { Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Force -Confirm:$false -Verbose }
```
tylerl0706: Same question about PackageManagement

scripts/travis.ps1
```diff
@@ -1,4 +1,11 @@
-
+Get-Module PowerShellGet,PackageManagement | Remove-Module -Force -Verbose
+powershell -Command { Install-Module -Name PowershellGet -MinimumVersion 1.6 -Scope CurrentUser -force -confirm:$false -verbose }
+powershell -Command { Install-Module -Name PackageManagement -MinimumVersion 1.1.7.0 -Scope CurrentUser -Force -Confirm:$false -Verbose }
```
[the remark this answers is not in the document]
rjmholt: Following @bmanikm's suggestion here https://github.com/PowerShell/PowerShellGet/issues/208#issuecomment-402260676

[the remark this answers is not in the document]
tylerl0706: why the `-Confirm:$false`? That shouldn't be needed. Interesting - alright that's fine if that's the way to do it.

src/PowerShellEditorServices/Session/PSReadLinePromptContext.cs
```diff
@@ -20,7 +20,7 @@ internal class PSReadLinePromptContext : IPromptContext {
             [System.Diagnostics.DebuggerHidden()]
             [System.Diagnostics.DebuggerStepThrough()]
             param()
-            return [Microsoft.PowerShell.PSConsoleReadLine]::ReadLine(
+            return [Microsoft.PowerShell.PSConsoleReadLine, Microsoft.PowerShell.PSReadLine2, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null]::ReadLine(
```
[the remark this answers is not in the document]
rjmholt: @rkeithhill, @SeeminglyScience turns out:
- PSCore 6.0 bundles and loaded PSReadLine.dll in $PSHome, so we need to use the assembly-qualified name of the type and we have to provide the version for it to work
- Windows PowerShell doesn't like assembly-qualified names without the Culture and PublicKeyToken, so I had to add that too...........

But it works and we can take it out when PSCore 6.0 support is ended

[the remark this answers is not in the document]
SeeminglyScience: On the bright side it's still a whole lot better than having to manually replace the DLL!

PR #747, opened 2018-09-17
Fix NRE when trying to start the stopwatch

rjmholt:
Discovered while debugging that the stopwatch in the execution timer wasn't being used properly.


rjmholt:
Oops on non-static field.

@tylerl0706 for some reason .NET people like to call that the [Color problem](https://blogs.msdn.microsoft.com/ericlippert/2009/07/06/color-color/)

PR #752, opened 2018-09-27
Enforce Get-Help offline dropback when page or internet is down

rjmholt:
I noticed it was hard to get the offline behaviour in @corbob's awesome addition to fire, so I made a minor edit where we test if the help page is accessible first.

@corbob GitHub won't let me add you as a reviewer, but if you've got time to review please do!

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -261,8 +261,9 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                 try {
                     $null = Microsoft.PowerShell.Core\Get-Help $CommandName -Online
                 } catch [System.Management.Automation.PSInvalidOperationException] {
-                    Microsoft.PowerShell.Core\Get-Help $CommandName -Full
-                }";
+                    return Microsoft.PowerShell.Core\Get-Help $CommandName -Full | Out-String
+                }
+                ";
 
             if (string.IsNullOrEmpty(helpParams)) { helpParams = "Get-Help"; }
 
@@ -270,6 +271,8 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                 .AddScript(CheckHelpScript, useLocalScope: true)
                 .AddArgument(helpParams);
 
+            // TODO: Rather than print the help in the console, we should send the string back
+            //       to VSCode to display in a help pop-up (or similar)
             await editorSession.PowerShellContext.ExecuteCommand<PSObject>(checkHelpPSCommand, sendOutputToHost: true);
             await requestContext.SendResult(null);
         }
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -264,8 +264,8 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                     $oldSslVersion = [System.Net.ServicePointManager]::SecurityProtocol
                     [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12
 
-                    # HEAD means we don't need the content itself back :)
-                    $status = (Invoke-WebRequest -Method Head -Uri $helpUri -ErrorAction Stop).StatusCode
+                    # HEAD means we don't need the content itself back, just the response header
+                    $status = (Invoke-WebRequest -Method Head -Uri $helpUri -TimeoutSec 5 -ErrorAction Stop).StatusCode
                     if ($status -lt 400) {
                         $null = Microsoft.PowerShell.Core\Get-Help $CommandName -Online
                         return
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -254,15 +254,23 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                     [String]$CommandName
                 )
                 try {
-                    $null = Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop
+                    $command = Microsoft.PowerShell.Core\Get-Command $CommandName -ErrorAction Stop
                 } catch [System.Management.Automation.CommandNotFoundException] {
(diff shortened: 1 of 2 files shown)


corbob:
Honestly I designed my addition to only check if it could launch the browser. This will definitely make that better for the scenario where you don't have internet, or the site is down.

I'll try to review this evening or tomorrow morning.

rjmholt:
My only concern about my change is that it will be slower -- so will let everyone else decide :)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -254,22 +254,33 @@ protected Task Stop()
...
                     $PSCmdlet.ThrowTerminatingError($PSItem)
                 }
                 try {
-                    $null = Microsoft.PowerShell.Core\Get-Help $CommandName -Online
-                } catch [System.Management.Automation.PSInvalidOperationException] {
-                    Microsoft.PowerShell.Core\Get-Help $CommandName -Full
-                }";
+                    $helpUri = [Microsoft.PowerShell.Commands.GetHelpCodeMethods]::GetHelpUri($command)
+                    # HEAD means we don't need the content itself back :)
+                    $status = (Invoke-WebRequest -Method Head -Uri $helpUri -ErrorAction Stop).StatusCode
```
rkeithhill: Yup, I've used HEAD before for this sort of check.

corbob: What will happen when Windows PowerShell can't get the page because Tls12 isn't enabled? Should we be adding `[System.Net.ServicePointManager]::SecurityProtocol += "Tls12"` to be safe?

Are there implications of adding Tls12 here? I understand that it could break some older sites which might be a concern for some users.

rkeithhill:
Oh, according to the docs, iwr is available in v3 - https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-6

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -254,22 +254,33 @@ protected Task Stop()
...
+                return Microsoft.PowerShell.Core\Get-Help $CommandName -Full
+                ";
 
             if (string.IsNullOrEmpty(helpParams)) { helpParams = "Get-Help"; }
 
             PSCommand checkHelpPSCommand = new PSCommand()
                 .AddScript(CheckHelpScript, useLocalScope: true)
                 .AddArgument(helpParams);
 
+            // TODO: Rather than print the help in the console, we should send the string back
```
corbob: This actually fits nicely into the changes that https://github.com/powershell/vscode-powershell/issues/884 morphed into.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rjmholt: We can wrap it in a `try`/`finally` and set it back after the request is made. Hopefully that's not too expensive to accomplish.

rjmholt: Since we're only requesting a site that MS owns, that should be guaranteed to work :)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -254,22 +254,39 @@ protected Task Stop()
...
-                } catch [System.Management.Automation.PSInvalidOperationException] {
-                    Microsoft.PowerShell.Core\Get-Help $CommandName -Full
-                }";
+                    $helpUri = [Microsoft.PowerShell.Commands.GetHelpCodeMethods]::GetHelpUri($command)
+
+                    $oldSslVersion = [System.Net.ServicePointManager]::SecurityProtocol
+                    [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12
+
+                    # HEAD means we don't need the content itself back :)
+                    $status = (Invoke-WebRequest -Method Head -Uri $helpUri -ErrorAction Stop).StatusCode
```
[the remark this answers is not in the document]
rjmholt: I would assume in an offline state we'd fail pretty quickly (testing it by unplugging my ethernet cable just then, offline is pretty quick).

But yeah, maybe we should add an explicit timeout. The lower it is the less of a performance hit we take when internet is down in a weird way, but the more likely we will accidentally get offline help for people with slow-but-working internet access.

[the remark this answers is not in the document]
rjmholt: 5s?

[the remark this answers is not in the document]
corbob: 5 seconds seems reasonable. I've seen latency as high as 3 - 4 seconds on high speed, but at that point you're better off thinking that you're offline.

rjmholt:
Wow while testing this I've discovered that you literally cannot start VSCode 1.27.2 without internet connection...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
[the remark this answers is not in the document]
rjmholt: Yeah. Just played with the timeout. Doesn't affect PS Core much either way, but Windows PowerShell seems to much prefer having a 5s timeout

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -254,22 +254,39 @@ protected Task Stop()
...
-                } catch [System.Management.Automation.PSInvalidOperationException] {
-                    Microsoft.PowerShell.Core\Get-Help $CommandName -Full
-                }";
+                    $helpUri = [Microsoft.PowerShell.Commands.GetHelpCodeMethods]::GetHelpUri($command)
+
+                    $oldSslVersion = [System.Net.ServicePointManager]::SecurityProtocol
+                    [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12
+
+                    # HEAD means we don't need the content itself back, just the response header
+                    $status = (Invoke-WebRequest -Method Head -Uri $helpUri -TimeoutSec 5 -ErrorAction Stop).StatusCode
```
SeeminglyScience: `Invoke-WebRequest` -> `Microsoft.PowerShell.Utility\Invoke-WebRequest`

PR #753, opened 2018-09-28
[Ignore] Prepare 1.9.0 release

rjmholt:
😃

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # PowerShell Editor Services Release History
 
+## v1.9.0
+### Thursday, September 27, 2018
+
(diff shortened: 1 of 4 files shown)


rkeithhill:
Not a biggie but what prompted the bump to `1.9.0`?

rjmholt:
The help feature and the API deprecation in it. But we could do 1.8.5 instead, I don't mind either. Ideally we could merge my other change in first (which shouldn't be in the changelog -- it's @corbob's feature).

rkeithhill:
Since the extension isn't really a re-usable "library" I don't feel that strongly about semver rules.  And it sounds like the help change could be viewed as a new feature anyway.  So all good.

PR #764, opened 2018-10-09
[Ignore] Use PowerShell download script for Travis builds

rjmholt:
Updates our travis configuration to use the latest PowerShell and not 6.0.0-beta<something>, which was breaking in Ubuntu Trusty.


rjmholt:
@rkeithhill Tbh I don't all that much about it either. I think the current implementation it all uses was written by @DarwinJS anyway. But, it's heavily relied on in the PowerShell repos, so it makes sense to deduplicate and trust it (we already do in vscode-PowerShell I'm pretty sure).

PR #782, opened 2018-10-23
Fix #779: NRE on Dispose in ExecutionTimer

### LINKED ISSUE #779: NRE on Dispose in ExecutionTimer
I'm breakpointing through the PSES to figure out how the Find References mechanism works for another issue and I'm getting NRE in `ExecutionTimer` quite frequently.

The NRE happens on the first line of the `Dispose()` method.

```
t_stopwatch.Stop();
```

It's happening here and in other similar blocks of code
```
using (Logger.LogExecutionTime($"Script analysis of {scriptFile.FilePath} completed."))
{
    semanticMarkers = await editorSession.AnalysisService.GetSemanticMarkersAsync(scriptFile);
}
```

Basically the `ThreadStatic` `t_stopwatch` field is assigned inside the `Logger.LogExecutionTime` call, but when we come back from the 'await' the code is/can be executed on another thread and `Logger.LogExecutionTime` is either `null`, the right `t_stopwatch` or one from another thread.

So before I submit a PR making `t_stopwatch` an instance field I wanted to ask why is it `ThreadStatic`? Am I going to break everything in a way I'm not seeing right now?

Thanks

(I'm running the application on Arch Linux, building with VS Code)

dee-see:
Fix #779: NRE on Dispose in ExecutionTimer

As discussed, `ThreadStatic` implementation replaced by an object pool. My object pool is very basic as I didn't feel like the current usage warranted anything more complex. Let me know if you think I should do anything different!

### DIFF
--- a/src/PowerShellEditorServices/Utility/ObjectPool.cs
+++ b/src/PowerShellEditorServices/Utility/ObjectPool.cs
@@ -7,13 +7,23 @@ using System.Collections.Concurrent;
 
 namespace Microsoft.PowerShell.EditorServices.Utility
 {
-    public class ObjectPool<T>
+    /// <summary>
+    /// A basic implementation of the object pool pattern.
+    /// </summary>
+    internal class ObjectPool<T>
         where T : new()
     {
         private ConcurrentBag<T> _pool = new ConcurrentBag<T>();
 
+        /// <summary>
+        /// Get an instance of an object, either new or from the pool depending on availability.
+        /// </summary>
         public T Rent() => _pool.TryTake(out var obj) ? obj : new T();
 
+        /// <summary>
+        /// Return an object to the pool.
+        /// </summary>
+        /// <param name="obj">The object to return to the pool.</param>
         public void Return(T obj) => _pool.Add(obj);
     }
 }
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices/Utility/ObjectPool.cs
```diff
@@ -0,0 +1,19 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Concurrent;
+
+namespace Microsoft.PowerShell.EditorServices.Utility
+{
+    public class ObjectPool<T>
```
rjmholt: The build is currently failing because XML comments are missing for these types.

`ObjectPool` should be `internal` I imagine, but it would also be worthwhile to comment them to explain their basic semantics if you've got time.

rjmholt: Also, the fact that the XML comments fails a Release build but not a Debug build is something we need to fix :)

rjmholt: Yeah, I hit similar problems as well when I was building on Linux.

I just never think to run a release build and then CI fails for something I wish I'd at least been warned about locally...

Anyway, the work to move off net451 has already been done downstream in the v2 branch, so soon enough it will be easy for Linux and macOS users to have exactly the same build experience as Windows.

PR #785, opened 2018-10-24
Adds ability to use separate pipes for reading and writing

ant-druha:
Some clients have blocking File I/O API, thus it is not possible for them to use single named pipe for reading and writing asynchronously. 
E.g. this is [needed](https://github.com/ant-druha/PowerShell/issues/6) for my PowerShell plugin for IntelliJ IDEs. 
For communication protocol I'm using [lsp4j](https://github.com/eclipse/lsp4j) library which has no issues with tcp protocol but it [does not support named pipes](https://github.com/eclipse/lsp4j/issues/250). So the proposed solution was to use two separate pipes. 

This request adds the ability to use separate pipes for reading and writing when the client specifies the "-SplitInOutPipes" switch for the startup script.

Tested that two pipes approach with my plugin (on OS X and Win7) works fine. Also tested with VS Code on OS X and Win7 also seems to work fine.

### DIFF
--- a/module/PowerShellEditorServices/PowerShellEditorServices.psm1
+++ b/module/PowerShellEditorServices/PowerShellEditorServices.psm1
@@ -31,20 +31,36 @@ function Start-EditorServicesHost {
         [string]
         $HostVersion,
 
+        [Parameter(ParameterSetName="Stdio",Mandatory=$true)]
         [switch]
         $Stdio,
 
+        [Parameter(ParameterSetName="NamedPipe",Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]
         [string]
         $LanguageServiceNamedPipe,
 
-        [string]
-        $LanguageServiceWriteNamedPipe,
-
+        [Parameter(ParameterSetName="NamedPipe")]
         [string]
         $DebugServiceNamedPipe,
 
+        [Parameter(ParameterSetName="NamedPipeHalfDuplex",Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]
         [string]
-        $DebugServiceWriteNamedPipe,
+        $LanguageServiceInNamedPipe,
+
+        [Parameter(ParameterSetName="NamedPipeHalfDuplex",Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $LanguageServiceOutNamedPipe,
+
+        [Parameter(ParameterSetName="NamedPipeHalfDuplex")]
+        [string]
+        $DebugServiceInNamedPipe,
+
+        [Parameter(ParameterSetName="NamedPipeHalfDuplex")]
+        [string]
+        $DebugServiceOutNamedPipe,
 
         [ValidateNotNullOrEmpty()]
         [string]
@@ -105,30 +121,27 @@ function Start-EditorServicesHost {
     $debugServiceConfig =
         Microsoft.PowerShell.Utility\New-Object Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportConfig
 
-    if ($Stdio.IsPresent) {
-        $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::Stdio
-        $debugServiceConfig.TransportType    = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::Stdio
-    }
-
-    if ($LanguageServiceNamedPipe) {
-        $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
-        if ($LanguageServiceWriteNamedPipe) {
-            $languageServiceConfig.Endpoint = "$LanguageServiceNamedPipe$([System.IO.Path]::DirectorySeparatorChar)$LanguageServiceWriteNamedPipe"
+    switch ($PSCmdlet.ParameterSetName) {
+        "Stdio" {
+            $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::Stdio
+            $debugServiceConfig.TransportType    = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::Stdio
         }
-        else
-        {
+        "NamedPipe" {
+            $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
             $languageServiceConfig.Endpoint = "$LanguageServiceNamedPipe"
+            if ($DebugServiceNamedPipe) {
+                $debugServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
+                $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe"
+            }
         }
-    }
-
-    if ($DebugServiceNamedPipe) {
-        $debugServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
-        if ($DebugServiceWriteNamedPipe) {
-            $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe$([System.IO.Path]::DirectorySeparatorChar)$DebugServiceWriteNamedPipe"
-        }
-        else
-        {
-            $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe"
+        "NamedPipeHalfDuplex" {
+            $sep = [System.IO.Path]::DirectorySeparatorChar
+            $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
+            $languageServiceConfig.Endpoint = "$LanguageServiceInNamedPipe$sep$LanguageServiceOutNamedPipe"
+            if ($DebugServiceInNamedPipe -and $DebugServiceOutNamedPipe) {
+                $debugServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
+                $debugServiceConfig.Endpoint = "$DebugServiceInNamedPipe$sep$DebugServiceOutNamedPipe"
+            }
         }
     }
 
--- a/module/PowerShellEditorServices/PowerShellEditorServices.psm1
+++ b/module/PowerShellEditorServices/PowerShellEditorServices.psm1
@@ -113,7 +113,7 @@ function Start-EditorServicesHost {
     if ($LanguageServiceNamedPipe) {
         $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
         if ($LanguageServiceWriteNamedPipe) {
-            $languageServiceConfig.Endpoint = "$LanguageServiceNamedPipe;$LanguageServiceWriteNamedPipe"
+            $languageServiceConfig.Endpoint = "$LanguageServiceNamedPipe$([System.IO.Path]::DirectorySeparatorChar)$LanguageServiceWriteNamedPipe"
         }
         else
         {
@@ -124,7 +124,7 @@ function Start-EditorServicesHost {
     if ($DebugServiceNamedPipe) {
         $debugServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
         if ($DebugServiceWriteNamedPipe) {
-            $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe;$DebugServiceWriteNamedPipe"
+            $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe$([System.IO.Path]::DirectorySeparatorChar)$DebugServiceWriteNamedPipe"
         }
         else
         {
--- a/module/PowerShellEditorServices/PowerShellEditorServices.psm1
+++ b/module/PowerShellEditorServices/PowerShellEditorServices.psm1
@@ -128,19 +128,20 @@ function Start-EditorServicesHost {
         }
         "NamedPipe" {
             $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
-            $languageServiceConfig.Endpoint = "$LanguageServiceNamedPipe"
+            $languageServiceConfig.InOutPipeName = "$LanguageServiceNamedPipe"
             if ($DebugServiceNamedPipe) {
                 $debugServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
-                $debugServiceConfig.Endpoint = "$DebugServiceNamedPipe"
+                $debugServiceConfig.InOutPipeName = "$DebugServiceNamedPipe"
             }
         }
         "NamedPipeHalfDuplex" {
-            $sep = [System.IO.Path]::DirectorySeparatorChar
             $languageServiceConfig.TransportType = [Microsoft.PowerShell.EditorServices.Host.EditorServiceTransportType]::NamedPipe
-            $languageServiceConfig.Endpoint = "$LanguageServiceInNamedPipe$sep$LanguageServiceOutNamedPipe"
+            $languageServiceConfig.InPipeName = $LanguageServiceInNamedPipe
+            $languageServiceConfig.OutPipeName = $LanguageServiceOutNamedPipe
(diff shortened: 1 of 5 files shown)


module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -173,8 +176,7 @@ function New-NamedPipeName {
 
     # We try 10 times to find a valid pipe name
     for ($i = 0; $i -lt 10; $i++) {
-        # add a guid to make the pipe unique
-        $PipeName = "PSES_$([guid]::NewGuid())"
+        $PipeName = "PSES_$([System.IO.Path]::GetRandomFileName())"
```
ant-druha: It makes pipe name length shorter.

Clarifications:
At system-level the pathname of a Unix domain socket is limited in length and some libraries also have such a restriction, see *Maximum path length* section on [junixsocket](https://code.google.com/archive/p/junixsocket/wikis/GettingStarted.wiki) that I use.

Because as it is now, the pipe name is really long esp. on Linux where it is the Temp folder (usually such as "/var/folders/92/2yqlfyrj3lqcvrhtdwxj8djc0000gp/T/") + long enough guid as a pipe name. The *junixsocket* library will throw an exception with the use of `[guid]::NewGuid`.

Btw Unix sockets (which named pipes are represented by on Linux) recommended path location is /var/run/ (see */var/run : Run-time variable data* section here: http://www.pathname.com/fhs/pub/fhs-2.3.html but std library creates pipe in a temp directory. So it might make sense to reconsider the directory on Linux.

module/PowerShellEditorServices/PowerShellEditorServices.psm1
```diff
@@ -37,9 +37,15 @@ function Start-EditorServicesHost {
...
         $LanguageServiceNamedPipe,
 
+        [string]
+        $LanguageServiceWriteNamedPipe,
+
         [string]
         $DebugServiceNamedPipe,
 
+        [string]
+        $DebugServiceWriteNamedPipe,
```
tylerl0706: I think this might be fine for this PR, but I just wanted to call out that this param block (and the one in `Start-EditorServices.ps1`) should be using ParameterSets to better organize all these parameters.

module/PowerShellEditorServices/Start-EditorServices.ps1
tylerl0706: > So it might make sense to reconsider the directory on Linux.

I'm with you, but IIRC I don't think we have this constructor available to us in .NET Core 2.0 (which is what PowerShell Core 6.0 uses) that lets us pass in a path name of where we want the pipe (or unix socket) to be saved 😢

tylerl0706: p.s. love the use of `GetRandomFileName`!

src/PowerShellEditorServices.Host/EditorServicesHost.cs
```diff
@@ -463,7 +463,19 @@ private IServerListener CreateServiceListener(MessageProtocolType protocol, Edit
 
                 case EditorServiceTransportType.NamedPipe:
                 {
-                    return new NamedPipeServerListener(protocol, config.Endpoint, this.logger);
+                    string endpoint = config.Endpoint;
+                    int splitIndex = endpoint.IndexOf(';');
```
tylerl0706: I'm more inclined to say that the character that separates the two file names should be the `DirectorySeparator` (`System.IO.Path.DirectorySeparatorChar`) since `;` could be a part of a valid file name. I don't think `GetRandomFileName` would ever spit out a `;` but at least if we use the `DirectorySeparatorChar` we can be sure.

tylerl0706: I spoke to @rjmholt about the use of `Endpoint` here. Historically, this was used to support TCP and NamedPipes but since we don't have TCP anymore, we could change this `EditorServiceTransportConfig` to have a `Pipe` and `WritePipe` property (but named better) that way you don't have to do this string splitting.

I think I'm ok if you don't want to do this in this PR - but it's pretty important so if you want to take it on I'd really appreciate it.

module/PowerShellEditorServices/PowerShellEditorServices.psm1
rkeithhill: Yup, `-Stdio` should be in one parameter set, `-LanguageServiceNamedPipe/DebugServiceNamedPipe` in another and `-LanguageServiceWriteNamedPipe/DebugServiceWriteNamedPipe` in yet another where it seems like there should also be -LanguageServiceReadNamedPipe/DebugServiceReadNamedPipe`.  That would make the parameters a bit clearer IMO.

module/PowerShellEditorServices/PowerShellEditorServices.psm1
```diff
@@ -106,12 +112,24 @@ function Start-EditorServicesHost {
 
     if ($LanguageServiceNamedPipe) {
```
rkeithhill: If we used parameter sets then here we'd check `if ($PSCmdlet.ParameterSetName -eq 'NamedPipeDuplex')` and perhaps $LanguageServicePipe would be a mandatory parameter? Not sure about $DebugLanguageServicePipe, I suppose you could create one but not the other?

rkeithhill: If we used parameter sets then here we'd check if ($PSCmdlet.ParameterSetName -eq 'NamedPipeDuplex') and perhaps $LanguageServicePipe would be a mandatory parameter? Not sure about $DebugLanguageServicePipe, I suppose you could create one but not the other?

module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -269,6 +271,9 @@ try {
     $languageServiceTransport = $null
     $debugServiceTransport = $null
 
+    $LanguageServiceWritePipeName = $null
```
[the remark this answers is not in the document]
rkeithhill: Not a script parameter. This variable is a local so it should be camelCased - ` $languageServiceWritePipeName`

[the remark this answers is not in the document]
rkeithhill: Ditto for `$DebugServiceWritePipeName` below.

module/PowerShellEditorServices/PowerShellEditorServices.psm1
ant-druha: Nice suggestions, about parameter sets, indeed, I'll correct this.

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/NamedPipeServerListener.cs
```diff
@@ -97,39 +126,50 @@ public override void Stop()
...
+
+                this.logger.Write(LogLevel.Verbose, $"Named write pipe server {writePipeServer} has been disposed.");
+            }
         }
 
         private void ListenForConnection()
         {
-            Task.Factory.StartNew(
-                async () =>
+            List<Task> connectionTasks = new List<Task> {WaitForConnectionAsync(this.pipeServer)};
```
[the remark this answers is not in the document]
rjmholt: Ah, yeah the coding style in this project isn't great at the moment -- we've been trying to move it over gradually to being less redundant and verbose.

ant-druha:
So, I [have organised](https://github.com/PowerShell/PowerShellEditorServices/pull/785/commits/c47e80b1ff9a66a1d0ac6cc6543e560a6b151aa9) startup script parameters into parameter sets (had to move code around a bit for this, sorry). 
Probably it is better to introduce an enumeration parameter for transport type, with names like e.g. "Stdio", "NamedPipe", "NamedPipeHalfDuplex" and optional parameters for all the pipes, getting rid of StdIo and SplitPipes switches. Then it would be simpler to work with it in script.

Also [introduced](https://github.com/PowerShell/PowerShellEditorServices/pull/785/commits/4e747a136fb655bab0ca511697489264c49d5864) separate properties for pipes in  `EditorServiceTransportType`.

I hope it is suffice for this PR to make it more or less better.

Looking forward to a new release :)

module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -15,7 +15,7 @@
 #       Services GitHub repository:
 #
 #       https://github.com/PowerShell/PowerShellEditorServices/blob/master/module/PowerShellEditorServices/Start-EditorServices.ps1
-
+[Cmdletbinding(DefaultParameterSetName="NamedPipe")]
```
SeeminglyScience:
```suggestion
[CmdletBinding(DefaultParameterSetName="NamedPipe")]
```

src/PowerShellEditorServices.Host/EditorServicesHost.cs
```diff
@@ -43,7 +43,10 @@ public class EditorServiceTransportConfig
         /// For Stdio it's ignored.
         /// For NamedPipe it's the pipe name.
         /// </summary>
-        public string Endpoint { get; set; }
+        public string InOutPipeName { get; set; }
+        public string OutPipeName { get; set; }
+        public string InPipeName { get; set; }
+        internal string Endpoint => OutPipeName != null && InPipeName!=null ? $"In pipe: {InPipeName} Out pipe: {OutPipeName}" : $" InOut pipe: {InOutPipeName}";
```
SeeminglyScience:
```suggestion
        internal string Endpoint => OutPipeName != null && InPipeName != null ? $"In pipe: {InPipeName} Out pipe: {OutPipeName}" : $" InOut pipe: {InOutPipeName}";
```

src/PowerShellEditorServices.Host/EditorServicesHost.cs
```diff
@@ -463,7 +466,15 @@ private IServerListener CreateServiceListener(MessageProtocolType protocol, Edit
 
                 case EditorServiceTransportType.NamedPipe:
                 {
-                    return new NamedPipeServerListener(protocol, config.Endpoint, this.logger);
+                    if (config.OutPipeName !=null && config.InPipeName !=null)
+                    {
+                        this.logger.Write(LogLevel.Verbose, $"Creating NamedPipeServerListener for ${protocol} protocol with two pipes: In: '" + config.InPipeName + "'. Out: '" + config.OutPipeName + "'");
```
SeeminglyScience:
```suggestion
                        this.logger.Write(LogLevel.Verbose, $"Creating NamedPipeServerListener for ${protocol} protocol with two pipes: In: '{config.InPipeName}'. Out: '{config.OutPipeName}'");
```

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/NamedPipeServerChannel.cs
```diff
@@ -11,35 +11,47 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol.Channel
...
+        private NamedPipeServerStream outPipeServer;
 
         public NamedPipeServerChannel(
-            NamedPipeServerStream pipeServer,
+            NamedPipeServerStream inOutPipeServer,
             ILogger logger)
         {
-            this.pipeServer = pipeServer;
+            this.inOutPipeServer = inOutPipeServer;
+            this.outPipeServer = null;
```
SeeminglyScience: nit: This line can be removed, the field will already be initialized with `null` by default.

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/NamedPipeServerListener.cs
```diff
@@ -18,17 +19,32 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.MessageProtocol.Channel
...
         public NamedPipeServerListener(
             MessageProtocolType messageProtocolType,
-            string pipeName,
+            string inOutPipeName,
+            ILogger logger)
+            : base(messageProtocolType)
+        {
+            this.logger = logger;
+            this.inOutPipeName = inOutPipeName;
+            this.outPipeName = null;
```
SeeminglyScience: nit: re: field initialization not needed

src/PowerShellEditorServices.Host/EditorServicesHost.cs
rjmholt: Is the `$` supposed to be there before `{protocol}`?

TylerLeonhardt:
@ant-druha do you think you'll be able to address this feedback? I just want to make sure this PR doesn't get lost because it's super important!

ant-druha:
> @ant-druha do you think you'll be able to address this feedback? I just want to make sure this PR doesn't get lost because it's super important!

Hi @TylerLeonhardt, sorry for the silence! Haven't had time to look at it, yet. Yes I will address the feedback this week or weekend, I hope it is okay.

src/PowerShellEditorServices.Host/EditorServicesHost.cs
ant-druha: There could be $DebugAdapter or $LanguageServer protocols. Example log:
```
Creating NamedPipeServerListener for $LanguageServer protocol with two pipes: In: 'PSES_saem21o3.ere'. Out: 'PSES_dmxrnado.rh5'
...
Creating NamedPipeServerListener for $DebugAdapter protocol with two pipes: In: 'PSES_jzzah5uk.4ek'. Out: 'PSES_w3ablqxx.uyb'
```

module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -65,14 +65,39 @@ param(
...
 
+    [Parameter(ParameterSetName="NamedPipe")]
+    [string]
+    $DebugServicePipeName = $null,
+
+    [Parameter(ParameterSetName="NamedPipeHalfDuplex",Mandatory=$true)]
+    [switch]
+    $SplitInOutPipes,
+
+    [Parameter(ParameterSetName="NamedPipeHalfDuplexParam",Mandatory=$true)]
```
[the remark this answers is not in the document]
ant-druha: > What's the reason for the `"Param"` parameter set? Is it so that you can either set a switch to split the pipe or nominate a pipename yourself?
> 
> I feel like we probably don't need so many different parameter sets -- what is the envisioned usage of each current set?

Right. I wanted to enforce the contract that one could either specify `SplitInOutPipes` Switch or be enforced to specify read/write pipe names directory. Ok, simplified the parameter sets to just one additional set for simplex named pipes.

module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -246,6 +268,37 @@ function Set-NamedPipeMode {
...
+function Test-NamedPipeName-OrCreate-IfNull {
+    param(
+        [string]
+        $PipeName
+    )
+    if (-not $PipeName) {
+        $PipeName = New-NamedPipeName
+    }
+    else {
+        if (-not (Test-NamedPipeName -PipeName $PipeName)) {
```
TylerLeonhardt: So what happens when Test-NamedPipeName equals `true`? Will it return `null`? Should that ever happen?

I feel as if the Test & Create part should be two separate functions.

ant-druha: The function will fail and abort with the error if the supplied by the user pipe name is not a valid name. If a valid name was supplied it just returns this pipe name. If no pipe name was provided it creates it.

src/PowerShellEditorServices.Protocol/MessageProtocol/Channel/NamedPipeServerListener.cs
```diff
@@ -89,47 +117,58 @@ public override void Start()
...
+                    await Task.WhenAll(connectionTasks);
+                    this.OnClientConnect(new NamedPipeServerChannel(this.inOutPipeServer, this.outPipeServer, this.logger));
+                }
+                catch (Exception e)
+                {
+                    this.logger.WriteException(
+                        "An unhandled exception occurred while listening for a named pipe client connection",
+                        e);
+
+                    throw e;
```
rkeithhill: Rethrow should never specify the exception - you'll lose stack info. Use `throw;` by itself.  Ideally our logger methods would return false so you could use C# exception filters e.g.:
```
catch (Exception e) when (this.logger.WriteException("...", e)) 
{
}
```
But we don't do that - yet.  @rjmholt Any chance we could modify the log methods to return `false` for this reason?

rjmholt: Yeah, we could definitely modify the methods.

Or else write wrapper methods that explain the intended caller context:
```csharp
catch (Exception e) when (_logger.WriteExceptionWithoutUnwinding("...", e))
{
}
```

Bit of a mouthful -- just some name that says "I return false for use with exception filters".

ant-druha:
Hi all. Can this PR be merged? Are there any blockers?

TylerLeonhardt:
Once @rkeithhill is good with this, we can merge it in

TylerLeonhardt:
@ant-druha all merged! Let me know when your extension is updated with this. I'd love to spread the word!

PR #789, opened 2018-11-04
Add support for a "Show Documentation" quick fix menu entry

rkeithhill:
This requires a corresponding PR to VSCode to handle the new command:
PowerShell.ShowCodeActionDocumentation

Changed the use of the Diagnostic.Code field to hold the PSSA rule name.
According to the LSP docs, this field is meant to be displayed and
tslint uses it to display the vioated rule. Update the logic to use a
combination of diagnostic fields to get a unique id in order to
retrieve the stored "correction edits" when the
textDocument/codeAction message is processed with the diagnostic
array returned earlier.

Also, observed a few times during startup that RunScriptDiagnostics was
getting called with an empty filesToAnalyze array.  Modified to return
early in that case.

Remove a private overload of DelayThenInvokeDiagnostics that wasn't
being used.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -27,7 +27,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
 {
     public class LanguageServer
     {
-        private static CancellationTokenSource existingRequestCancellation;
+        private static CancellationTokenSource s_existingRequestCancellation;
 
         private static readonly Location[] s_emptyLocationResult = new Location[0];
 
@@ -48,6 +48,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
         private LanguageServerEditorOperations editorOperations;
         private LanguageServerSettings currentSettings = new LanguageServerSettings();
 
+        // The outer key is the file's uri, the inner key is a unique id for the diagnostic
         private Dictionary<string, Dictionary<string, MarkerCorrection>> codeActionsPerFile =
             new Dictionary<string, Dictionary<string, MarkerCorrection>>();
 
@@ -1182,6 +1183,7 @@ function __Expand-Alias {
             return symbolName.IndexOf(query, StringComparison.OrdinalIgnoreCase) >= 0;
         }
 
+        // https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction
         protected async Task HandleCodeActionRequest(
             CodeActionParams codeActionParams,
             RequestContext<CodeActionCommand[]> requestContext)
@@ -1194,8 +1196,17 @@ function __Expand-Alias {
             {
                 foreach (var diagnostic in codeActionParams.Context.Diagnostics)
                 {
-                    if (!string.IsNullOrEmpty(diagnostic.Code) &&
-                        markerIndex.TryGetValue(diagnostic.Code, out correction))
+                    if (string.IsNullOrEmpty(diagnostic.Code))
+                    {
+                        this.Logger.Write(
+                            LogLevel.Warning,
+                            $"textDocument/codeAction skipping diagnostic with empty Code field: {diagnostic.Source} {diagnostic.Message}");
+
+                        continue;
+                    }
+
+                    string diagnosticId = GetUniqueIdFromDiagnostic(diagnostic);
+                    if (markerIndex.TryGetValue(diagnosticId, out correction))
                     {
                         codeActionCommands.Add(
                             new CodeActionCommand
@@ -1203,6 +1214,17 @@ function __Expand-Alias {
                                 Title = correction.Name,
                                 Command = "PowerShell.ApplyCodeActionEdits",
                                 Arguments = JArray.FromObject(correction.Edits)
+                            });
+                    }
+
+                    if (string.Equals(diagnostic.Source, "PSScriptAnalyzer", StringComparison.OrdinalIgnoreCase))
+                    {
+                        codeActionCommands.Add(
+                            new CodeActionCommand
+                            {
+                                Title = $"Show documentation for \"{diagnostic.Code}\"",
+                                Command = "PowerShell.ShowCodeActionDocumentation",
+                                Arguments = JArray.FromObject(new[] { diagnostic.Code })
                             });
                     }
                 }
@@ -1454,15 +1476,15 @@ function __Expand-Alias {
             // If there's an existing task, attempt to cancel it
             try
             {
-                if (existingRequestCancellation != null)
+                if (s_existingRequestCancellation != null)
                 {
                     // Try to cancel the request
-                    existingRequestCancellation.Cancel();
+                    s_existingRequestCancellation.Cancel();
 
                     // If cancellation didn't throw an exception,
                     // clean up the existing token
-                    existingRequestCancellation.Dispose();
-                    existingRequestCancellation = null;
+                    s_existingRequestCancellation.Dispose();
+                    s_existingRequestCancellation = null;
                 }
             }
             catch (Exception e)
@@ -1479,11 +1501,17 @@ function __Expand-Alias {
                 return cancelTask.Task;
             }
 
+            // If filesToAnalzye is empty, nothing to do so return early.
+            if (filesToAnalyze.Length == 0)
+            {
+                return Task.FromResult(true);
+            }
+
             // Create a fresh cancellation token and then start the task.
             // We create this on a different TaskScheduler so that we
             // don't block the main message loop thread.
             // TODO: Is there a better way to do this?
-            existingRequestCancellation = new CancellationTokenSource();
+            s_existingRequestCancellation = new CancellationTokenSource();
             Task.Factory.StartNew(
                 () =>
                     DelayThenInvokeDiagnostics(
@@ -1494,7 +1522,7 @@ function __Expand-Alias {
                         editorSession,
                         eventSender,
                         this.Logger,
-                        existingRequestCancellation.Token),
+                        s_existingRequestCancellation.Token),
                 CancellationToken.None,
                 TaskCreationOptions.None,
                 TaskScheduler.Default);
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1194,8 +1196,17 @@ private bool IsQueryMatch(string query, string symbolName)
             {
                 foreach (var diagnostic in codeActionParams.Context.Diagnostics)
                 {
-                    if (!string.IsNullOrEmpty(diagnostic.Code) &&
-                        markerIndex.TryGetValue(diagnostic.Code, out correction))
+                    if (string.IsNullOrEmpty(diagnostic.Code))
+                    {
+                        this.Logger.Write(
```
rjmholt: 👍

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1639,13 +1652,23 @@ private async Task ClearMarkers(ScriptFile scriptFile, EventContext eventContext
...
+        }
+
         private static Diagnostic GetDiagnosticFromMarker(ScriptFileMarker scriptFileMarker)
         {
             return new Diagnostic
             {
                 Severity = MapDiagnosticSeverity(scriptFileMarker.Level),
                 Message = scriptFileMarker.Message,
-                Code = scriptFileMarker.Source + Guid.NewGuid().ToString(),
+                Code = scriptFileMarker.RuleName,
```
rjmholt: This is so much nicer!

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1639,13 +1652,23 @@ private async Task ClearMarkers(ScriptFile scriptFile, EventContext eventContext
...
         }
 
+        private static string GetUniqueIdFromDiagnostic(Diagnostic diagnostic)
+        {
+            string source = diagnostic.Source ?? "?";
+            string code = diagnostic.Code ?? "?";
+            string severity = diagnostic.Severity != null ? diagnostic.Severity.ToString() : "?";
+            Position start = diagnostic.Range.Start;
+            Position end = diagnostic.Range.End;
+            return $"{source}_{code}_{severity}_{start.Line}:{start.Character}-{end.Line}:{end.Character}";
```
rjmholt: I have a mild trepidation about allocating a string ID for every diagnostic marker, but I think it's definitely worth a shot -- we'll soon find out if the cache is worth it.

A couple of alternatives might be:
1. Use a string builder to create the ID (I think there are documented advantages over a string template, but theoretically it seems like a string template should be just as performant).
2. Combine hashcodes of the elements. Only problem being that .NET 451 and .NET Standard 2.0 don't have a nice `HashCode.Combine()` method like .NET Core 2.1 does.

rkeithhill: Hmm, looking at some internet perf reviews of string interpolation in C# ... OK will switch to StringBuilder.   :-) 

BTW the previous approach allocated a string as well - `Guid.NewGuid().ToString()`.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1182,6 +1183,7 @@ private bool IsQueryMatch(string query, string symbolName)
             return symbolName.IndexOf(query, StringComparison.OrdinalIgnoreCase) >= 0;
         }
 
+        // https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction
```
TylerLeonhardt: omg we should have comments like this all over... so great! (not in this PR :smile:  )

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1534,6 +1540,12 @@ private static FileChange GetFileChangeDetails(Range changeRange, string insertS
             ILogger Logger,
             CancellationToken cancellationToken)
         {
+            // If filesToAnalzye is empty, nothing to do so return early.
+            if (filesToAnalyze.Length == 0)
```
TylerLeonhardt: This if check seems to be in both places. Here and [L1505](https://github.com/PowerShell/PowerShellEditorServices/pull/789/files#diff-54e1b8c127a6e69bb92d6ebdd738e0e5R1505). One can probably be deleted, right?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1479,11 +1501,17 @@ private static FileChange GetFileChangeDetails(Range changeRange, string insertS
                 return cancelTask.Task;
             }
 
+            // If filesToAnalzye is empty, nothing to do so return early.
+            if (filesToAnalyze.Length == 0)
+            {
+                return Task.FromResult(true);
```
TylerLeonhardt: I'm wondering if we want to start switching to:

`Task.CompletedTask`

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.completedtask?redirectedfrom=MSDN&view=netframework-4.7.2

Not sure if we can but wanted to call it out.

rkeithhill: That is a bit clear with regard to "intention".  This property doesn't go back to NET 4.5.1 though.  Do we care about that version of .NET?  It was intro'd in 4.6 according to these docs - https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.completedtask?view=netframework-4.7.2#System_Threading_Tasks_Task_CompletedTask

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: It is perhaps a bit paranoid but I wanted to avoid an unnecessary `Task.Delay` if there were no files to analyze.  Right now, only the method above (which also checks for 0 files) is the only caller but in the future - who knows.  I can take it out if you prefer.  It's probably a bit too paranoid.  :-)

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: gah... you're totally right. We still care about 4.5.1 because of v3/v4.

Once we cross 2.0 we can switch to using `CompletedTask`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: well I think you can keep this one and safely deleted the other. Won't add too much extra work because it never gets to the `Task.Delay` below, right?

I guess it's taking time to create the Task which is a bit of a pain...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1534,6 +1540,12 @@ private static FileChange GetFileChangeDetails(Range changeRange, string insertS
             ILogger Logger,
             CancellationToken cancellationToken)
         {
+            // If filesToAnalzye is empty, nothing to do so return early.
+            if (filesToAnalyze.Length == 0)
+            {
+                return;
```
TylerLeonhardt: Shouldn't this be `Task.FromResult(true);`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: If nothing else calls this function, I think you should keep the other and get rid of this one actually

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: Good question.  I copied line 1558 which just does a return.  I "think" that is OK for a `void` return Task - isn't it?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: Agreed on which one is more important. I'll remove this one.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
SeeminglyScience: Yeah, you only need `Task.FromResult` if the method isn't tagged as `async`.  When it's tagged as `async` the compiler generates the code for managing task objects for you.

In fact if you did add `Task.FromResult(true)` you'd have to await it, and then you'd have to `return` on the next line.

rkeithhill:
Addressed all the current PR change requests.

rjmholt:
Pinging @TylerLeonhardt for approval

rjmholt:
@TylerLeonhardt, nice username btw 😉

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1639,13 +1661,38 @@ private async Task ClearMarkers(ScriptFile scriptFile, EventContext eventContext
...
+        {
+            Position start = diagnostic.Range.Start;
+            Position end = diagnostic.Range.End;
+
+            var sb = new StringBuilder(256);
+            sb.Append(diagnostic.Source ?? "?");
+            sb.Append("_");
+            sb.Append(diagnostic.Code ?? "?");
+            sb.Append("_");
+            sb.Append((diagnostic.Severity != null) ? diagnostic.Severity.ToString() : "?");
```
rjmholt: Alternate possibility here is `diagnostic.Severity?.ToString() ?? "?"`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1216,22 +1218,34 @@ private bool IsQueryMatch(string query, string symbolName)
...
+                            Title = $"Show documentation for \"{diagnostic.Code}\"",
+                            Command = "PowerShell.ShowCodeActionDocumentation",
+                            Arguments = JArray.FromObject(new[] { diagnostic.Code })
+                        });
                 }
             }
 
             await requestContext.SendResult(
-                codeActionCommands.ToArray());
+            codeActionCommands.ToArray());
```
TylerLeonhardt: nit: re-indent or put on one line

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1639,13 +1661,38 @@ private async Task ClearMarkers(ScriptFile scriptFile, EventContext eventContext
...
+            sb.Append("_");
+            sb.Append((diagnostic.Severity != null) ? diagnostic.Severity.ToString() : "?");
+            sb.Append("_");
+            sb.Append(start.Line);
+            sb.Append(":");
+            sb.Append(start.Character);
+            sb.Append("-");
+            sb.Append(end.Line);
+            sb.Append(":");
+            sb.Append(end.Character);
```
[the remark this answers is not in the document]
TylerLeonhardt: small nit... it might be useful to do:

```csharp
StringBuilder sb = new StringBuilder(256)
    .Append(diagnostic.Source ?? "?")
    .Append("_")
    .Append(diagnostic.Code ?? "?")
    .Append("_")
    .Append((diagnostic.Severity != null) ? diagnostic.Severity.ToString() : "?")
    .Append("_")
    .Append(start.Line)
    .Append(":")
    .Append(start.Character)
    .Append("-")
    .Append(end.Line)
    .Append(":")
    .Append(end.Character)

return sb.ToString();
```

so that we can throw a breakpoint on the return line and see what the contents of sb are.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: Yeah, short enough to be one line.  I'll fix that.  Not sure how that got un-indented.  Hmm...

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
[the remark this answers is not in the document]
rkeithhill: Heh, I was just going to suggest this very thing.  :-)  Will fix.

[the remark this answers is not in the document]
rjmholt: Honestly my biggest desire is for a function-return-variable in debug view.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: Doh! Missed this one.  I'll fix it tonight.

PR #790, opened 2018-11-05
Add managed thread id to log output to add debugging threading issues

rkeithhill:
Also update format to display method before filename, makes it easy to
visually scan the log to see what is going on.

Here is log output before this change:
```
2018-11-03 17:26:10.560 [NORMAL] C:\PowerShellEditorServices\src\PowerShellEditorServices.Host\EditorServicesHost.cs: In method 'StartLanguageService', line 194:
```
And after:
```
2018-11-03 17:21:55.858 [NORMAL] [tid:15] In method 'StartLanguageService' C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\src\PowerShellEditorServices.Host\EditorServicesHost.cs:194:
```

### DIFF
--- a/src/PowerShellEditorServices/Utility/PsesLogger.cs
+++ b/src/PowerShellEditorServices/Utility/PsesLogger.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Threading;
 using Serilog.Core;
 
 namespace Microsoft.PowerShell.EditorServices.Utility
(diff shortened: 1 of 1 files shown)


TylerLeonhardt:
@rkeithhill can you double check that vscode still gives you syntax highlighting like it does today?



pretty sure it's just looking for dates and numbers but I just want to double check :smile:

rkeithhill:
I don't see much difference in the old output.  The new thread id number is colorized blue:

image

rkeithhill:
How about I simplify/lessen the output a bit more.  I don't think the `[]` around the thread id is necessary and after look at the logs for a long time, I don't think the word "method" adds much info.  Considering how large these logs can get, it is probably worth being frugal with what we output.  So with that in mind, this is what the above suggested changes look like:

image

For 2.0.0, if Serilog supported `ReadOnlySpan<char>` I'd love to slice into the source paths, changing this:
```
C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices\src\PowerShellEditorServices.Host\EditorServicesHost.cs
```
to
```
src\PowerShellEditorServices.Host\EditorServicesHost.cs
```
I'd compute the slice offset once and stash it in a static variable.  But alas, Serilog doesn't appear to support `ReadOnlySpan<char>`.  We could use Substring that would allocate another string on every call to the logger.  Not a good trade-off IMO.  

If folks agree with these changes, I'll push my commit with them and then I'd like to merge this.

rkeithhill:
OK, going for the presumptive close on the previous post.  :-)  If folks object, I'll rollback that last commit.

Issue #190, opened 2016-03-15
Script arguments to debug adapter should be wrapped in quotes if they contain spaces


daviwil:
See the following two issues for more information.

https://github.com/Microsoft/vscode/issues/4057
https://github.com/PowerShell/vscode-powershell/issues/117

rkeithhill:
We have to be careful/smart about this.  If you wrap `"-Index", "1, 2, 3"` with quotes you get `-Index "1, 2, 3"` and now we've changed the type of the argument from `[int[]]` to `[string]`.  You could say the folks should do this instead `"-Index", "1", "2", "3"` or `"-Index 1, 2, 3"`.  The first will work the second will totally break again because `SomeCommand "-Index 1, 2, 3"` is a lot different than `SomeCommand -Index 1, 2, 3`.  

I dunno on this one.  Seems like it would be a bit dangerous to do this.

Another solution would be to only support a single string and not an array of strings for `args`.  Then it should make it more obvious that individual arguments with spaces need internal quoting e.g. `"-Index 1, 2, 3 -Path 'C:\program files'"`.  The downside is that this would be a breaking change.

daviwil:
Yeah, you're right, maybe we should just be accepting a single string.  There's not really a good reason to accept an array of strings since it complicates everything.

Maybe we can just change our schema in VS Code for now to specify `string` for the `args` field so that the user gets a warning if they're using an array.  In our LaunchRequest we can change the Args property to be `JToken` so that either a string or an array can be passed.  We'd have to do an extra step to determine if Args contains a `JValue` or a `JArray` and then act accordingly.

daviwil:
Forgot to should add, we'd deprecate the array support after a version or two once we have appropriate documentation showing how to do args the right way.  At the time of deprecation we could even give a specific error message from the debug adapter warning the user of this.

rkeithhill:
Another somewhat lame option if we were to keep array of strings is to add more help in the scheme e.g.:

![vscodeargswithspacetip](https://cloud.githubusercontent.com/assets/5177512/14047164/ae9c5e7a-f26b-11e5-9236-7c407ef7af7b.png)

The problem with this is that it isn't very likely users would hover over the `args` field to get this tip.  Meh, probably better to revert back to using just a single string.  

[...]

PR #794, opened 2018-11-18
Fix Pester CodeLens run/debug by not quoting params/already quoted args

rkeithhill:
Requires a corresponding PR in vscode-powershell.

We should look this one over good as it has potential to break folks.

I think the odds are low.  Essentially we added the QuoteEscapString
method because we get raw paths from VSCode when debugging
the current file.  Those paths are quoted because they may contain
spaces and because we use single quotes, existing single quotes
have to be doubled e.g C:\Don'tUseQuotesInPaths to
'C:\Don''tUseQuotesInPaths'.

This unfortunately was always being invoked on args. Now we
check if the arg is a parameter (starts with '-') or is already quoted.  The thinking is if the users is specifying a quoted arg, it's on them
to handle any necessary escaping.

BTW is there another char that can act as the start of a parameter -
like the longer dash - whatever that is called.

Longer term, we need a better solution for how to handle paths
from VSCode.  Maybe we handle them there and eliminate the
escaping logic in PSES?

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -305,12 +305,22 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                 var sb = new StringBuilder();
                 for (int i = 0; i < launchParams.Args.Length; i++)
                 {
-                    sb.Append(PowerShellContext.QuoteEscapeString(launchParams.Args[i]));
-                    if (i < launchParams.Args.Length - 1)
+                    string arg = launchParams.Args[i];
+                    if (arg.StartsWith("-") || arg.StartsWith("'") || arg.StartsWith("\""))
+                    {
+                        sb.Append(arg);
+                    }
+                    else
+                    {
+                        sb.Append(PowerShellContext.QuoteEscapeString(arg));
+                    }
+
+                    if (i < (launchParams.Args.Length - 1))
                     {
                         sb.Append(' ');
                     }
                 }
+
                 arguments = sb.ToString();
                 Logger.Write(LogLevel.Verbose, "Script arguments are: " + arguments);
             }


rkeithhill:
Ran across another case where the path quote escaping it biting us.  I tried to pass these two parameters to debug launch `-PesterOption` and `@{IncludeVSCodeMarker=$true}` and we wound up with this in the script args string `-PesterOption '@{IncludeVSCodeMarker=$true}'`.  Sigh, guess we could special case `@` and `$` and whatever but now I'm really thinking we punt on this whole path escaping thing and check the paths on the VSCode before we send them.  Thoughts?

TylerLeonhardt:
> check the paths on the VSCode before we send them.

Can you expand on this?

rkeithhill:
For some history, check out:

https://github.com/PowerShell/PowerShellEditorServices/issues/190
https://github.com/Microsoft/vscode/issues/4057

Related issues:
https://github.com/PowerShell/vscode-powershell/issues/1526

OK, I think this is the PR that broke the Pester CodeLens provider:
https://github.com/PowerShell/PowerShellEditorServices/pull/765/files

The PR above changed to quote and escape every argument.  I don't we can do that.  If I pass an argument of `@{Recurse = $true}` if that is converted to `'@{Recurse = $true}'` that changes the arg from a hashtable to string literal.

I'm not sure what the right answer is but I think we should avoid messing with the debug script args in PSES.  That said, we have to see if we can avoid regression here:

https://github.com/PowerShell/vscode-powershell/issues/1400
https://github.com/PowerShell/vscode-powershell/issues/1526

rkeithhill:
Hmm, we might need to rethink a few things on the #765 - sorry I didn't catcht these during the review:

image

On line 772 (RHS), we are automatically escaping the `scriptToLaunch`.  Yeah, we can't do that.  We renamed this field in the debug config from `program` to `script` to allow folks to enter arbitrary script - not just a filename. For instance, here is a debug config I use from time-to-time:

```
        {
            "type": "PowerShell",
            "request": "launch",
            "name": "PowerShell Pester Tests",
            "script": "Invoke-Pester -PesterOption @{IncludeVSCodeMarker=$true}",
            "args": [],
            "cwd": "${workspaceFolder}"
        },
```
So not only can you enter a command instead of a script, you can even supply parameters.

rjmholt:
Yes, I see I was overzealous in #765.

rjmholt:
I think to make this work we're going to have to throw away whatever our existing contract was with people -- some people are going to have to be broken so that all things can work in some way.

It sounds like we need to leave quoting/escaping up to people themselves.

Maybe we can have a check method that fires a warning if it thinks an argument should be quoted but isn't.

So:
- No string quoting/escaping on input, users pass complicated arguments at their own risk
- Possibly have a method on the VSCode side to log a warning if it thinks an input might be badly formed
- Document our best practices and what not to do. I get the impression that because we've just left it up to people to work out debug configurations, they've all developed different ways of doing things. We should instead write down what we know works and make sure we can point to it in markdown in the repo.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -305,12 +305,22 @@ protected void Stop()
                 var sb = new StringBuilder();
                 for (int i = 0; i < launchParams.Args.Length; i++)
```
rjmholt: I vote we replace all this logic with `arguments = string.Join(launchParams.Args, " ")`

rkeithhill: Sounds good to me.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -305,12 +305,22 @@ protected void Stop()
...
                 {
-                    sb.Append(PowerShellContext.QuoteEscapeString(launchParams.Args[i]));
-                    if (i < launchParams.Args.Length - 1)
+                    string arg = launchParams.Args[i];
+                    if (arg.StartsWith("-") || arg.StartsWith("'") || arg.StartsWith("\""))
```
rjmholt: As you say @rkeithhill, there are [several characters](https://github.com/PowerShell/PowerShell/blob/dedddb6b59b8b01fadb4312bbddc7992a6cbbe0a/src/System.Management.Automation/engine/parser/tokenizer.cs#L4300-L4303) PowerShell accepts to begin a parameter.

rkeithhill:
> * No string quoting/escaping on input, users pass complicated arguments at their own risk

Agreed and this applys to debug launch config `args` field *and* the pop up you get when you launch `Debug Current File w/Script Args`.

It could be I'm one of the few folks (involved with the original design) what the `script` field can specify arbitrary script.  In fact, the schema description for this field says: `Optional: Absolute path to the PowerShell script to launch under the debugger.`    Now we do ship with one debug configuration with this script settting `Invoke-Pester` but we don't ship any that look like this `"script": "Invoke-Pester -SomParam somArg"`.  With this in mind, maybe we only guarantee to support the case of `"script": <script-path> | <command-name>`.  This would simplify the logic in the `ExecuteScriptWithArgs`.  If you want to specify the args to a `<command-name>` then you have to use the `"args":` field. I like this idea.

We've gotten by for a long time with that code in `ExecuteScriptWithArgs`. The old code (pre #765) failed because it didn't escape all the troublesome characters like `&` and `(`.  PR #765 has a better approach of just single quote the filepath (if it is an existing file) and double up any interior single quotes.  I think this is all we need to make this work pretty.  I'll know more after some testing.

rjmholt:
> If you want to specify the args to a <command-name> then you have to use the "args": field. I like this idea.

100% agree. It makes `args` not redundant and `<command> | <script-path>` seems very consistent with PowerShell's own invocation concepts.

So this means:
- `script`/`program`: Unescaped path or command name - we will escape it (I'm happy if we ask people to escape it too).
    -  (Ideally we could change the name to `command`, and deprecate old ones, but that's a good way to make life harder I guess)
- `args`: User-escaped list of parameters/arguments to the `script`. Ideally we could validate/sanitise this string (probably on the VSCode side) to prevent weird bugs from unpaired quotes or things.

?

Any other things we need to nail down?

rkeithhill:
So this work - funky path in script with args:
```
    Launch script is: . 'c:\Users\Keith\.vscode\extensions\ms-vscode.powershell-1.9.0\examples\Foo & (x86) [bar]\foo.ps1' 50 10 blah
```

rjmholt:
Do you think that will work @rkeithhill? Want to hammer this out with you so we've got all the bases covered

rkeithhill:
Agreed on all the "not touching args" comments.  RE the script field, I think we are applying LSP rules to the VSCode debug protocol that don't apply.  Here is the spec on the launch request:

https://github.com/Microsoft/vscode-debugadapter-node/blob/9bcba8996b37e4916849d061325b8529e55be9d7/debugProtocol.json#L673

Note that command is a `string` with no mention of uri.  Also note this comment in the description:
> Since launching is debugger/runtime specific, the arguments for this request are not part of this specification.

Also  note how the VSCode sample/mock debugger handles this field (`program` is the equivalent of our `script` field):

https://github.com/Microsoft/vscode-mock-debug/blob/042d19a27a8e3a08f27a24110506b53fbecc75ce/src/extension.ts#L50

It simply uses `${file}` and doesn't convert to a uri.

I believe we should keep the `script` processing as-is on the VSCode side.  On the PSES side, all we need to do is single quote the value (escaping embedded single quotes)  if the value is determined to be an existing file.  This is exactly what PS does for you when you tab-complete a filename and that handles all the other special chars.

rjmholt:
Yeah, that was my original inclination as well, to also pass the script path verbatim and rely on users to escape that path properly themselves.

But I can also imagine sending it through as a URI mainly to simplify our API requirements.

I don't feel strongly either way. The first seems more consistent but the second seems less likely to annoy people.

In the first case, I'm wondering:
- Should we do any escaping at all (including put quotes around the `script`)?
- If we do, should we also regex escape?

While I can see arguments for escaping, it's sort of what's gotten us into trouble so far. I feel like it would ideally behave like `LiteralPath`, but (1) the logic we pass it to has no `LiteralPath` functionality and (2) `args` are passed verbatim; users must escape those themselves. So I think we should just pass `script` verbatim as well. Then in the default config scenario, the extension should escape the path it sends down.

rkeithhill:
> rely on users to escape that path properly themselves.

They shouldn't need to do that. The change that you made that had the biggest impact was to switch from doing this:
```
. C:\Program` Files` (x86)\`[foo`]'bar arg1 arg2
```
to this:
```
. 'C:\Program Files (x86)\[foo']''bar' arg1 arg2
```
This fixes a lot of funky path issues by simply single quoting the path instead of trying to escape all the various chars which we weren't doing before your PR.  IIRC we only escaped `[]` and `<space>`.  That is why folks were running into issues with `&` and `(x86)`.  

This covers the `AddScript` case we use when there are args.  When there are no args, we use `AddCommand` and I've verified that you can pass a string with all those funky chars  in the path and PowerShell handles it correctly.

rjmholt:
Ah, so having us do the escaping is helpful because it depends on which `PSCommand` builder method we use as to whether we want to escape or not?

As usual, I'm in favour of taking as much of the magic out as possible, but if the above is the case, then it makes sense to promise to escape the path if necessary.

Regex escaping still troubles me a bit, but I suppose we can cross that bridge if we need to later.

rjmholt:
That all makes sense to me.

The only part I want to really verify is:

> This means we don't do **any** regex escaping in this scenario.

If I have a script like `C:\Users\me\Documents\My Scripts\[script].ps1` or even `/home/me/Documents/My Scripts/scr*pt.ps1`, will that work along all the code paths we use?

When I was experimenting with different pathological script names for #765, I found that there wasn't a great mechanism for dot-sourcing such a script and that single quotes didn't do anything. `PSCommand.AddCommand` I think works with these script names, `PSCommand.AddScript` I can't remember, and straight string templating seemed to be the least safe.

I ended up opening https://github.com/PowerShell/PowerShell/issues/7999 and https://github.com/PowerShell/PowerShell/issues/7995 as a result.

rkeithhill:
According to the Windows FileExplorer a path cannot have a `*` in it:

image

But I'm updating my test path to be more abusive - including `$ @{}` - and the tab-complete path works:
```
11-20 09:13:50 22ms 56> . '.\Foo & (x86) [bar] isn''t @{good} dir& $name\foo.ps1' arg1 arg2
In foo.ps1 with args: arg1, arg2
C:\Users\Keith
```
Just to re-iterate I'm not suggesting this particular change for anywhere else - just in this case where we are dot sourcing the user's script.  That `WildcardEscapePath` method you wrote is working good for the breakpoint paths AFAICT.  :-)

TylerLeonhardt:
I think on Unix a * is valid. Not sure if that changes anything.

rjmholt:
Yeah `*` isn't valid on Windows, but it is on *nix (hence `/home/me/...`). Ok cool, I just wanted to check that. I'm pretty happy then that we've got the right behaviour.

PR #808, opened 2018-12-07
Win7 OSArchitecture crash workaround

rjmholt:
Avoids calling the System.Runtime.InteropServices.RuntimeInformation API that crashes in Win7.

See https://github.com/PowerShell/vscode-powershell/issues/1633.


src/PowerShellEditorServices.Host/EditorServicesHost.cs
```diff
@@ -493,6 +495,16 @@ private IServerListener CreateServiceListener(MessageProtocolType protocol, Edit
             }
         }
 
+        private static string GetOSArchitecture()
+        {
+            if (IntPtr.Size == 8)
```
SeeminglyScience: This will tell you the architecture of the current process, so this could be a x86 process running on a x64 OS.

Does `Environment.Is64BitOperatingSystem` work for us?

rjmholt:
This doesn't work yet. Looking into it

rjmholt:
Ok, I've fixed my foolishness

PR #815, opened 2018-12-12
Fix crash when untitled files are opened as PowerShell

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/1649.

Finding dot-sourced references when we parse script files currently relies on the file having a directory, which is not true for untitled files.

This is a simple fix for the crash that occurs as a result.

If there's interest, I could try and improve the situation by changing the visitor to have a null PSScriptRoot mean ignore PSScriptRoot paths, or even try to make an untitled dir... (1st feels like it has merit, second maybe not).

### DIFF
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -649,7 +649,9 @@ namespace Microsoft.PowerShell.EditorServices
                     .ToArray();
 
             // Untitled files have no directory
-            // TODO: Make untitled files support dot-sourced references
+            // Discussed in https://github.com/PowerShell/PowerShellEditorServices/pull/815.
+            // Rather than working hard to enable things for untitled files like a phantom directory,
+            // users should save the file.
             if (IsUntitledPath(this.FilePath))
             {
                 return;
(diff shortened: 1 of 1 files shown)


rkeithhill:
BTW this scenario via `F8` does not work in ISE.  
```
C:\Users\Keith
12-15 16:40:25 56ms 2> . $PSScriptRoot\test.ps1
. : The term '\test.ps1' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or 
if a path was included, verify that the path is correct and try again.
```
OTOH ISE doesn't crash.  I dunno, I think the right answer might be what @rjmholt has implemented for the untitled scenario - just punt on analyzing dot sourced files using $PSScriptRoot.  Seems like the more we try to support this scenario, the more we are setting folks up for failure later (no breakpoint support, etc).  I view this scenario for folks just wanting to play around a little bit.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -648,7 +648,14 @@ private void ParseFileContents()
                     .Select(ScriptFileMarker.FromParseError)
                     .ToArray();
 
-            //Get all dot sourced referenced files and store  them
+            // Untitled files have no directory
+            // TODO: Make untitled files support dot-sourced references
```
rkeithhill: I wouldn't bother but that is just my 2 cents.

rjmholt:
I'm going to favour simplicity here. We can work out if we really want to support untitled files later I think.

PR #818, opened 2018-12-13
(maint) Add Xunit Trait Filter

glennsarti:
Previously when running tests all tests would need to be run, however when developing this
can take a while.  This commit adds an option to add a trait filter to XUnit
as per the Xunit documentation [1]. Later commits may add the actual trait
attributes to the tests.

[1] https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests

### DIFF
--- a/PowerShellEditorServices.build.ps1
+++ b/PowerShellEditorServices.build.ps1
@@ -11,7 +11,9 @@ param(
 
     [string]$ModulesJsonPath = "$PSScriptRoot/modules.json",
 
-    [string]$DefaultModuleRepository = "PSGallery"
+    [string]$DefaultModuleRepository = "PSGallery",
+
+    [string]$TestFilter = ''
 )
 
 #Requires -Modules @{ModuleName="InvokeBuild";ModuleVersion="3.2.1"}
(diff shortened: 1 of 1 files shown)


PowerShellEditorServices.build.ps1
```diff
@@ -223,24 +225,31 @@ function UploadTestLogs {
     }
 }
 
+function XunitTraitFilter {
+    $TraitFilter = ''
+    # Reference https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests
+    if (![string]::IsNullOrEmpty($TestFilter)) { $TraitFilter = "-trait $TestFilter" }
```
rkeithhill: Since `$TestFilter` is type string, it is sufficient to use the more canonical `if (!$TestFilter) { ... }`

rkeithhill: Actually this whole function could be simplified to just:
```
if ($TestFilter) { "-train $TestFilter" } else { "" }
```

rjmholt:
```powershell
if (!$TestFilter)
{
    return ""
}

"-train $TestFilter"
```

glennsarti: I can change it but I'm not really a fan of that kind of operation. It makes things like `if ("False") { Write-Out "This is True" }` confusing.  I prefer explicit checks

rkeithhill: `if ($TestFilter)` is still an explicit check of "is string null or empty" and sorry but `"False"` is not null or empty.  I mean,  underneath PS is just doing string.IsNullOrEmpty for you.

rkeithhill: Also, in the vein of please follow the coding style that already exists in the script, you will see this style:
```
if (-not $name)
...
if ($env:APPVEYOR) {
```

glennsarti: Changed.

rjmholt: My feeling is that it's a common convention in Python, JS and PowerShell for a reason. It can be abused, but this case is precisely the design intent. Same as `return value || default`.

In PowerShell's case, calling the C# APIs is sometimes less robust; PowerShell's coercions absorb more possible bad cases with less code.

PR #829, opened 2018-12-15
Update various NuGet packages

bergmeister:


bergmeister:
Thanks for upstreaming my branch. I also updated the `System.IO.Pipes.AccessControl` NuGet package from `4.5.0` to `4.5.1`
P.S. (on an unrelated note): Today I managed to build and test the vs-code ps extension together with my other PSES and my PSSA branches to validate my 3 other vscode-powershell/PSES PRs end to end as they all depend on each others, we will have to wait with the merge though until PSSA 1.18 gets released to the public in a few weeks (hopefully) as those will be using new features.

PR #834, opened 2018-12-18
Update version for preview builds

rjmholt:


appveyor.yml
```diff
@@ -1,4 +1,4 @@
-version: '1.10.2{build}'
+version: '1.10.3{build}'
```
rkeithhill: Seems like we should just have to set this version once - here - and then use a small PS script (or inline commands) to modify the `PowerShellEditorServices.Common.props` and `module/PowerShellEditorServices/PowerShellEditorServices.psd1` files before starting the build.

rjmholt: Given that in time we'll move to Azure DevOps, we should probably use the module version as the canonical one.

But that complicates life since this yaml file needs to be checked in with the version number.

PR #838, opened 2018-12-28
Fix NullRefEx bug when accessing scriptFile.ReferencedFiles

rkeithhill:
This happens because the ScriptFile ctor does not initialize all its
public props.  I added initialization for the other public props
except for the ScriptAst prop.  I don't see an empty Ast.  Perhaps null
is OK for this prop?

This address vscode-powershell bug https://github.com/PowerShell/vscode-powershell/issues/1675

No real need for ScriptTokens to have a backing field. Make it auto like the other props.

Also, for the 2.0.0 branch, we should see if we can use Array.Empty<>()
for initialization.  It isn't availble to net45x.  :-(

### DIFF
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -146,17 +146,13 @@ namespace Microsoft.PowerShell.EditorServices
             TextReader textReader,
             Version powerShellVersion)
         {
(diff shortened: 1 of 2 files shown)


rkeithhill:
Working on it.  Man it sure would be nice if the xUnit / VS Test Explorer integration weren't so crappy.  Can't seem to step into tests and test output won't show me line numbers.  Sigh...

PR #856, opened 2019-01-22
Send Pester describe line and info whether Pester 4.6.0 is installed to PowerShell.RunPesterTests command

bergmeister:
This is the 1st PR needed as a pre-requisite of the implementation of https://github.com/PowerShell/vscode-powershell/issues/1710
The extension needs to know whether Pester 4.6.0 is available and the line number of the describe to be able to call new Pester syntax that will enable to execute a describe block by line number, therefore enabling the run of describe blocks with an interpolated string.
The work to determine the version is only done once in a background job and once it has finished, it will be cached. Is there a better way to persist such information to the file system if Pester >4.6.0 was available (which is unlikely to change)
Note that this change does not break the vs-code extension because we only ADD arguments


src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -54,7 +93,7 @@ public PesterCodeLensProvider(EditorSession editorSession)
                     new ClientCommand(
                         "PowerShell.RunPesterTests",
                         "Run tests",
-                        new object[] { scriptFile.ClientFilePath, false /* No debug */, pesterSymbol.TestName })),
+                        new object[] { scriptFile.ClientFilePath, false /* No debug */, pesterSymbol.TestName, pesterSymbol.ScriptRegion.StartLineNumber, _pesterV4_6_0_OrHigherAvailable })),
```
rkeithhill: This could be:
```
new object[] { scriptFile.ClientFilePath, false /* No debug */, pesterSymbol.TestName, (_pesterV4_6_0_OrHigherAvailable ? pesterSymbol.ScriptRegion.StartLineNumber : null)})),
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private void DeterminePesterVersion()
+        {
+            Task.Run(() =>
+            {
+                using (var powerShell = System.Management.Automation.PowerShell.Create())
+                {
+                    powerShell.AddCommand("Get-Module")
+                              .AddParameter("ListAvailable")
```
rkeithhill: The thing I'm not crazy about here is that someone may have 4.6.0 installed but an earlier version loaded.  If we go this route, perhaps we should double check to make sure an earlier version of Pester is not loaded into PSIC.  If so we could import-module with `-MinimumVersion 4.6` to ensure that version is loaded.  Just thinking out loud here so take this with a grain of salt.

bergmeister: Yes, I thinking this as well, but calling `Get-Command Invoke-Pester` is much more expensive and takes more than 2 seconds on my (modern) machine, calling `Get-Module` takes only a quarter of a second the first time. We do the same type of checking for other modules such as PSSA and Plaster.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: Hmm, the C# compiler needed a bit more help with that, my best solution is
```
 _pesterV4_6_0_OrHigherAvailable ? pesterSymbol.ScriptRegion.StartLineNumber : default(int?);
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -84,6 +84,9 @@ private void DeterminePesterVersion()
             PesterSymbolReference pesterSymbol,
             ScriptFile scriptFile)
         {
+            // A value of null is a signal to PSES that the available Pester version does not support
+            // running Describe blocks by name (the test name will used instead then)
+            int? describeBlockLineNumber = _pesterV4_6_0_OrHigherAvailable ? pesterSymbol.ScriptRegion.StartLineNumber : default(int?);
```
rkeithhill: Isn't `default(int?)` just `null`?  I get using `default(T)` in a generic class or method but not sure it is needed here.  Unless I'm missing something - which is ENTIRELY possible.  :-)

bergmeister: Yes, but...
image

bergmeister: Yes, `default(int?)` is null. I changed it to the following, which is probably the most readable version by casting the first `int` to `int?`:
```csharp
int? describeBlockLineNumber = _pesterV4_6_0_OrHigherAvailable ? (int?)pesterSymbol.ScriptRegion.StartLineNumber : null;
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
rjmholt: > What I would definitely not do is execute Remove-Module or Import-Module because at this point the user has lost control over his/her shell

If we're executing Pester, the user must at some point cede control of their shell. That's the fundamental tricky thing about the PSIC -- we have to share with the user. The question is to what extent.

My concern is that the PSIC means a user can just remove/replace Pester at any time, invalidating an earlier one-off query.

One possibility is that we just extend what we do now:
- Do a once off check for Pester 4.6.0 installation (and may need to check what version is loaded, if any)
- If Pester 4.6.0 is not available, we do what we do now
- If the test name has no variables, we do what we do now
- Otherwise, we try executing with the new Pester syntax and if there's a ParameterBindingException on Invoke-Pester, we fall back to what we do now.

As an aside, I feel like that's pretty complicated (and could add significant overhead) for what's not a huge win as a feature. But if Pester 4.6.0 is well adopted, this might get easier. I certainly think the line number invocation is a more natural way to do things.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
             _editorSession = editorSession;
             _symbolProvider = new PesterDocumentSymbolProvider();
+
+            DeterminePesterVersion();
+        }
+
+        /// <summary>
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private void DeterminePesterVersion()
```
rjmholt: I'm worried that this is a task in disguise. As a `void` method around a `Task.Run` invocation, it either blocks the caller or it's a race condition around the `_pesterV4_6_0_OrHigherAvailable` variable.

This strikes me as much better implemented as a `Task<bool>`:

```csharp
private async Task<bool> HasPester460Async()
{
    var pesterModSpec = new Hashtable()
    {
        { "ModuleName", "Pester" },
        { "ModuleVersion", new Version(4, 6) }
    };

    var gmoListPesterCommand = new PSCommand()
        .AddCommand("Microsoft.Powershell.Core\\Get-Module")
            .AddParameter("ListAvailable")
            .AddParameter("FullyQualifiedName", pesterModSpec);

    IEnumerable<PSModuleInfo> modules = await _powerShellContext.ExecuteCommandAsync<PSModuleInfo>(
        gmoListPesterCommand,
        sendOutputToHost: false,
        sendErrorToHost: false);

    return modules != null && modules.Any();
}
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
SeeminglyScience: @rjmholt 

> Otherwise, we try executing with the new Pester syntax and if there's a ParameterBindingException on Invoke-Pester, we fall back to what we do now.

Actually we could just do that, and then warn about required version.  That exception also has a `ParameterName` property we could use to verify that it's really a version problem.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
SeeminglyScience: 👍 to @rjmholt's comment. At the very least it would be nice to store the result as a task to avoid race conditions.

rjmholt: The other thing is that making this an async method makes it much harder to call from the constructor. I think that's a good thing, because I'm not a big fan of complex constructor logic, and tasks in constructors make me uneasy about PSES' already bad thread situation.

Instead, the best solution to me feels like a [lazy async](https://blog.stephencleary.com/2012/08/asynchronous-lazy-initialization.html) field created in the constructor around the method above

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+                {
+                    powerShell.AddCommand("Get-Module")
+                              .AddParameter("ListAvailable")
+                              .AddParameter("Name", "Pester");
+                    var result = powerShell.Invoke();
+                    if (result != null && result.Count > 0)
+                    {
+                        foreach (var module in result)
+                        {
+                            if (module.BaseObject is PSModuleInfo psmoduleInfo)
```
SeeminglyScience:
```suggestion
                            if (module.BaseObject is PSModuleInfo psModuleInfo)
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+                            {
+                                if (psmoduleInfo.Version > new Version(4, 6))
```
SeeminglyScience:
```suggestion
                                if (psModuleInfo.Version > new Version(4, 6))
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+        private void DeterminePesterVersion()
+        {
+            Task.Run(() =>
+            {
+                using (var powerShell = System.Management.Automation.PowerShell.Create())
+                {
+                    powerShell.AddCommand("Get-Module")
+                              .AddParameter("ListAvailable")
+                              .AddParameter("Name", "Pester");
+                    var result = powerShell.Invoke();
```
SeeminglyScience:
```suggestion
                    IEnumerable<PSObject> result = powerShell.Invoke();
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -33,6 +40,38 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+                    if (result != null && result.Count > 0)
+                    {
+                        foreach (var module in result)
```
SeeminglyScience:
```suggestion
                        foreach (PSObject module in result)
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: It needs to be an `ICollection` at least because we call `.Count` on it, which is better (faster) than calling `.Any()` on it.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: The idea was that the default value of a boolean is false, which means the behavior will be the old one, when the async task finishes, then the value can be true and only then the system would switch over to the new improved behaviour, therefore I saw this race condition as by design. I am not a fan of constructor work either because it can make unit testing work hard, I could put the work in an `Initialize()` method instead if you prefer that.
I suggest to rather put a blocking wait when `GetPesterLens()` gets called to wait until the async task has reported the result if you want to completely avoid existing behaviour if someone clicks the Run Tests button too quickly immediately after opening the editor on a slow machine. I would then try to write the code that there is no block if the describe block string is not interpolated.
Would you be happy with that?

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: seeing failed attempts in the integrated console does not make for a good UX/impression, I would rather pre-pend some `if($IsPester460){}else{}` logic before doing that. Please my suggested alternative in the other comments below.

bergmeister:
Ok, I did some analysis/debugging and it seems that `PesterCodeLensProvider` gets constructed when opening any `.ps1` file. Although it would be nice to pre-initialise this in the background, I removed the async task from the constructor (I was thinking if we start adding more of that stuff, vscode will soon turn into a startup monster like `Visual Studio`). Therefore I changed it to a lazy evaluated and cached property that gets called in `GetPesterLens`, which will only get called once someone has a Pester tests file with a describe block.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,12 +72,40 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
 
             return codeLensResults;
         }
 
+        /// <summary>
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private static bool DeterminePesterVersion()
+        {
+            using (var powerShell = System.Management.Automation.PowerShell.Create())
```
TylerLeonhardt: Here you should use the `PSCommand` class I think. This is our abstraction over the PowerShell object to reuse runspaces.

Correct me if I'm wrong @rkeithhill & @SeeminglyScience

rkeithhill: In addition, PowerShellContext has quite a few `Execute` methods. Most are async but there should be some sync versions that execute commands.

bergmeister: Ok, I was thinking of passing a `PowerShellContext` to the constructor as I have seen that other places use that for the runspace. I did not know about `PSCommand`, sounds like an additional improvement (that I might look into applying for PSSA as well). Even if there are only async methods, we can still just call `Task.Wait()` to make it sync

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -25,6 +26,12 @@ internal class PesterCodeLensProvider : FeatureProviderBase, ICodeLensProvider
         /// </summary>
         private IDocumentSymbolProvider _symbolProvider;
 
+        /// <summary>
+        /// Pester 4.6.0 introduced a new ScriptblockFilter parameter to be able to run a test based on a line,
+        /// therefore knowing this information is important.
+        /// </summary>
+        private Lazy<bool> _pesterV4_6_0_OrHigherAvailable;
```
rjmholt:
```suggestion
        private readonly Lazy<bool> _pesterV4_6_0_OrHigherAvailable;
```

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,12 +74,43 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
             return codeLensResults;
         }
 
+        /// <summary>
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private bool DeterminePesterVersion()
+        {
+            var powerShell = new PSCommand();
+            powerShell.AddCommand("Get-Module")
```
rjmholt: Can omit the `; powerShell` here and chain the builder into the constructor if you like

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,12 +74,43 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+                      .AddParameter("ListAvailable")
+                      .AddParameter("Name", "Pester");
+
+            IEnumerable<PSObject> result = Task.Run(() =>
+            {
+                return _editorSession.PowerShellContext.ExecuteCommandAsync<PSObject>(powerShell);
+            }).Result;
+
+            if (result != null && result.Any())
```
rjmholt: Style suggestion to prevent indentation running away (["if you need more than 3 levels, you're screwed"](https://www.kernel.org/doc/html/v4.10/process/coding-style.html)):

```csharp
if (result == null) // result.Any() is redundant, since foreach is below
{
    return false;
}

foreach (PSObject module in result)
{
    if (module.BaseObject is PSModuleInfo psModuleInfo
        && psModuleInfo.Version >= s_pesterLineInvocationVersion)
    {
        return true;
    }
}

return false;
```

bergmeister:
I have replied and reacted to the PR comments. I think this PR is ready to be merged now, since its parent PR in the extension is now also approved as well.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,12 +74,43 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private bool DeterminePesterVersion()
+        {
+            var powerShell = new PSCommand()
+                .AddCommand("Get-Module")
+                .AddParameter("ListAvailable")
+                .AddParameter("Name", "Pester");
+
+            IEnumerable<PSObject> result = Task.Run(() =>
```
TylerLeonhardt: Is there a reason you're using `.Result` instead of using async/await?

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,12 +74,43 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
+                return _editorSession.PowerShellContext.ExecuteCommandAsync<PSObject>(powerShell);
+            }).Result;
+
+            if (result == null)
+            {
+                return false;
+            }
+
+            var minimumPesterVersionSupportingInlineInvocation = new Version(4, 6);
+            foreach (PSObject module in result)
```
TylerLeonhardt: What I'm worried about here is... 

Lets look at the `PSModulePath`:
```
/Users/tyler/.local/share/powershell/Modules                       # user level
/usr/local/share/powershell/Modules                                # all users level
/usr/local/microsoft/powershell/6-preview/Modules                  # internal
```

let's say they have version 4.5 in their _user module path_ ( but 4.6 in their _allusers module path_

This function will return true but 4.5 is what will get imported when doing:

`Import-Module Pester` 
or during module auto-loading

I think we should only check the first item in the result. This should be the Pester that would get imported.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: To make it synchronous since I moved it to the last possible point (when a .tests.ps1 file with a Describe block gets opened). If I did not call .Result (i.e. make it async but without await) then it would default to the old Pester syntax in the first 1-2 seconds until the value has been determined (which people did not like in the first version of the PR). It is still using Task APIs because the PSCommand APIs that were suggested are all async.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: Is that really the behaviour of PowerShell to not use the latest available version? Can you please point to some docs or code for my own education. How do you know that just looking at the first item is enough? `Get-Module` really needs to be improved otherwise to be more useful/predictive...
If that is really the case, then should we not rather call `(Get-Command Invoke-pester).Version` instead (which has the drawback of being 3 times more expensive for some reason...

TylerLeonhardt: Yes that's the behavior of PowerShell. I'm unaware of any docs out in the wild (cc @SteveL-MSFT or @joeyaiello maybe?).

The `PSModulePath` acts the same way that the `PATH` environment variable does - it goes down the path looking for the command you want to run, then executes the first one it finds.

A simple test:

```powershell
→ Install-Module Pester -Scope AllUsers -Force

→ Install-Module Pester -Scope CurrentUser -Force -MaximumVersion 4.5

→ Get-Command Invoke-Pester

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Function        Invoke-Pester                                      4.5.0      Pester

→ Get-Module -ListAvailable Pester


    Directory: /Users/tyler/.local/share/powershell/Modules

ModuleType Version    Name                                PSEdition ExportedCommands
---------- -------    ----                                --------- ----------------
Script     4.5.0      Pester                              Desk      {Describe, Context, It, Should…}

    Directory: /usr/local/share/powershell/Modules

ModuleType Version    Name                                PSEdition ExportedCommands
---------- -------    ----                                --------- ----------------
Script     4.6.0      Pester                              Desk      {Describe, Context, It, Should…}

```

TylerLeonhardt: > How do you know that just looking at the first item is enough?

Because when you do `Get-Module` the first item written to the pipeline is first on the `PSModulePath`. That might not be correct though... `Get-Command` is probably better for this.

bergmeister: Hmm, yes, this sounds like `Get-Command` is probably the better option. I guess we win the investment back later when `Invoke-Pester` actually gets called. Personally I would keep it in a non-blocking task though and just have it behave like the old system in the first 2 seconds of opening a Pester file...

TylerLeonhardt: Yeah... can @rjmholt and @SeeminglyScience help me understand why this should be synchronous?

rkeithhill: AFAICT the bulk of the Pester code lens implementation is sync (other than LSP message reading/writing).  This code is a bit old so maybe we weren't doing as much async then (??) or it could be that the specific symbol we are looking for results in a "fast enough" AST search.

One issue I have with this approach is that it doesn't take into account that I might have already imported an older version of Pester e.g. `Import-Module Pester -RequiredVersion 4.1.1`.  So we should check `Get-Module Pester` first to see if the module is already loaded before getting the list of Pester module.  

Another thing to consider is this feature seems to be implemented before PS Core was shipped.  That is, Windows PowerShell comes with Pester but not PS Core - right?  But this feature is still enabled resulting in this user exp:
```
Invoke-Pester : The term 'Invoke-Pester' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```
Perhaps we should consider shipping Pester with the extension like we do with PSSA?  That would ensure this feature works on a vanilla PS Core install.  This would ensure we have loaded a minimum version of Pester - like we do for PSSA.  But we still allow folks to install **later** versions of Pester and those would get loaded.  Again, very similiar to how we load PSSA.

TylerLeonhardt: @rkeithhill you touch on a lot of great points.

1. I think we _should_ make this Async. Lets take the `Lazy<bool>` and turn it into a `Lazy<Task<bool>>` and lazily await the value. Shouldn't be too hard to plumb that. The original caller of this function is async already.

2. 
> we should check `Get-Module Pester` first to see if the module is already loaded before getting the list of Pester module

This makes sense. But `Get-Command Invoke-Pester` should cover this if the module is already imported, right? `Get-Command Invoke-Pester` should be all we need, I think.

3. 

> Perhaps we should consider shipping Pester with the extension like we do with PSSA?

Some people rely on specific version of Pester. PowerShell itself relied on a specific version of Pester for a while in preparation for Core support from Pester.

I don't think we can add Pester into PSES and grab the latest version on the PSModulePath because I don't want to hurt customers by forcing a newer version of Pester on them and all of a sudden their CI and their local are different.

PSSA does fall in that camp, a bit... but PSSA is more for local analysis than anything else, I feel.

bergmeister: @TylerLeonhardt I agree with point 2 and 3. Re point to, we could potentially offer to install the latest version of Pester as a 3rd option in the dialog, but I see this as an optional, nice to have feature that can happen after this PR.
Regarding point1: I am not sure if we are on the same page: At the moment, when GetPesterLens gets called (which is when a `.tests.ps1` file gets opened and is therefore as late as possible), then we need to know the version of Pester already and have to wait for the result. When I first opened the PR I ran the check in a background thread (`Task.Run` without `.Result`), which meant that until it came back, it reported `false` (which is what most people did not like). I can change it back to this, but I do not need to make it an `Lazy<Task<bool>>` for that, I'd just remove the blocking call to `.Result`. The reason why I have to use `Task.Run` in the first place is because the recommended Execute APIs on the recommended `PSCommand` class only have async methods, therefore it does not matter at all if the function that we call is already async or not.
If you really wanted true async, then we'd have to fix calling methods up to 2 layers above to make it async top to bottom (which is quite a bit more involved for this change), otherwise you wouldn't get any real benefit from it

bergmeister: I made it use `Get-Command` now, which makes the code lens appear a second later due to it being more expensive and I also implemented async-await all the way to the top in a minimum viable way.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -63,18 +74,53 @@ public PesterCodeLensProvider(EditorSession editorSession)
...
 
             return codeLensResults;
         }
 
+        /// <summary>
+        /// Used to determine the value of <see cref="_pesterV4_6_0_OrHigherAvailable"/> as a background task.
+        /// </summary>
+        private async Task<bool> DeterminePesterVersion()
+        {
+            //if (System.Diagnostics.Debugger.IsAttached)
```
TylerLeonhardt: nit: can you remove this commented out code

src/PowerShellEditorServices.Host/CodeLens/ReferencesCodeLensProvider.cs
```diff
@@ -50,7 +49,7 @@ public ReferencesCodeLensProvider(EditorSession editorSession)
         /// </summary>
         /// <param name="scriptFile">The PowerShell script file to get code lenses for.</param>
         /// <returns>An array of CodeLenses describing all functions in the given script file.</returns>
-        public CodeLens[] ProvideCodeLenses(ScriptFile scriptFile)
+        public async Task<CodeLens[]> ProvideCodeLenses(ScriptFile scriptFile)
```
TylerLeonhardt: You made this function `async`, but didn't add any `await` statements so `async` isn't needed here.

src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
bergmeister: Sorry, I forgot to take this out when I was attaching the debugger. This is definitely not a nit, this is an oops

TylerLeonhardt: happens to the best of us 😊

src/PowerShellEditorServices.Host/CodeLens/ReferencesCodeLensProvider.cs
bergmeister: I had to change it here as well due to the signature change in the interface. The `PesterCodeLensProvider` properly implements async-await, which is what counts

TylerLeonhardt: Interesting! Typically the compiler yells if you have an `async` but no `await` but I guess since the interface requires it, the compiler doesn't yell.

bergmeister: Yep, it emits a warning though that this function won't run async as expected

TylerLeonhardt: You could changed the Function to return:

```csharp
return await Task.FromResult(acc.ToArray());
```

to make it happy I think.

bergmeister: I don't think this is a good idea because you might as well just suppress the warning. The compiler is happy, it is just kind to let us know that this other implementation will actually run synchronously despite the async declaration, Task.FromResult will still make it run sync but add unnecessary overhead. The warning is good to have because it is a way of tracking possible future improvement and also letting methods being called from this method know that they will actually need to adapt on a higher level if someone made them truly async as well

TylerLeonhardt:
@bergmeister just so I'm 100% clear. This PR tests if the Describe block is valid to show CodeLens at the _file open_ time?

Is there any reason for doing that over just displaying the CodeLens for _all_ Describe blocks and then show errors (or better recommendations saying "go get 4.6" 😄 ) for unsupported ones accordingly?

bergmeister:
As far as I understand:
When returning the code lens, it will show in the editor. Somehow, showing the code lens also requires providing the information about the line number and the boolean if Pester is compatible upfront, which is information being sent to the vscode extension. From a high level, yes, you'd want to separate the showing of the code lens from gathering the information last minute when someone actually clicks on it, but this would require a re-architecture.
At the moment, the typescript code knows all this information already when the user clicks on the test button. We'd need to send a request to PSES to gather the compatibility information last minute. Can someone comment on how much involved this would be?

rkeithhill:
I propose we close this in favor of PR #873

PR #859, opened 2019-01-22
Upgrade PowerShellStandard.Library, PowerShell.SDK, NET.Test.SDK and Serilog NuGet packages to latest released version and enable AppVeyor build on any branch

bergmeister:
This time I did not upgrade the XUnit packages as per the last PR feedback

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,11 +3,6 @@ image: Visual Studio 2017
 clone_depth: 10
 skip_tags: true
 
-branches:
-  only:
-  - master
-  - 2.0.0
-
 environment:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true  # Don't download unneeded packages
   DOTNET_CLI_TELEMETRY_OPTOUT: true        # Don't send telemetry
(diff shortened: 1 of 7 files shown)


appveyor.yml
```diff
@@ -3,11 +3,6 @@ image: Visual Studio 2017
 clone_depth: 10
 skip_tags: true
 
-branches:
```
rjmholt: @TylerLeonhardt do you think `master` and `legacy/1.x` is what we might want here?

test/PowerShellEditorServices.Test.Host/PowerShellEditorServices.Test.Host.csproj
```diff
@@ -22,13 +22,13 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="6.1.0-rc.1" />
+    <PackageReference Include="Microsoft.PowerShell.SDK" Version="6.1.0" />
```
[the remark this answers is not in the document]
rjmholt: No, that would be perfect. This is just for testing (not distributed to users), and there should be no breaking changes between 6.1.1 and 6.1.2 anyway.

appveyor.yml
rjmholt: I suspect we were misusing this now that I'm reading up on it. Looks like it should be fine to take out

bergmeister:
@TylerLeonhardt Ok, I bumped the ps sdk to the latest version. I think we can merge now.

PR #862, opened 2019-01-23
Handle arbitrary exceptions when recursing workspace

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/1704 (not the original issue, but that's tracked elsewhere).

This should also go into `legacy/1.x`

### DIFF
--- a/src/PowerShellEditorServices/Workspace/Workspace.cs
+++ b/src/PowerShellEditorServices/Workspace/Workspace.cs
@@ -350,6 +350,7 @@ namespace Microsoft.PowerShell.EditorServices
                     this.logger.WriteHandledException(
                         $"Could not enumerate files in the path '{folderPath}' due to an exception",
                         e);
+                    continue;
                 }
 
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -350,6 +350,7 @@ private void RecursivelyEnumerateFiles(string folderPath, ref List<string> found
                     this.logger.WriteHandledException(
                         $"Could not enumerate files in the path '{folderPath}' due to an exception",
                         e);
+                    continue;
```
rkeithhill: Minor nit, all the other `continue` usages above are preceded by a blank line.  Ditto the one below on line 401.

rkeithhill:
Is this going to make it into 1.11.0?

rjmholt:
> Is this going to make it into 1.11.0?

Unfortunately no (I think that's just been published). But I'm sure we'll have a 1.11.1

PR #867, opened 2019-02-07
Invoke expand-archive using module qualified name

rkeithhill:
I have another module with an Expand-Archive command, which breaks my
build.


rjmholt:
Is it worth also fully qualifying `Invoke-WebRequest` and `Remove-Item`?

rkeithhill:
I created a snapin back in 2006/7 on the pre-cursor to CodePlex.  The snapin (now module) is called PSCX and we had Expand-Archive way before PowerShell got it.  Ditto for Unblock-File, Convert-Xml, Get/Set-Clipboard, Format-Hex and a few others.

I'd be happy to module qualify iwr and ri.  Let me push an update to this PR.

rkeithhill:
You know what, upon further reflection, I'm going to fix this problem in my PSCX install.  So RE this PR ... nevermind.  :-)

PR #872, opened 2019-02-28
Fix unable to open files in problems/peek windows issue

rkeithhill:
This is due to PSES not properly storing a language client path in the
ClientFilePath property of ScriptFile.  This change ensures that a
ClientFilePath is always stored in the text document Uri that a LSP
client expects.  This code has mostly been provided by
@seeminglyScience. Thanks for the reference to your code.
This fixes issue [1732 in the vscode-powershell repo](https://github.com/PowerShell/vscode-powershell/issues/1732).

BTW I've also verified that there are issues when you use the reference code lens where refs to other files that you haven't loaded, will not load when you click on the file in the peek window.

### DIFF
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -3,13 +3,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using Microsoft.PowerShell.EditorServices.Utility;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Management.Automation;
 using System.Management.Automation.Language;
+using System.Runtime.InteropServices;
+using Microsoft.PowerShell.EditorServices.Utility;
 
 namespace Microsoft.PowerShell.EditorServices
 {
@@ -27,6 +28,7 @@ namespace Microsoft.PowerShell.EditorServices
         };
 
         private Version powerShellVersion;
+        private string _clientPath;
 
         #endregion
 
@@ -50,7 +52,20 @@ namespace Microsoft.PowerShell.EditorServices
         /// <summary>
         /// Gets the path which the editor client uses to identify this file.
         /// </summary>
-        public string ClientFilePath { get; private set; }
+        public string ClientFilePath
+        {
+            get { return _clientPath; }
+
+            private set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+
+                _clientPath = GetPathAsClientPath(value);
+            }
+        }
 
         /// <summary>
         /// Gets or sets a boolean that determines whether
@@ -563,6 +578,43 @@ namespace Microsoft.PowerShell.EditorServices
         #endregion
 
         #region Private Methods
+
+        private static string GetPathAsClientPath(string path)
+        {
+            const string fileUriPrefix = "file:///";
+
+            if (path.StartsWith("untitled:", StringComparison.Ordinal))
+            {
+                return path;
+            }
+
+            if (path.StartsWith("file:///", StringComparison.Ordinal))
+            {
+                return path;
+            }
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return new Uri(path).AbsoluteUri;
+            }
+
+            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
+            // URI encoded. System.Uri won't do that, so we manually create the URI.
+            var newUri = System.Web.HttpUtility.UrlPathEncode(path);
+            int colonIndex = path.IndexOf(":");
+            for (var i = colonIndex - 1; i >= 0; i--)
+            {
+                newUri.Remove(i, 1);
+                newUri.Insert(i, char.ToLowerInvariant(path[i]).ToString());
+            }
+
+            return newUri
+                .Remove(colonIndex, 1)
+                .Insert(colonIndex, "%3A")
+                .Replace("\\", "/")
+                .Insert(0, fileUriPrefix)
+                .ToString();
+        }
 
         private void SetFileContents(string fileContents)
         {
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -60,7 +60,7 @@ namespace Microsoft.PowerShell.EditorServices
         {
             get
             {
-                return (this.ClientFilePath == null )
+                return (this.ClientFilePath == null)
                     ? string.Empty
                     : Workspace.ConvertPathToDocumentUri(this.ClientFilePath);
             }
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -578,43 +578,6 @@ namespace Microsoft.PowerShell.EditorServices
 
         #region Private Methods
 
-        private static string GetPathAsClientPath(string path)
-        {
-            const string fileUriPrefix = "file:///";
-
-            if (path.StartsWith("untitled:", StringComparison.Ordinal))
-            {
-                return path;
-            }
-
-            if (path.StartsWith("file:///", StringComparison.Ordinal))
-            {
-                return path;
-            }
-
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                return new Uri(path).AbsoluteUri;
-            }
-
-            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
-            // URI encoded. System.Uri won't do that, so we manually create the URI.
-            var newUri = System.Web.HttpUtility.UrlPathEncode(path);
-            int colonIndex = path.IndexOf(":");
-            for (var i = colonIndex - 1; i >= 0; i--)
-            {
-                newUri.Remove(i, 1);
-                newUri.Insert(i, char.ToLowerInvariant(path[i]).ToString());
-            }
-
-            return newUri
-                .Remove(colonIndex, 1)
-                .Insert(colonIndex, "%3A")
-                .Replace("\\", "/")
-                .Insert(0, fileUriPrefix)
-                .ToString();
-        }
-
         private void SetFileContents(string fileContents)
         {
             // Split the file contents into lines and trim
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -60,7 +60,7 @@ namespace Microsoft.PowerShell.EditorServices
         {
             get
             {
-                return (this.ClientFilePath == null)
+                return this.ClientFilePath == null
                     ? string.Empty
                     : Workspace.ConvertPathToDocumentUri(this.ClientFilePath);
             }
--- a/src/PowerShellEditorServices/Workspace/ScriptFile.cs
+++ b/src/PowerShellEditorServices/Workspace/ScriptFile.cs
@@ -28,7 +28,6 @@ namespace Microsoft.PowerShell.EditorServices
         };
 
         private Version powerShellVersion;
-        private string _clientPath;
 
         #endregion
 
@@ -52,18 +51,18 @@ namespace Microsoft.PowerShell.EditorServices
         /// <summary>
         /// Gets the path which the editor client uses to identify this file.
         /// </summary>
-        public string ClientFilePath
+        public string ClientFilePath { get; private set; }
+
+        /// <summary>
+        /// Gets the file path in LSP DocumentUri form.  The ClientPath property must not be null.
+        /// </summary>
+        public string DocumentUri
         {
-            get { return _clientPath; }
-
-            private set
+            get
             {
-                if (value == null)
-                {
-                    throw new ArgumentNullException(nameof(value));
-                }
-
-                _clientPath = GetPathAsClientPath(value);
+                return (this.ClientFilePath == null )
+                    ? string.Empty
+                    : Workspace.ConvertPathToDocumentUri(this.ClientFilePath);
             }
         }
 
--- a/src/PowerShellEditorServices/Workspace/Workspace.cs
+++ b/src/PowerShellEditorServices/Workspace/Workspace.cs
@@ -351,7 +351,7 @@ namespace Microsoft.PowerShell.EditorServices
                     this.logger.WriteHandledException(
                         $"Could not enumerate files in the path '{folderPath}' due to an exception",
                         e);
-                    
+
                     continue;
                 }
 
@@ -400,7 +400,7 @@ namespace Microsoft.PowerShell.EditorServices
                 this.logger.WriteHandledException(
                     $"Could not enumerate directories in the path '{folderPath}' due to an exception",
                     e);
-                
+
                 return;
             }
 
@@ -623,6 +623,59 @@ namespace Microsoft.PowerShell.EditorServices
             sb.Append(fileUri.Substring(12)); // The rest of the URI after the colon
 
             return sb.ToString();
+        }
+
+        /// <summary>
+        /// Converts a file system path into a DocumentUri required by Language Server Protocol.
+        /// </summary>
+        /// <remarks>
+        /// When sending a document path to a LSP client, the path must be provided as a
+        /// DocumentUri in order to features like the Problems window or peek definition
+        /// to be able to open the specified file.
+        /// </remarks>
+        /// <param name="path">
+        /// A file system path. Note: if the path is already a DocumentUri, it will be returned unmodified.
+        /// </param>
+        /// <returns>The file system path encoded as a DocumentUri.</returns>
+        internal static string ConvertPathToDocumentUri(string path)
+        {
+            const string fileUriPrefix = "file:///";
+
+            if (path.StartsWith("untitled:", StringComparison.Ordinal))
+            {
+                return path;
+            }
+
+            if (path.StartsWith(fileUriPrefix, StringComparison.Ordinal))
+            {
+                return path;
+            }
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return new Uri(path).AbsoluteUri;
+            }
+
+            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
+            // URI encoded. System.Uri won't do that, so we manually create the URI.
+            var newUri = System.Web.HttpUtility.UrlPathEncode(path);
+            int colonIndex = path.IndexOf(':');
+            for (var i = colonIndex - 1; i >= 0; i--)
+            {
+                newUri.Remove(i, 1);
+                newUri.Insert(i, char.ToLowerInvariant(path[i]).ToString());
+            }
+
+            // On a Linux filesystem, you can have multiple colons in a filename e.g. foo:bar:baz.txt
+            if (colonIndex >= 0)
+            {
+                newUri = newUri.Replace(":", "%3A");
+            }
+
+            return newUri
+                .Replace('\\', '/')
+                .Insert(0, fileUriPrefix)
+                .ToString();
         }
 
         #endregion
--- a/src/PowerShellEditorServices/Workspace/Workspace.cs
+++ b/src/PowerShellEditorServices/Workspace/Workspace.cs
@@ -640,6 +640,7 @@ namespace Microsoft.PowerShell.EditorServices
         internal static string ConvertPathToDocumentUri(string path)
         {
             const string fileUriPrefix = "file:///";
+            int colonIndex;
 
             if (path.StartsWith("untitled:", StringComparison.Ordinal))
             {
@@ -654,13 +655,24 @@ namespace Microsoft.PowerShell.EditorServices
             if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 // On a Linux filesystem, you can have multiple colons in a filename e.g. foo:bar:baz.txt
-                return new Uri(path).AbsoluteUri.Replace(":", "%3A");
+                string absoluteUri = new Uri(path).AbsoluteUri;
+
+                // First colon is part of the protocol scheme, see if there are other colons in the path
+                int firstColonIndex = absoluteUri.IndexOf(':');
+                if (absoluteUri.IndexOf(':', firstColonIndex + 1) >= 0)
+                {
+                    absoluteUri =
+                        absoluteUri.Substring(0, firstColonIndex + 1) +
+                        absoluteUri.Substring(firstColonIndex + 1).Replace(":", "%3A");
+                }
+
+                return absoluteUri;
             }
 
             // VSCode file URIs on Windows need the drive letter lowercase, and the colon
             // URI encoded. System.Uri won't do that, so we manually create the URI.
             var newUri = System.Web.HttpUtility.UrlPathEncode(path);
-            int colonIndex = path.IndexOf(':');
+            colonIndex = path.IndexOf(':');
             if (colonIndex > 0)
             {
                 int driveLetterIndex = colonIndex - 1;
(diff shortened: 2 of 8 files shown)


rkeithhill:
Well, looks like that bombs a lot of tests. I'll take a look at the tests tomorrow and see if this is a case where we can "fix" the tests.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -564,6 +579,43 @@ public BufferRange GetRangeBetweenOffsets(int startOffset, int endOffset)
...
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return new Uri(path).AbsoluteUri;
+            }
+
+            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
+            // URI encoded. System.Uri won't do that, so we manually create the URI.
+            var newUri = System.Web.HttpUtility.UrlPathEncode(path);
+            int colonIndex = path.IndexOf(":");
```
rjmholt: I've already written logic for this somewhere in PSES I think, so ideally we should use that as a static method

rjmholt: Also, I opened an issue on VSCode about it and they tried changing it (to their credit) and it didn't work out. The URI spec says to use `:` for drive letters, but alas...

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -564,6 +579,43 @@ public BufferRange GetRangeBetweenOffsets(int startOffset, int endOffset)
 
         #region Private Methods
 
+        private static string GetPathAsClientPath(string path)
+        {
+            const string fileUriPrefix = "file:///";
```
rjmholt: This doesn't seem to be used -- or at least is repeated below

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rjmholt: Here: https://github.com/PowerShell/PowerShellEditorServices/blob/1a86d4b159c17f382aa148942598e140d7c0070d/src/PowerShellEditorServices/Workspace/Workspace.cs#L595-L626

rjmholt: Oh we need the opposite method!

rjmholt: We should stick them in the same place

rkeithhill:
OK, the latest update leaves only one (non-test) use of `ClientFilePath` and that happens starting here:

https://github.com/PowerShell/PowerShellEditorServices/blob/1a86d4b159c17f382aa148942598e140d7c0070d/src/PowerShellEditorServices/Session/RemoteFileManager.cs#L596

which calls (and uses ClientPath):

https://github.com/PowerShell/PowerShellEditorServices/blob/1a86d4b159c17f382aa148942598e140d7c0070d/src/PowerShellEditorServices/Extensions/FileContext.cs#L240

which sends this insert test request to the LSP client:

https://github.com/PowerShell/PowerShellEditorServices/blob/1a86d4b159c17f382aa148942598e140d7c0070d/src/PowerShellEditorServices.Protocol/Server/LanguageServerEditorOperations.cs#L46

Not sure if we could have a latent bug here where we pass a path that hasn't been converted to a DocumentUri.  So I've left this one alone.  I think it is used when you edit a remote file.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rkeithhill: Done.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rkeithhill: Fixed.

rkeithhill:
OK, I think this is ready for review again.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -50,7 +51,18 @@ public string Id
         /// <summary>
         /// Gets the path which the editor client uses to identify this file.
         /// </summary>
-        public string ClientFilePath { get; private set; }
+        public string ClientFilePath { get; private set; } 
+
+        /// <summary>
+        /// Gets the ClientPath in LSP DocumentUri form.  The ClientPath property must not be null.
+        /// </summary>
+        public string ClientPathAsDocumentUri
```
SeeminglyScience: Would `DocumentUri` make sense here instead?

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -50,7 +51,18 @@ public string Id
...
+        {
+            get 
+            {
+                return Workspace.ConvertPathToDocumentUri(this.ClientFilePath);
```
SeeminglyScience: Could we make it return `null` or `string.Empty` if `ClientFilePath` is null?  `ScriptFile` is actually accessible in a few ways without reflection.

```suggestion
                return this.ClientFilePath == null
                    ? string.Empty
                    : Workspace.ConvertPathToDocumentUri(this.ClientFilePath);
```

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -624,6 +624,59 @@ private static string UnescapeDriveColon(string fileUri)
...
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return new Uri(path).AbsoluteUri;
+            }
+
+            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
+            // URI encoded. System.Uri won't do that, so we manually create the URI.
+            var newUri = System.Web.HttpUtility.UrlPathEncode(path);
+            int colonIndex = path.IndexOf(":", StringComparison.Ordinal);
```
SeeminglyScience: Would this work here instead?

```suggestion
            int colonIndex = path.IndexOf(':');
```

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rkeithhill: Probably makes more sense.  I was trying to "tie" the two properties together by name but I think eventually we want `ClientPath` to just go away.  I'm waiting on word from @TylerLeonhardt about the remote file editing scenario, besides the tests, that is the only other place that uses `ClientPath`.

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Do you think that will shut up the Codacy locale warning?

SeeminglyScience: If not, in this instance it may be better to ignore or suppress it.  I'm guessing the method for finding the index of an entire string is a lot more complicated than the one for a single character.  It'd be nice if there was an equivalent culture specific overload though :/

rjmholt: Is one a LINQ-provided method? I've seen ones like `Contains()` where LINQ provides the `char` overload, and I wonder if the `string`-specific version is possibly faster. It probably uses [Boyer-Moore](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_string-search_algorithm), which for a longer string would be faster than sequential search, but here probably degenerates to "about as fast as a char search" but with locale settings possible.

SeeminglyScience: Nah, LINQ doesn't usually deal with indexes since `IEnumerable<>` doesn't implement an indexer.  Closest they have is `ElementAt`.

`string.IndexOf(char)` goes right to `string.IndexOf(char, int, int)` which is an internal call

`string.IndexOf(string, StringComparison)` goes through a few layers of checks and methods before hitting `CompareInfo.InternalFindNLSStringEx` which is a QCall.

The first almost definitely has better performance, but that's not really my concern.  I think that warning only makes sense if being more explicit doesn't drastically change the code path.  Also maybe it's just me but `char` seems to have a much heavier implication of ordinal compare.

Neither will be a measurable change though, so nbd ya'll disagree.

rjmholt: (Looks like .NET [doesn't use Boyer-Moore in this case](https://github.com/dotnet/corefx/blob/5e3fe25d1236cd0192dd966f7286fe008d5cf875/src/Common/src/CoreLib/System/Globalization/CompareInfo.cs#L545-L591), but it does look fairly tuned at least)

TylerLeonhardt:
This is great @rkeithhill! A couple things...

Can you resolve the merge conflict?

Is it possible to add tests for this change? (I only ask because I think this part of the code is very testable)

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rkeithhill: Done

src/PowerShellEditorServices/Workspace/ScriptFile.cs
rkeithhill: Done.

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Trying the signature that takes a char. We'll see if Codacy complains.

rkeithhill:
Resovled the merge conflicts and addresses other PR comments. I'll look into tests next.

rkeithhill:
BTW it would be nice if running tests locally was more convenient and more reliable.  For instance, when I run the tests from the Visual Studio Test Explorer window, I get 55 failures, yet I get zero failures fom the command line.  That is, when the tests don't hang.  Often when running from the command line, the tests hang.  :-(

rjmholt:
> BTW it would be nice if running tests locally was more convenient and more reliable. For instance, when I run the tests from the Visual Studio Test Explorer window, I get 55 failures, yet I get zero failures fom the command line. That is, when the tests don't hang. Often when running from the command line, the tests hang. :-(

Yeah testing is what I want to focus on when I next get the chance. Once we have better testing we can make the changes we so badly need to make without breaking everything else

rkeithhill:
@TylerLeonhardt Good call on adding tests.  I think I've found an issue but I need to spin up my Ubuntu VM to verify.

TylerLeonhardt:
Let me know if there's anything I can do to help!

TylerLeonhardt:
Commit: Try to fix tests for Linux ✔️

Nice! Do you think this is ready for review now?

rkeithhill:
Yes. It is ready for review. Also could you test this on macOS or Linux?  In the released version of the extension, open the examples folder.  Insert a few lines of script into `StopTest.ps1` that will generate PSSA warnings (use iex or unused variable). The save & close the file.  Now reload VSCode and open Stop-Process2.ps1.  After PS initializes, click on the references on the `Stop-Process2` function.  While the reference is being displayed in the peek window, you should see a problem appear in the Problems window for the script you added.  Double click on the problem to navigate to it. That is when you should see this:

image

Now build VSCode with this PR pulled down in PSES and try the above experiment again and it should work this time.  

BTW sometimes I've seen references fail before when you try to navigate to a file. Still trying to find a repro for that one.

TylerLeonhardt:
Before:
image

After 😊
image

And when I click on them, they navigate successfully!

Awesome work!

rkeithhill:
Thanks.  So you are seeing the extension dev host use the ISE color theme as well.  Hmm...  I wonder what is up with that.

rkeithhill:
@SeeminglyScience Would you mind taking a peek at this again?  I changed some of the logic and  would like to get you to review it.  The main thing I found is that Linux allows file names like `foo:bar:baz.txt`.

src/PowerShellEditorServices/Workspace/ScriptFile.cs
```diff
@@ -52,6 +53,19 @@ public string Id
...
         public string ClientFilePath { get; private set; }
 
+        /// <summary>
+        /// Gets the file path in LSP DocumentUri form.  The ClientPath property must not be null.
+        /// </summary>
+        public string DocumentUri
+        {
+            get
+            {
+                return (this.ClientFilePath == null )
```
SeeminglyScience:
```suggestion
                return (this.ClientFilePath == null)
```

Or

```suggestion
                return this.ClientFilePath == null
```

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,64 @@ private static string UnescapeDriveColon(string fileUri)
...
+                // On a Linux filesystem, you can have multiple colons in a filename e.g. foo:bar:baz.txt
+                string absoluteUri = new Uri(path).AbsoluteUri;
+
+                // First colon is part of the protocol scheme, see if there are other colons in the path
+                int firstColonIndex = absoluteUri.IndexOf(':');
+                if (absoluteUri.IndexOf(':', firstColonIndex + 1) >= 0)
+                {
+                    absoluteUri =
+                        absoluteUri.Substring(0, firstColonIndex + 1) +
+                        absoluteUri.Substring(firstColonIndex + 1).Replace(":", "%3A");
```
SeeminglyScience: Would be nice to cut out some string allocations here

```csharp
absoluteUri = new StringBuilder()
    .Append(absoluteUri, firstColonIndex + 1, absoluteUri.Length - firstColonIndex)
    .Replace(":", "%3A")
    .Insert(0, absoluteUri, firstColonIndex + 1)
    .ToString();
```

Or maybe `Append` the whole `absoluteUri` and use the `Replace` overload that takes index/count

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,64 @@ private static string UnescapeDriveColon(string fileUri)
...
+            int colonIndex = path.IndexOf(':');
+            if (colonIndex > 0)
+            {
+                int driveLetterIndex = colonIndex - 1;
+                var driveLetter = char.ToLowerInvariant(path[driveLetterIndex]).ToString();
+                newUri = newUri.Remove(driveLetterIndex, 2);
+                newUri = newUri.Insert(driveLetterIndex, driveLetter + "%3A");
+            }
+
+            return newUri.Replace('\\', '/').Insert(0, fileUriPrefix);
```
SeeminglyScience: re: strings, maybe

```csharp
var newUri = new StringBuilder(System.Web.HttpUtility.UrlPathEncode(path));
int colonIndex = path.IndexOf(':');
if (colonIndex > 0)
{
    int driveLetterIndex = colonIndex - 1;
    char driveLetter = char.ToLowerInvariant(path[driveLetterIndex]);
    newUri
        .Remove(driveLetterIndex, 2)
        .Insert(driveLetterIndex, driveLetter)
        .Insert(driveLetterIndex + 1, "%3A");
}

return newUri.Replace('\\', '/').Insert(0, fileUriPrefix).ToString();
```

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Good suggestion!  I like it.

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Yeah, that's better. :-)  Thanks!

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Interesting that `Insert()` doesn't do what you (or I) think it does.  It inserts **copies** of the string where the third parameter is the number of copies.  So going to have one more alloc - `.ToCharArray()`.

rkeithhill:
To quote Lego Batman - "What the heck!?"
```
Build started
2
git clone -q --depth=10 https://github.com/PowerShell/PowerShellEditorServices.git C:\projects\powershelleditorservices
3
fatal: unable to access 'https://github.com/PowerShell/PowerShellEditorServices.git/': Could not resolve host: github.com
4
Command exited with code 128
```
How do you go about kicking a new build without having to push an unnecessary commit?

rkeithhill:
OK, just about ready to merge this.  @TylerLeonhardt could you run that test one  more time since we changed the Linux/macOS code path a bit (based on @SeeminglyScience feedback)?  Thanks.

src/PowerShellEditorServices/Workspace/Workspace.cs
SeeminglyScience: > Interesting that Insert() method doesn't do what you (or I) think it does. It inserts copies of the string where the third parameter is the number of copies.

Yeah, didn't expect that! Good catch 🙂

test/PowerShellEditorServices.Test/Session/ScriptFileTests.cs
```diff
@@ -580,5 +581,31 @@ public void PropertiesInitializedCorrectlyForUntitled()
...
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+            {
+                path = @"C:\Users\AmosBurton\projects\Rocinate\ProtoMolecule.ps1";
+                scriptFile = new ScriptFile(path, path, emptyStringReader, PowerShellVersion);
+                Assert.Equal("file:///c%3A/Users/AmosBurton/projects/Rocinate/ProtoMolecule.ps1", scriptFile.DocumentUri);
+            }
+            else
+            {
+                // Test the following only on Linux and macOS.
+                path = "/home/AmosBurton/projects/Rocinate/ProtoMolecule.ps1";
```
TylerLeonhardt: what an obscure reference :P

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,68 @@ private static string UnescapeDriveColon(string fileUri)
...
+                string absoluteUri = new Uri(path).AbsoluteUri;
+
+                // First colon is part of the protocol scheme, see if there are other colons in the path
+                int firstColonIndex = absoluteUri.IndexOf(':');
+                if (absoluteUri.IndexOf(':', firstColonIndex + 1) >= 0)
+                {
+                    absoluteUri = new StringBuilder()
+                        .Append(absoluteUri, firstColonIndex + 1, absoluteUri.Length - firstColonIndex - 1)
+                        .Replace(":", "%3A")
+                        .Insert(0, absoluteUri.ToCharArray(0, firstColonIndex + 1))
```
TylerLeonhardt: StringBuilder has an overload:
```
StringBuilder StringBuilder.Replace(string oldValue, string newValue, int startIndex, int count)
```

So we could do something like:

```csharp
absoluteUri = new StringBuilder(absoluteUri)
    .Replace(
        oldValue: ":", 
        newValue:"%3A",
        startIndex: firstColonIndex + 1
        count: absoluteUri.Length) // in the event that every character is a `:` lol
    .ToString();
```

test/PowerShellEditorServices.Test/Session/ScriptFileTests.cs
rkeithhill: I'm all about obscure references.  :-)  Probably because I'm on book 7 now - been living with it for a while.  After we saw Lego Batman, all my Slack startup messages were Batman quotes.

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,68 @@ private static string UnescapeDriveColon(string fileUri)
...
+
+            // VSCode file URIs on Windows need the drive letter lowercase, and the colon
+            // URI encoded. System.Uri won't do that, so we manually create the URI.
+            var newUri = new StringBuilder(System.Web.HttpUtility.UrlPathEncode(path));
+            int colonIndex = path.IndexOf(':');
+            if (colonIndex > 0)
+            {
+                int driveLetterIndex = colonIndex - 1;
+                char driveLetter = char.ToLowerInvariant(path[driveLetterIndex]);
+                newUri
```
TylerLeonhardt: I think this can also use that overload I mentioned:

```csharp
newUri.Replace(path[driveLetterIndex], driveLetter, driveLetterIndex, 1);
```

and
```csharp
newUri.Replace(":", "%3A", colonIndex, 1);
```

TylerLeonhardt: these could be combined into one `Replace` call too in theory.

rkeithhill:
Can someone rekick the build?

src/PowerShellEditorServices/Workspace/Workspace.cs
TylerLeonhardt: I realize that you don't store this to a variable. Does it replace in-place?

test/PowerShellEditorServices.Test/Session/ScriptFileTests.cs
```diff
@@ -580,5 +581,31 @@ public void PropertiesInitializedCorrectlyForUntitled()
...
+                scriptFile = new ScriptFile(path, path, emptyStringReader, PowerShellVersion);
+                Assert.Equal("file:///home/AmosBurton/projects/Rocinate/ProtoMolecule.ps1", scriptFile.DocumentUri);
+
+                path = "/home/AmosBurton/projects/Rocinate/Proto:Mole:cule.ps1";
```
TylerLeonhardt: Should we also add a test case for when the path contains at `\` on non-Windows or should that really not matter?

rkeithhill: Probably but right now VSCode chokes on that test case so not possible to encounter AFAICT from VSCode.  Also the code needs to be updated to support `\` (convert to `%5C`).  Thoughts?

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,69 @@ private static string UnescapeDriveColon(string fileUri)
...
```
TylerLeonhardt: I realize that you don't store this to a variable. Does it replace in-place?

test/PowerShellEditorServices.Test/Session/ScriptFileTests.cs
TylerLeonhardt: System.Uri should already do this?

```
[System.Uri]::new("/user\foo/bar").AbsoluteUri
file:///user%5Cfoo/bar
```

rkeithhill: Sorry, `Uri` handles this but not `UrlPathEncode`.  Been a bit too focused on that code path.  For the Linux/macOS code path, we use `Uri` so yeah, we should be good. I'll add the test to verify that.

src/PowerShellEditorServices/Workspace/Workspace.cs
rkeithhill: Yes, the StringBuilder Replace method replaces in-place within the StringBuilder.

TylerLeonhardt: and it also returns the object reference so you can chain them I guess? Neat.

src/PowerShellEditorServices/Workspace/Workspace.cs
```diff
@@ -625,6 +625,69 @@ private static string UnescapeDriveColon(string fileUri)
...
+            }
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // On a Linux filesystem, you can have multiple colons in a filename e.g. foo:bar:baz.txt
+                string absoluteUri = new Uri(path).AbsoluteUri;
+
+                // First colon is part of the protocol scheme, see if there are other colons in the path
+                int firstColonIndex = absoluteUri.IndexOf(':');
+                if (absoluteUri.IndexOf(':', firstColonIndex + 1) >= 0)
```
TylerLeonhardt: nit: it's just slightly more correct because we expect it to be an index greater than firstColonIndex.
```suggestion
                if (absoluteUri.IndexOf(':', firstColonIndex + 1) > firstColonIndex)
```

TylerLeonhardt:
those flaky tests are so frustrating but I haven't had time to fix/mitigate them... I'm thinking about implementing 5x retry logic.

TylerLeonhardt:
@SeeminglyScience @rjmholt if you guys wanna give this one more look after my comments... otherwise I think this is good to go in

TylerLeonhardt:
p.s. this PR still works on macOS 😄

SeeminglyScience:
Perfection

TylerLeonhardt:
@rjmholt is OOF, merging this in. I marked it for backport. @rkeithhill we should cherry pick this back and test it.

PR #873, opened 2019-03-03
Return the start line number for Describe block

rkeithhill:
Supports PR 1776 in vscode-powershell

### DIFF
--- a/src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
+++ b/src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
@@ -54,7 +54,11 @@ namespace Microsoft.PowerShell.EditorServices.CodeLenses
                     new ClientCommand(
                         "PowerShell.RunPesterTests",
                         "Run tests",
-                        new object[] { scriptFile.ClientFilePath, false /* No debug */, pesterSymbol.TestName })),
+                        new object[] {
+                            scriptFile.ClientFilePath,
+                            false /* No debug */,
+                            pesterSymbol.TestName,
+                            pesterSymbol.ScriptRegion.StartLineNumber })),
 
                 new CodeLens(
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Host/CodeLens/PesterCodeLensProvider.cs
```diff
@@ -54,7 +54,11 @@ public PesterCodeLensProvider(EditorSession editorSession)
                     new ClientCommand(
                         "PowerShell.RunPesterTests",
                         "Run tests",
-                        new object[] { scriptFile.ClientFilePath, false /* No debug */, pesterSymbol.TestName })),
+                        new object[] {
+                            scriptFile.ClientFilePath,
+                            false /* No debug */,
+                            pesterSymbol.TestName,
+                            pesterSymbol.ScriptRegion.StartLineNumber })),
```
rkeithhill: Can ScriptRegion ever be null? Given that we get the PesterSymbolReference from the AST, I don't think so. We could make this `pesterSymbol.ScriptRegion?.StartLineNumber ?? null` but I'm not sure that is necessary.

rjmholt: There's nothing in the immediate constructor to prevent it being null. It feels like we shouldn't need it, but having it crash would be much worse than just the null check.

The `?? null` shouldn't be needed in that case I don't think -- the `?.StartLineNumber` should resolve to null by itself.

rkeithhill: Duh.  Good point.  :-)  I'll make the change just to be on the safe side.

rkeithhill:
The two test failures are debug bouncer related (typical/intermittent failure). Can someone kick the build again or I can just merge this?

PR #875, opened 2019-03-07
Add attach to local runspace.

adamdriscoll:
Adds support for attaching to a local runspace. This extends the existing support for attaching to a runspace in a remote process. This enables debugging for modules that create additional runspaces such as ThreadJob, PoshRSJob and UnivesalDashboard.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
@@ -20,8 +20,6 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
 
         public string ProcessId { get; set; }
 
-        public int RunspaceId { get; set; }
-
-        public string LocalRunspaceId { get; set; }
+        public string RunspaceId { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
+++ b/src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
@@ -21,5 +21,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.DebugAdapter
         public string ProcessId { get; set; }
 
         public int RunspaceId { get; set; }
+
+        public string LocalRunspaceId { get; set; }
     }
 }
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -439,22 +439,18 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                 // event gets fired with the attached runspace.
 
                 int runspaceId = 1;
-                if (string.IsNullOrEmpty(attachParams.LocalRunspaceId))
+                if (int.TryParse(attachParams.RunspaceId, out runspaceId))
                 {
-                    runspaceId = attachParams.RunspaceId > 0 ? attachParams.RunspaceId : 1;
-                }
-                else if (int.TryParse(attachParams.LocalRunspaceId, out int localRunspaceId))
-                {
-                    runspaceId = localRunspaceId;
+                    runspaceId = runspaceId > 0 ? runspaceId : 1;
                 }
                 else
                 {
                     Logger.Write(
                         LogLevel.Error,
-                        $"Attach request failed, '{attachParams.LocalRunspaceId}' is an invalid value for the processId.");
+                        $"Attach request failed, '{attachParams.RunspaceId}' is an invalid value for the processId.");
 
                     await requestContext.SendErrorAsync(
-                        "A positive integer must be specified for the LocalRunspaceId field.");
+                        "A positive integer must be specified for the RunspaceId field.");
 
                     return;
                 }
--- a/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
@@ -403,7 +403,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                 _isRemoteAttach = true;
             }
 
-            if (int.TryParse(attachParams.ProcessId, out int processId) && (processId > 0))
+            if ((int.TryParse(attachParams.ProcessId, out int processId) && (processId > 0)) || attachParams.ProcessId == "current")
             {
                 PowerShellVersionDetails runspaceVersion =
                     _editorSession.PowerShellContext.CurrentRunspace.PowerShellVersion;
@@ -416,16 +416,18 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.Server
                     return;
                 }
 
-                await _editorSession.PowerShellContext.ExecuteScriptStringAsync(
+                if (attachParams.ProcessId != "current") {
+                    await _editorSession.PowerShellContext.ExecuteScriptStringAsync(
                     $"Enter-PSHostProcess -Id {processId}",
                     errorMessages);
 
-                if (errorMessages.Length > 0)
-                {
-                    await requestContext.SendErrorAsync(
-                        $"Could not attach to process '{processId}'");
+                    if (errorMessages.Length > 0)
+                    {
+                        await requestContext.SendErrorAsync(
+                            $"Could not attach to process '{processId}'");
 
-                    return;
+                        return;
+                    }
                 }
 
                 // Clear any existing breakpoints before proceeding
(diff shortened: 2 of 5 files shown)


adamdriscoll:
Not seeing this test failure locally. I am seeing other failures that aren't listed in AppVeyor that seem unrelated to my PR.

src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
```diff
@@ -21,5 +21,7 @@ public class AttachRequestArguments
         public string ProcessId { get; set; }
 
         public int RunspaceId { get; set; }
+
+        public string LocalRunspaceId { get; set; }
```
rkeithhill: Just asking but ... do we really need two different fields for runspace id?

rkeithhill: BTW if it would be more convenient if `RunspaceId` was a string, we change the type I think.  We'd just update the VSCode debug attach code to stringify the runspace id number.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -435,7 +437,28 @@ protected void Stop()
...
                 // InitializedEvent will be sent as soon as the RunspaceChanged
                 // event gets fired with the attached runspace.
-                int runspaceId = attachParams.RunspaceId > 0 ? attachParams.RunspaceId : 1;
+
+                int runspaceId = 1;
+                if (string.IsNullOrEmpty(attachParams.LocalRunspaceId))
+                {
+                    runspaceId = attachParams.RunspaceId > 0 ? attachParams.RunspaceId : 1;
+                }
+                else if (int.TryParse(attachParams.LocalRunspaceId, out int localRunspaceId))
```
rkeithhill: Seems like we could `int.TryParse()` on a string typed RunspaceId.  In the normal "attach" case we'd just set it to "1".  Your new debug config would set it to whatever the user picks from the runspace id pick list.

src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
adamdriscoll: OK. I can make that change. That's pretty much the reason I added another field was that I didn't wanna mess with RunspaceId's type but I'm happy to adjust that to eliminate the new property.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
adamdriscoll: Sweet. Works for me. I will update these PRs sometime this weekend.

src/PowerShellEditorServices.Protocol/DebugAdapter/AttachRequest.cs
adamdriscoll: I guess I was curious how this affects existing launch.json files. Is there something we need to do to ensure that they get upgraded in some way or will a number just work?

adamdriscoll: I just tried this. Seems to still work. It complains in the launch.json that the number should be a string but the attaching still works.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -417,16 +417,18 @@ protected void Stop()
                     return;
                 }
 
-                await _editorSession.PowerShellContext.ExecuteScriptStringAsync(
+                if (attachParams.ProcessId != "current") {
```
TylerLeonhardt: Shouldn't need this if because the outer `if` tries to parse the ProcessId as an integer (L411):

```csharp
if (processIdIsSet && int.TryParse(attachParams.ProcessId, out int processId) && (processId > 0))
```

adamdriscoll: Good catch. Updated the PR.

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -469,7 +471,24 @@ protected void Stop()
             // will block the debug adapter initialization process.  The
             // InitializedEvent will be sent as soon as the RunspaceChanged
             // event gets fired with the attached runspace.
-            int runspaceId = attachParams.RunspaceId > 0 ? attachParams.RunspaceId : 1;
+
+            int runspaceId = 1;
+            if (int.TryParse(attachParams.RunspaceId, out runspaceId))
+            {
+                runspaceId = runspaceId > 0 ? runspaceId : 1;
```
TylerLeonhardt: I would maybe factor this into the `if`

```csharp
if (!int.TryParse(attachParams.RunspaceId, out runspaceId) || runspaceId <= 0)
{
    Logger.Write(
        LogLevel.Error,
        $"Attach request failed, '{attachParams.RunspaceId}' is an invalid value for the processId.");

    await requestContext.SendErrorAsync(
        "A positive integer must be specified for the RunspaceId field.");

    return;
}
```

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
```diff
@@ -469,7 +471,24 @@ protected void Stop()
             // will block the debug adapter initialization process.  The
             // InitializedEvent will be sent as soon as the RunspaceChanged
             // event gets fired with the attached runspace.
-            int runspaceId = attachParams.RunspaceId > 0 ? attachParams.RunspaceId : 1;
+
+            int runspaceId = 1;
```
TylerLeonhardt: nit: we use var when it's obvious

```suggestion
            var runspaceId = 1;
```

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1218,6 +1221,36 @@ public static string GetDecoratedSymbolName(SymbolReference symbolReference)
...
+            RequestContext<GetRunspaceResponse[]> requestContext)
+        {
+            var runspaceResponses = new List<GetRunspaceResponse>();
+
+            if (this.editorSession.PowerShellContext.LocalPowerShellVersion.Version.Major >= 5)
+            {
+                var psCommand = new PSCommand();
+                psCommand.AddCommand("Get-Runspace");
+
+                var runspaces = await editorSession.PowerShellContext.ExecuteCommandAsync<PSObject>(psCommand);
```
TylerLeonhardt: can you replace `var` with the type since this is a tad more complicated than `new Blah()`

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1218,6 +1221,36 @@ public static string GetDecoratedSymbolName(SymbolReference symbolReference)
             await requestContext.SendResultAsync(result);
         }
 
+        protected async Task HandleGetRunspaceRequestAsync(
+            object noParams,
+            RequestContext<GetRunspaceResponse[]> requestContext)
+        {
+            var runspaceResponses = new List<GetRunspaceResponse>();
+
+            if (this.editorSession.PowerShellContext.LocalPowerShellVersion.Version.Major >= 5)
```
TylerLeonhardt: Was `Get-Runspace` added in 5?

src/PowerShellEditorServices.Protocol/Server/DebugAdapter.cs
adamdriscoll: 👍

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: Also, instead of PSObject you should be able to do `System.Management.Automation.Runspaces.Runspace` and then you don't have to use dynamic below.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
adamdriscoll: That's a really good question. I just took that check from the process picker but I'm not sure when Debug-Runspace\Get-Runspace was added. I can try and find out.

adamdriscoll: image

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
adamdriscoll: 👍

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
TylerLeonhardt: Well looks like that answers our question :)

TylerLeonhardt:
@adamdriscoll if you couldn't tell already, that test is flaky... I've been meaning to fix it, but I haven't had the time. I will address it as soon as I can.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1246,6 +1256,12 @@ public static string GetDecoratedSymbolName(SymbolReference symbolReference)
                             });
                     }
                 }
+
+                if (processId != null && processId.ToString() != "current") {
```
TylerLeonhardt: Since you do this comparison a couple times, maybe do a:
```csharp
bool isCurrentProcess = processId != null && processId.ToString() != "current";
```

and use that.

adamdriscoll: Works for me. Done.

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1222,17 +1222,27 @@ public static string GetDecoratedSymbolName(SymbolReference symbolReference)
         }
 
         protected async Task HandleGetRunspaceRequestAsync(
-            object noParams,
+            object processId,
```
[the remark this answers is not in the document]
adamdriscoll: You're right. Changed to a string.

PR #895, opened 2019-03-27
Add warning to parameter validation set

Benny1007:
Add "Warning" to LogLevel parameter validation set of Start-EditorServices.ps1

### DIFF
--- a/module/PowerShellEditorServices/Start-EditorServices.ps1
+++ b/module/PowerShellEditorServices/Start-EditorServices.ps1
@@ -46,7 +46,7 @@ param(
     [ValidateNotNullOrEmpty()]
     $LogPath,
 
-    [ValidateSet("Diagnostic", "Normal", "Verbose", "Error")]
+    [ValidateSet("Diagnostic", "Normal", "Verbose", "Warning", "Error")]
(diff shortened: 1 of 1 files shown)


module/PowerShellEditorServices/Start-EditorServices.ps1
```diff
@@ -46,7 +46,7 @@ param(
     [ValidateNotNullOrEmpty()]
     $LogPath,
 
-    [ValidateSet("Diagnostic", "Normal", "Verbose", "Error")]
+    [ValidateSet("Diagnostic", "Normal", "Verbose", "Warning", "Error")]
```
rkeithhill: One minor suggestion - could you change the order of `Normal` and `Verbose` since the log level order is actually: Diagnostic, Verbose, Normal, Warning, Error.  I don't think it really matters that much but might as well have the correct order here since it is already close to the correct order.  :-)

Benny1007: sure, done.

rkeithhill: Thanks!

PR #902, opened 2019-04-01
Improve path auto-completion

rkeithhill:
Need to determine if there are other quote scenarios where we could improve auto-completion.

Fixes https://github.com/PowerShell/vscode-powershell/issues/1398

### DIFF
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
@@ -58,6 +58,12 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         Folder = 19
     }
 
+    public enum InsertTextFormat
+    {
+        PlainText = 1,
+        Snippet = 2,
+    }
+
     [DebuggerDisplay("NewText = {NewText}, Range = {Range.Start.Line}:{Range.Start.Character} - {Range.End.Line}:{Range.End.Character}")]
     public class TextEdit
     {
@@ -69,6 +75,11 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
     [DebuggerDisplay("Kind = {Kind.ToString()}, Label = {Label}, Detail = {Detail}")]
     public class CompletionItem
     {
+        public CompletionItem()
+        {
+            this.InsertTextFormat = InsertTextFormat.PlainText;
+        }
+
         public string Label { get; set; }
 
         public CompletionItemKind? Kind { get; set; }
@@ -85,6 +96,8 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
         public string FilterText { get; set; }
 
         public string InsertText { get; set; }
+
+        public InsertTextFormat InsertTextFormat { get; set; }
 
         public Range Range { get; set; }
 
--- a/src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
+++ b/src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
@@ -75,11 +75,6 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
     [DebuggerDisplay("Kind = {Kind.ToString()}, Label = {Label}, Detail = {Detail}")]
     public class CompletionItem
     {
-        public CompletionItem()
-        {
-            this.InsertTextFormat = InsertTextFormat.PlainText;
-        }
-
         public string Label { get; set; }
 
         public CompletionItemKind? Kind { get; set; }
@@ -97,7 +92,7 @@ namespace Microsoft.PowerShell.EditorServices.Protocol.LanguageServer
 
         public string InsertText { get; set; }
 
-        public InsertTextFormat InsertTextFormat { get; set; }
+        public InsertTextFormat InsertTextFormat { get; set; } = InsertTextFormat.PlainText;
 
         public Range Range { get; set; }
 
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -1915,6 +1915,8 @@ function __Expand-Alias {
         {
             string detailString = null;
             string documentationString = null;
+            string completionText = completionDetails.CompletionText;
+            InsertTextFormat insertTextFormat = InsertTextFormat.PlainText;
 
             if ((completionDetails.CompletionType == CompletionType.Variable) ||
                 (completionDetails.CompletionType == CompletionType.ParameterName))
@@ -1956,6 +1958,14 @@ function __Expand-Alias {
                     }
                 }
             }
+            else if (((completionDetails.CompletionType == CompletionType.File) ||
+                      (completionDetails.CompletionType == CompletionType.Folder)) &&
+                     (completionText.EndsWith("\"") || completionText.EndsWith("'")))
+            {
+                int len = completionDetails.CompletionText.Length;
+                completionText = completionDetails.CompletionText.Insert(len - 1, "$0");
+                insertTextFormat = InsertTextFormat.Snippet;
+            }
 
             // Force the client to maintain the sort order in which the
             // original completion results were returned. We just need to
@@ -1966,7 +1976,8 @@ function __Expand-Alias {
 
             return new CompletionItem
             {
-                InsertText = completionDetails.CompletionText,
+                InsertText = completionText,
+                InsertTextFormat = insertTextFormat,
                 Label = completionDetails.ListItemText,
                 Kind = MapCompletionKind(completionDetails.CompletionType),
                 Detail = detailString,
@@ -1975,7 +1986,7 @@ function __Expand-Alias {
                 FilterText = completionDetails.CompletionText,
                 TextEdit = new TextEdit
                 {
-                    NewText = completionDetails.CompletionText,
+                    NewText = completionText,
                     Range = new Range
                     {
                         Start = new Position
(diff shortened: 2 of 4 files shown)


src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
```diff
@@ -69,6 +75,11 @@ public class TextEdit
     [DebuggerDisplay("Kind = {Kind.ToString()}, Label = {Label}, Detail = {Detail}")]
     public class CompletionItem
     {
+        public CompletionItem()
+        {
+            this.InsertTextFormat = InsertTextFormat.PlainText;
```
TylerLeonhardt: Maybe just use a default parameter value instead of putting it in the constructor?

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
```diff
@@ -58,6 +58,12 @@ public enum CompletionItemKind
         Folder = 19
     }
 
+    public enum InsertTextFormat
+    {
+        PlainText = 1,
```
TylerLeonhardt: Does this have to be 1 based?

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1956,6 +1958,14 @@ private static CompletionItemKind MapCompletionKind(CompletionType completionTyp
                     }
                 }
             }
+            else if (((completionDetails.CompletionType == CompletionType.File) ||
+                      (completionDetails.CompletionType == CompletionType.Folder)) &&
+                     (completionText.EndsWith("\"") || completionText.EndsWith("'")))
+            {
+                int len = completionDetails.CompletionText.Length;
+                completionText = completionDetails.CompletionText.Insert(len - 1, "$0");
```
TylerLeonhardt: can you explain what this insert is doing and what the `$0` is used for?

(and maybe add a comment :) )

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
rkeithhill: Default parameter value where?  I could just nuke this constructor and every where we use a Completion (just two places), you have to set the InsertTextFormat property.

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
rkeithhill: Well, that is what LSP expects and in general, we've also mapped 1:1 between our enum values and what LSP expects.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.languageserver.protocol.inserttextformat?view=visualstudiosdk-2017
https://developercommunity.visualstudio.com/content/problem/258457/lsp-hangs-on-completions-if-inserttextformat-is-no.html

src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
rkeithhill: Sure!

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
```diff
@@ -86,6 +97,8 @@ public class CompletionItem
 
         public string InsertText { get; set; }
 
+        public InsertTextFormat InsertTextFormat { get; set; }
```
TylerLeonhardt: Here:
```suggestion
        public InsertTextFormat InsertTextFormat { get; set; } = InsertTextFormat.PlainText;
```

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
TylerLeonhardt: See below!

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
TylerLeonhardt: Although I thought enums defaulted to the first item in the enum. I could be wrong.

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
rkeithhill: Ah, default *property* value. Heh, I started with that.  :-)  But then it seemed odd to me that the init appeared so far down the class def.  That is one thing I don't like about auto-property initialization.  I guess I'm just used to being able to see class member initialization towards the beginning of a class def.  But I'm happy to make that change if you'd prefer that.

src/PowerShellEditorServices.Protocol/LanguageServer/Completion.cs
rkeithhill: It is 0, so it is unfortunate that this LSP starts at 1.

TylerLeonhardt: That's so sad... oh well. I would prefer the property default value instead of the constructor, personally.

rkeithhill:
This PR is ready to go.

test/PowerShellEditorServices.Test.Host/LanguageServerTests.cs
```diff
@@ -257,6 +258,97 @@ public async Task CompletesDetailOnCommandSuggestion()
...
+            InsertTextFormat insertTextFormat;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                expectedPathSnippet = @"'.\TestFiles\Folder With Spaces$0'";
+                insertTextFormat = InsertTextFormat.Snippet;
+            }
+            else
+            {
+                expectedPathSnippet = @"./TestFiles/Folder\ With\ Spaces";
```
TylerLeonhardt: Is this valid? On linux I get:

```
Get-ChildItem ./Folder\ With\ Spaces/  
Get-ChildItem : A positional parameter cannot be found that accepts argument 'Spaces/'.
At line:1 char:1
+ Get-ChildItem ./Folder\ With\ Spaces/
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidArgument: (:) [Get-ChildItem], ParameterBindingException
+ FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand
```

test/PowerShellEditorServices.Test.Host/TestFiles/CompleteFunctionName.ps1
```diff
@@ -4,4 +4,6 @@ function My-Function
 $Cons
 My-
 Get-Proc
-$HKC
\ No newline at end of file
+$HKC
+Get-ChildItem ./TestFiles/Folder
+Get-ChildItem ./TestFiles/CompleteF
```
TylerLeonhardt: Do you want to add a test to make sure that it continues to complete after it resolves? In otherwords:

```pwsh
Get-ChildItem ./TestFiles/Folder
```
goes to:
```pwsh
Get-ChildItem './TestFiles/Folder With Spaces'
```
but in theory it should keep going:
```pwsh
Get-ChildItem './TestFiles/Folder With Spaces/AnotherFolder'
```

test/PowerShellEditorServices.Test.Host/LanguageServerTests.cs
rkeithhill: Good catch, I must have been in Bash when I got that auto-completion of that folder.  I'll fix it.

test/PowerShellEditorServices.Test.Host/TestFiles/CompleteFunctionName.ps1
rkeithhill: It won't go to that unless you type a `/` char.  The main point of this point is to put the cursor at the right spot so you only have to type a `/` to get further auto-completion.  BTW this behaves differently than the command-line completion where it automatically adds the `/` char.  I "think" we don't get that in PSES because we use the AST for completions but not 100% sure on that.

TylerLeonhardt: interesting! Thanks for sharing. Might be still worth having a test for this:

```
Get-ChildItem './TestFiles/Folder With Spaces/AnotherFolder/'
```

and that completions should still show up after the last `/`?

test/PowerShellEditorServices.Test.Host/LanguageServerTests.cs
rkeithhill: Fixed

test/PowerShellEditorServices.Test.Host/TestFiles/CompleteFunctionName.ps1
rkeithhill: The way the tests work, you setup a partial completion in a file, then send the completion request for that spot in the file.  I could set up another line in the file to check but I don't how that would add value over the two lines I've added for testing folder and file completion.

TylerLeonhardt: Yeah what we have is fine.

test/PowerShellEditorServices.Test.Host/LanguageServerTests.cs
```diff
@@ -257,6 +258,97 @@ public async Task CompletesDetailOnCommandSuggestion()
...
+            {
+                expectedPathSnippet = @".\TestFiles\CompleteFunctionName.ps1";
+            }
+            else
+            {
+                expectedPathSnippet = "./TestFiles/CompleteFunctionName.ps1";
+            }
+
+            // Change dir to root of this test project's folder
+            await this.SetLocationForServerTest(this.TestRootDir);
```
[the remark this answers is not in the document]
rkeithhill: All of the completion tests in this file have the same Codacy issue.  I think it would be better to fix these all together in a separate PR.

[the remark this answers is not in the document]
TylerLeonhardt: sounds good!

TylerLeonhardt:
@rkeithhill I think @rjmholt and I have decided that we, personally, will not spend the time to backport this, but if you'd like to do it, we would accept it!

rkeithhill:
I'll backport it.  I'd really like to have this in the next update.

PR #917, opened 2019-04-10
Backport: Fix issue with reference code lens not working with UNC paths (#908)

rkeithhill:
* Fix issue with reference code lens not working with UNC paths

* Refactor to avoid extra Insert ops. Simplifies macOS/Linux code path.

* Still need to unescape / on macOs & Linux

* Replace compile-time check for .NET Core with runtime check


rkeithhill:
After this is merged, can we point a few folks that were having issues with this to the VSIX that gets built? I'd like to verify it fixes all the failure cases we've seen.  There are a few out there, like the script parse NRE on extent that I'm not sure if this fixes.

PR #942, opened 2019-04-18
dependabot and Discord badges

TylerLeonhardt:
because we need more badges obviously.


rkeithhill:
You can't have enough flair ... or badges.  :-)

PR #943, opened 2019-04-19
Update parse log to handle Read/WrteMessageAsync

rkeithhill:
Add progress percentage to parse command.

Add textDocument/completion to ValidaeSet of MesssageNameparam 
on Get-PsesRpcMessageResponseTime command.


rkeithhill:
Not sure what is up with the buid failures - must be that the transition to Azure-Pipelines isn't quite done yet?  Anyway, this is just a change to the PSESAnalyzer tool so it shouldn't impact the build at all.

PR #971, opened 2019-05-28
Add list of frontends to README

rjmholt:
Resolves https://github.com/PowerShell/PowerShellEditorServices/issues/969

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ experience in almost any editor or integrated development environment (IDE).
 
 ## PowerShell Language Server Protocol clients using PowerShell Editor Services
 
-The functionality in PSES is already available in the following editor extensions:
+The functionality in PowerShell Editor Services is already available in the following editor extensions:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,14 @@ PowerShell Editor Services is a PowerShell module that provides common
 functionality needed to enable a consistent and robust PowerShell development
 experience in almost any editor or integrated development environment (IDE).
 
+## Front-ends using PSES
```
rkeithhill: "Front-ends" is probably more recognizable to folks but technically these are language service client apps, right?

rjmholt: Yeah "LSP clients" is maybe better?

rkeithhill: That's a more "correct" description IMO. Whether that is meaningful to the average user who looks at this ... I dunno.

rjmholt: Well I'm hoiping that users who get to the PSES README probably understand it somewhat

rjmholt:
I'll merge this in when the LSP page is updated, closing the issue this resolves

PR #974, opened 2019-06-07
Defend against crash when no PSSA

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/2007.

Formatting doesn't check to see if the AnalysisService is null before a method invocation.

### DIFF
--- a/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
+++ b/src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
@@ -1380,6 +1380,12 @@ function __Expand-Alias {
             DocumentFormattingParams formattingParams,
             RequestContext<TextEdit[]> requestContext)
         {
+            if (this.editorSession.AnalysisService == null)
+            {
+                await requestContext.SendErrorAsync("Script analysis is not enabled in this session");
+                return;
+            }
+
             var result = await FormatAsync(
                 formattingParams.TextDocument.Uri,
                 formattingParams.options,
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices.Protocol/Server/LanguageServer.cs
```diff
@@ -1380,6 +1380,12 @@ private bool IsQueryMatch(string query, string symbolName)
             DocumentFormattingParams formattingParams,
             RequestContext<TextEdit[]> requestContext)
         {
+            if (this.editorSession.AnalysisService == null)
+            {
+                await requestContext.SendErrorAsync("Script analysis is not enabled in this session");
```
SeeminglyScience: Does this wait for the error to be dismissed?  I can't remember if the default for `waitForCompletion` is `true` or `false`.

Also does this appear every formatting request or only once per session?  If the former, do we want to make it once per session so that `formatOnSave` isn't obnoxious or is this uncommon enough that we shouldn't worry about it?

rjmholt: As far as I know, this just sends an LSP error message back and the client decides how to handle it. But I don't think it blocks beyond sending the message

SeeminglyScience: Right right I'm thinking of the custom show error message request we do, nevermind 🙂

rjmholt: This might make VSCode show an error every time though...

rkeithhill: That should be easy to test.

PR #982, opened 2019-06-22
Update README.md

kiennq:
Mentioned [lsp-powershell](https://github.com/kiennq/lsp-powershell), which is also an Emacs [lsp-mode](https://github.com/emacs-lsp/lsp-mode) client for PowerShell that leverage `PowerShellEditorServices`


kiennq:
> I should ask, is `lsp-language` what all the Emacs LSP plugins use as a naming convention?

Most of them, I think so. You can get an exhaustive table [here](https://github.com/emacs-lsp/lsp-mode#supported-languages)

PR #1021, opened 2019-09-03
Add AutoCorrectAliases setting (PR to be made in VS-Code repo as well) to add support for optionally correcting aliases as well (added in PSSA 1.18.2)

bergmeister:


bergmeister:
@poshchan please retry ci

PoshChan:
@bergmeister, you are not authorized to request a rebuild

bergmeister:
@SteveL-MSFT Although I am in the file of authorized poshchan users [here](https://github.com/PowerShell/PowerShellEditorServices/blob/master/.poshchan/settings.json), I do not seem to be able to request a retry (see above)

src/PowerShellEditorServices.Protocol/Server/LanguageServerSettings.cs
```diff
@@ -292,7 +284,25 @@ private Hashtable GetCustomPSSASettingsHashtable(int tabSize, bool insertSpaces)
                     {"PSUseCorrectCasing", new Hashtable {
                         {"Enable", UseCorrectCasing}
                     }},
-                }}
+                };
+            if (AutoCorrectAliases)
+            {
+                ruleConfigurations.Add("PSAvoidUsingCmdletAliases", new Hashtable());
```
rjmholt: The current pattern seems to always include the rule entry and set `Enable` to the boolean given. Is there a reason that shouldn't be done here?

Also, does this need to set `Enable = true` in the hashtable body?

bergmeister: The `PSAvoidUsingCmdletAliases` is a script analysis rule that inherits only from `IScriptRule` (see [here](https://github.com/PowerShell/PSScriptAnalyzer/blob/master/Rules/AvoidAlias.cs)). Formatting rules typically inherit from [ConfigurableRule](https://github.com/PowerShell/PSScriptAnalyzer/blob/master/Engine/Generic/ConfigurableRule.cs), which is what makes them have the `Enable` property. This is just the API of this specific rule, therefore although it would be nice to have them all consistent, I don't think we could even change that in PSSA without breaking people that use the `PSAvoidUsingCmdletAliases` for code analysis in their setting files.

PR #1114, opened 2019-11-24
Fix Import-EditorCommand bug

sk82jack:
It appears that PR #1056 involved renaming the `extensionService` field to `_extensionService` but this wasn't changed in the `Import-EditorCommand` function causing an error when trying to use it.

Following a discussion with @TylerLeonhardt on the PowerShell Discord server I've submitted this PR to fix the error by updating the field name in `Import-EditorCommand`.

I didn't see any tests or anything that needed to be changed but let me know if you need me to do anything else in order for this to be accepted 🙂

### DIFF
--- a/module/PowerShellEditorServices/Commands/Public/Import-EditorCommand.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/Import-EditorCommand.ps1
@@ -57,14 +57,11 @@ function Import-EditorCommand {
                 return $moduleInfo.ExportedFunctions.Values
             }
         }
-        $flags = [Reflection.BindingFlags]'Instance, NonPublic'
-        $extensionService = $psEditor.GetType().
-                                      GetField('_extensionService', $flags).
-                                      GetValue($psEditor)
+        $editorCommands = @{}
 
-        $editorCommands = $extensionService.GetType().
-                                            GetField('editorCommands', $flags).
-                                            GetValue($extensionService)
+        foreach ($existingCommand in $psEditor.GetCommands()) {
+            $editorCommands[$existingCommand.Name] = $existingCommand
+        }
     }
     process {
         switch ($PSCmdlet.ParameterSetName) {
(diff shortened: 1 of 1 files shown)


module/PowerShellEditorServices/Commands/Public/Import-EditorCommand.ps1
```diff
@@ -59,7 +59,7 @@ function Import-EditorCommand {
         }
         $flags = [Reflection.BindingFlags]'Instance, NonPublic'
         $extensionService = $psEditor.GetType().
-                                      GetField('extensionService', $flags).
+                                      GetField('_extensionService', $flags).
                                       GetValue($psEditor)
 
         $editorCommands = $extensionService.GetType().
```
SeeminglyScience: Actually you can replace all of these with `$editorCommands = $psEditor.GetCommands()`.  That API didn't exist at the time, but now that it does that should definitely be used instead.

sk82jack: Ah yeah that looks a lot nicer. I can change it to that but I'm not sure about the correct procedure in terms of not cluttering up the commit history - should I do an interactive rebase to clean up or leave it up to the maintainer to squash & merge?

SeeminglyScience: > leave it up to the maintainer to squash & merge?

That one :)

Also taking a closer look you'll probably want to do something like:

```powershell
$editorCommands = [System.Collections.Generic.Dictionary[string, Microsoft.PowerShell.EditorServices.Services.PowerShellContext.EditorCommand]]::new(
    [System.StringComparer]::OrdinalIgnoreCase)

foreach ($existingCommand in $psEditor.GetCommands()) {
    $editorCommands.Add($existingCommand.Name, $existingCommand)
}
```

sk82jack: Thanks for the pointers - will get that implemented and confirm it all still works

Out of interest why do you suggest a strongly typed dictionary over a regular hashtable? Performance or just clearer intent/more control or something?

SeeminglyScience: Actually I was just matching the type of the field, `hashtable` will probably work just fine and be a bit more clean.  Good call :)

sk82jack: Awesome. Just pushed new version 🙂

PR #1141, opened 2020-01-02
Make startup banner simpler

rjmholt:
Resolves https://github.com/PowerShell/vscode-powershell/issues/2370


bergmeister:
Thank god for that, I hated the banner, why do we need a banner at all?

rjmholt:
> why do we need a banner at all?

A lot of people don't realise there's a difference between the integrated console and normal PowerShell. The banner helps them see where they are and helps us see when bug reports belong elsewhere.

Also it seemed like a fun thing that other CLI tools do, especially ones that put you in some kind of an environment like the Azure Functions CLI.

But hey, some people didn't like it (some being more vitriolic than others) and we're getting rid of it in a single cycle. That's how the preview is supposed to work.

And if you want it back, you can put it in your profile.

We're keeping the little banner (with an opt-out) just so that it's clear which console you're in by default.

bergmeister:
Yh, I agree, that's the point of previews: experiment and revert mistakes quickly.
One can differentiate the terminals in the dropdown but once cannot even read it. Maybe shortening it something like `PS Integrated terminal` or PSIC (although most people will be rather confused by that acronym)
image

TylerLeonhardt:
@bergmeister I think that'd be a change in vscode-powershell but I like `PS Integrated Console`. Helps us stand out from just `pwsh` or `powershell`

PR #1024, opened 2019-09-10
Backport #1023 and #1021

rjmholt:
Backport #1023 and #1021

PR #1142, opened 2020-01-03
1.x: Fix wiring of WhitespaceInsideBrace and WhitespaceAroundPipe that got broken in #1024

bergmeister:
Fixes https://github.com/PowerShell/vscode-powershell/issues/2375
More settings got broken in this bad cherry pick, I can only assume that there was a merge conflict that was resolved incorrectly.
To make it simpler in the future, I copied the method from master (hence the whitespace diff)
cc @TylerLeonhardt


src/PowerShellEditorServices.Protocol/Server/LanguageServerSettings.cs
```diff
@@ -249,58 +249,62 @@ public CodeFormattingSettings(CodeFormattingSettings codeFormattingSettings)
...
-                    {"CheckOperator", WhitespaceAroundOperator},
-                    {"CheckSeparator", WhitespaceAfterSeparator}
+                { "PSUseConsistentWhitespace", new Hashtable {
+                    { "Enable", true },
+                    { "CheckOpenBrace", WhitespaceBeforeOpenBrace },
+                    { "CheckOpenParen", WhitespaceBeforeOpenParen },
+                    { "CheckOperator", WhitespaceAroundOperator },
+                    { "CheckSeparator", WhitespaceAfterSeparator },
+                    { "CheckInnerBrace", WhitespaceInsideBrace },
+                    { "CheckPipe", WhitespaceAroundPipe },
```
bergmeister: This line and the line above is the essential fix

src/PowerShellEditorServices.Protocol/Server/LanguageServerSettings.cs
```diff
@@ -249,58 +249,62 @@ public CodeFormattingSettings(CodeFormattingSettings codeFormattingSettings)
...
-            return new Hashtable
+            return new Hashtable()
             {
-                {"IncludeRules", new string[] {
-                     "PSPlaceCloseBrace",
-                     "PSPlaceOpenBrace",
-                     "PSUseConsistentWhitespace",
-                     "PSUseConsistentIndentation",
-                     "PSAlignAssignmentStatement",
-                     "PSAvoidUsingCmdletAliases",
```
bergmeister: This entry is actually not necessary and was removed in PR #1056, hence why this changed is brought back as well as the essential required thing is `ruleConfigurations.Add("PSAvoidUsingCmdletAliases", new Hashtable());` a few lines above

bergmeister:
@rjmholt  I could but the whole point of bringing the formatting changes from master into the 1.x is branch that cherry-picking would work again without silly white-space merge conflicts that could lead to cause mistakes like this. Do you still not agree?

rjmholt:
> cherry-picking would work again without silly white-space merge conflicts that could lead to cause mistakes like this

If that's the original cause of the issue then that's a fair point. I'll let @TylerLeonhardt decide there.

There shouldn't be any more cherry picks though. This PR is probably the only change we'll make before a final release of the 1.x extension. Later this month the preview branch will become stable and preview/stable will behave like they were originally intended to.

TylerLeonhardt:
Yeah I'm fine with that. I just reviewed the whole thing line-by-line. Looks good.

TylerLeonhardt:
@rjmholt can you review as well though?

bergmeister:
Diff becomes much easier when using this option:
image

PR #1151, opened 2020-01-10
Probe .NET Framework dir for dependencies

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/2393.

Allows loading of netstandard.dll on downlevel .NET Framework versions.

We falsely assumed that getting PowerShell to load an assembly module would automatically resolve dependencies in the same directory, but this is actually a property of PowerShell 6+ rather than .NET.

Now we look in the .NET Framework dir on an assembly resolve event before looking in the shared dependencies directory.


rjmholt:
Tested on WinSrv2016 with .NET 4.6.2 and 4.7

PR #1175, opened 2020-02-01
Check _analysisService is null

mattn:
_analysisService is possibly null


rjmholt:
Thanks for opening this. I'm actually about to make changes to this file to account for the root cause of this, so I don't want to drag you into a PR in the meantime. Essentially the way we instantiate services has changed drastically and the model in which the AnalysisService could be null has changed a fair bit.

So rather than drag you through a PR, it might be best for me to go ahead and make that change. Also, let me know if you disagree or if I'm taking too long to deliver on that. Cheers

mattn:
Okay, feel free to close this. Thanks.

rjmholt:
Closing in favour of https://github.com/PowerShell/PowerShellEditorServices/pull/1179

PR #1199, opened 2020-02-18
Introduce Get-EditorServicesParserAst to expand on ExpandAlias

romero126:
Get-EditorServicesParserAst will allow for ExpandAlias or other ast features to look at comment blocks or Functions defined.

This also addresses an issue with running the Module in Constrained Language Mode.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/Commands/GetEditorServicesParserAst.cs
+++ b/src/PowerShellEditorServices.Hosting/Commands/GetEditorServicesParserAst.cs
@@ -1,0 +1,66 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.ObjectModel;
+using System.IO;
+using System.Linq;
+using System.Management.Automation;
+using Microsoft.PowerShell.Commands;
+
+namespace Microsoft.PowerShell.EditorServices.Commands
+{
+    
+    /// <summary>
+    /// The Get-EditorServicesParserAst command will parse and expand out data parsed by ast.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "EditorServicesParserAst")]
+    public sealed class GetEditorServicesParserAst : PSCmdlet
+    {
+
+        /// <summary>
+        /// The Scriptblock or block of code that gets parsed by ast.
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        public string ScriptBlock { get; set; }
+
+        /// <summary>
+        /// Specify a specific command type
+        /// [System.Management.Automation.CommandTypes]
+        /// </summary>
+        [Parameter(Mandatory = true)]        
+        public CommandTypes CommandType { get; set; }
+
+        /// <summary>
+        /// Specify a specific token type
+        /// [System.Management.Automation.PSTokenType]
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        public PSTokenType PSTokenType { get; set; }
+
+        protected override void EndProcessing()
+        {
+            var errors = new Collection<PSParseError>();
+
+            var tokens =
+                System.Management.Automation.PSParser.Tokenize(ScriptBlock, out errors)
+                .Where(token => token.Type == this.PSTokenType)
+                .OrderBy(token => token.Content);
+
+            foreach (PSToken token in tokens)
+            {
+                if (PSTokenType == PSTokenType.Command)
+                {
+                    var result = SessionState.InvokeCommand.GetCommand(token.Content, CommandType);
+                    WriteObject(result);
+                }
+                else {
+                    WriteObject(token);
+                }
+
+            }
+        }
+    }
+}
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Handlers/ExpandAliasHandler.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Handlers/ExpandAliasHandler.cs
@@ -40,37 +40,14 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
         public async Task<ExpandAliasResult> Handle(ExpandAliasParams request, CancellationToken cancellationToken)
         {
-            const string script = @"
-function __Expand-Alias {
 
-    param($targetScript)
-
-    [ref]$errors=$null
-
-    $tokens = [System.Management.Automation.PsParser]::Tokenize($targetScript, $errors).Where({$_.type -eq 'command'}) |
-                    Sort-Object Start -Descending
-
-    foreach ($token in  $tokens) {
-        $definition=(Get-Command ('`'+$token.Content) -CommandType Alias -ErrorAction SilentlyContinue).Definition
-
-        if($definition) {
-            $lhs=$targetScript.Substring(0, $token.Start)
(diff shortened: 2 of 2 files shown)


src/PowerShellEditorServices.Hosting/Commands/GetEditorServicesParserAst.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Specify a specific token type
+        /// [System.Management.Automation.PSTokenType]
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        public PSTokenType PSTokenType { get; set; }
+
+        protected override void EndProcessing()
+        {
+            var errors = new Collection<PSParseError>();
```
TylerLeonhardt: Codacy Issue found: [Remove this useless assignment to local variable 'errors'.](https://app.codacy.com/manual/TylerLeonhardt/PowerShellEditorServices/pullRequest?prid=5026608)

TylerLeonhardt:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- src/PowerShellEditorServices.Hosting/Commands/GetEditorServicesParserAst.cs  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/TylerLeonhardt/PowerShellEditorServices/pullRequest?prid=5026608&bid=16565300)

src/PowerShellEditorServices.Hosting/Commands/GetEditorServicesParserAst.cs
```diff
@@ -0,0 +1,66 @@
...
+            var tokens =
+                System.Management.Automation.PSParser.Tokenize(ScriptBlock, out errors)
+                .Where(token => token.Type == this.PSTokenType)
+                .OrderBy(token => token.Content);
+
+            foreach (PSToken token in tokens)
+            {
+                if (PSTokenType == PSTokenType.Command)
+                {
+                    var result = SessionState.InvokeCommand.GetCommand(token.Content, CommandType);
```
TylerLeonhardt: You've done a great job of converting the code from PowerShell to C# that I'm wondering if you could put this code in the Handler below instead of in its own cmdlet... then the only thing you'd need to do in script is the call to `Get-Command` (by using `_powerShellContextService.ExecuteCommandAsync`) and that would be ConstrainedLanguage mode compliant.

Also, `Get-Command` takes in an array for `-Name` so you should be able to get the names in a foreach, throw them into an array, and then pass that into `Get-Command` so we only invoke PowerShell once.

rjmholt: Basically rather than implementing the cmdlet, you can do something like this:

```csharp
ScriptBlockAst scriptAst = Parser.ParseInput(request.Text, out Token[] _, out ParseError[] _);

var commandsToExpand = new List<string>();
foreach (Ast foundAst in scriptAst.FindAll(ast => ast is CommandAst, true))
{
    if (!(foundAst is CommandAst commandAst))
    {
        continue;
    }

    string commandName = commandAst.GetCommandName();

    if (commandName != null)
    {
        commandsToExpand.Add(commandName);
    }
}

var psCommand = new PSCommand()
    .AddCommand("Get-Command")
    .AddParameter("Name", commandsToExpand)
    .AddParameter("ErrorAction", "Ignore");

IEnumerable<string> expandedCommands = _powerShellContextService.InvokeCommandAsync<string>(psCommand);
```

rjmholt: Posted this to document the thought, but it's non-trivial at this point.

You now need to be able to insert the expanded command names back into the original AST while not modifying anything else.

That will require some amount of custom logic to put the script back together from the AST, and there's not currently a nice API for that beyond writing a new AST visitor...

We could try this on tokens instead, but I'm not sure that's the best idea.

romero126: Originally I had thought we can do much more with the tokenizer in order to remove repetitious code that will come up. Like Expand Alias, Expand Command.. etc.. However I don't really see much more of a use case other than those two.

So moving forward.. 
We can do something similar to this in the pipeline. It looks like we just need to create a pipeline object and chain commands to it. 

`get-command -Name gci | ? { $_ -is [System.Management.Automation.AliasInfo] } | % ResolvedCommand | % { $_.ModuleName + "\" + $_.Name }`

rjmholt: The problem with using the pipeline is that the more you use the pipeline the longer you exclude other services from using it. So basically you only want to use the pipeline for things you absolutely need its context for, which in this case is alias resolution.

Beyond that, we want to do as much as we can in off-thread C# code.

The problem is that the old function rebuilt the whole script one command at a time by rebuilding the string for each command. The new cmdlet doesn't do that and my code above doesn't have an implementation for that either.

We need to be able to reconstruct the whole script with the expanded commands given in.

SeeminglyScience: > We need to be able to reconstruct the whole script with the expanded commands given in.

Ideally the `ExpandAliasResult` would be changed to take a `WorkspaceEdit`.  Then you could do something like this:

```csharp
ScriptBlockAst ast = Parser.ParseInput(request.Text, out _, out _);
CommandAst[] commands = ast
    .FindAll(a => a is CommandAst, searchNestedScriptBlocks: true)
    .Cast<CommandAst>()
    .ToArray();

var commandNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
foreach (CommandAst command in commands)
{
    string commandName = command.GetCommandName();
    if (string.IsNullOrEmpty(commandName))
    {
        continue;
    }

    commandNames.Add(commandName);
}

var psCommand = new PSCommand()
    .AddCommand("Microsoft.PowerShell.Core\\Get-Command")
    .AddParameter("CommandTypes", CommandTypes.Alias)
    .AddParameter("ErrorAction", ActionPreference.Ignore)
    .AddParameter("Name", commandNames.ToArray());

AliasInfo[] aliases = (await _powerShellContextService
    .ExecuteCommandAsync<AliasInfo>(psCommand, sendErrorToHost: false).ConfigureAwait(false))
    .ToArray();

var aliasMap = new Dictionary<string, string>(
    capacity: aliases.Length,
    StringComparer.OrdinalIgnoreCase);

foreach (AliasInfo alias in aliases)
{
    aliasMap.Add(alias.Name, alias.Definition);
}

var edits = new List<TextEdit>(capacity: commands.Length);
foreach (CommandAst command in commands)
{
    string commandName = command.GetCommandName();
    if (string.IsNullOrEmpty(commandName))
    {
        continue;
    }

    if (!aliasMap.TryGetValue(commandName, out string definition))
    {
        continue;
    }

    IScriptExtent extent = command.Extent;
    var edit = new TextEdit()
    {
        NewText = definition,
        Range = new Range(
            new Position(extent.StartLineNumber - 1, extent.StartColumnNumber - 1),
            new Position(extent.EndLineNumber - 1, extent.EndColumnNumber - 1))
    };

    edits.Add(edit);
}

// make and return WorkspaceEdit
```

andyleejordan:
Sorry, just going to close this, it's been 4 years. Still up for grabs if someone wants to improve the way that expand alias function works. It's still a TODO.

PR #1210, opened 2020-02-27
Fix bad context state after debugging

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/2364.
Fixes https://github.com/PowerShell/vscode-powershell/issues/2488.
May fix https://github.com/PowerShell/vscode-powershell/issues/2396.

After debugging is stopped, the PowerShellContextService was being marked as being in session state aborted. Now we set the state to ready.

This fix isn't ideal, since it's made without a full understanding of the conceptual state machine of the PowerShellContextService, but it does fix the issues for the coming release while we sift through the actual architectural cause of the problem (and ideally fix it down the line)

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
@@ -1240,6 +1240,10 @@ namespace Microsoft.PowerShell.EditorServices.Services
                 this.PromptNest.GetPowerShell(isReadLine: false).BeginStop(null, null);
             }
 
+            // TODO:
+            // This lock and state reset are a temporary fix at best.
+            // We need to investigate how the debugger should be interacting
+            // with PowerShell in this cancellation scenario.
             this.sessionStateLock.Wait();
             try
             {
--- a/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
@@ -1225,14 +1225,10 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
             if (this.PromptNest.IsInDebugger)
             {
+                this.versionSpecificOperations.StopCommandInDebugger(this);
                 if (shouldAbortDebugSession)
                 {
-                    this.versionSpecificOperations.StopCommandInDebugger(this);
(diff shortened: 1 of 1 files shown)


SeeminglyScience:
Can you test <kbd>CTRL</kbd> + <kbd>C</kbd> in various situations? Like hit a debugger stop, manually run `Start-Sleep 10`, try to <kbd>CTRL</kbd> + <kbd>C</kbd>?

I feel like I did something like this early on and it broke something but I can't remember what or why...

rjmholt:
> Like hit a debugger stop, manually run Start-Sleep 10, try to CTRL + C?

This one seems to work nicely.

I imagine there is an issue somewhere here. But if we don't find it, I think this change should go in, since the devil we know is quite bad

rjmholt:
Ah, it's the second `Ctrl-C` that gets ignored. I'm ok with that one for now. If we can find a better implementation before Tuesday, we can put it in this PR. Otherwise will bugfix it later

rjmholt:
@SeeminglyScience do you know where we process ctrl C? I'd like to get a better picture of what's going on so I can think about what a fix might look like

rjmholt:
Ok, I've now made it so that `sleep 10` works. I'm not sure if this could corrupt some state, but it seems not to have in my limited uses.

TylerLeonhardt:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/TylerLeonhardt/PowerShellEditorServices/pullRequest?prid=5079709&bid=16709818)

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -745,6 +744,7 @@ public Task<RunspaceHandle> GetRunspaceHandleAsync(CancellationToken cancellatio
                 // Don't change our SessionState for ReadLine.
                 if (!executionOptions.IsReadLine)
                 {
+                    await this.sessionStateLock.WaitAsync();
```
TylerLeonhardt: Codacy Issue found: [Add '.ConfigureAwait(false)' to this call to allow execution to continue in any thread.](https://app.codacy.com/manual/TylerLeonhardt/PowerShellEditorServices/pullRequest?prid=5079709)

rjmholt: Good point @TylerLeonhardt

PR #1214, opened 2020-02-28
Fix startup in AllSigned execution policy

rjmholt:
Fixes https://github.com/PowerShell/vscode-powershell/issues/2504.

We were setting the execution policy before importing PSReadLine. Now we set it afterward.

Not sure what happens with PSSA here.

Ideally we could solve this by linking against those libraries rather than importing them as modules.

EDIT: It also turns out that Execution Policy scope order wasn't what we thought it would be, so we were setting a lower scope first. We now set the CurrentUser scope ahead of the LocalMachine scope.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
@@ -367,11 +367,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
                     powerShellVersion.ToString());
             }
 
-            if (VersionUtils.IsWindows)
-            {
-                this.SetExecutionPolicy();
-            }
-
             // Set up the runspace
             this.ConfigureRunspace(this.CurrentRunspace);
 
@@ -427,6 +422,11 @@ namespace Microsoft.PowerShell.EditorServices.Services
             else
             {
                 this.PromptContext = new LegacyReadLineContext(this);
+            }
+
+            if (VersionUtils.IsWindows)
+            {
+                this.SetExecutionPolicy();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -2102,25 +2102,24 @@ private void SetExecutionPolicy()
...
-            // This is the order of precedence we want to follow, skipping the Process scope
+            // We want to ignore policy settings, since we'll already have those anyway.
+            // Then we need to look at the CurrentUser setting, and then the LocalMachine setting.
             //
             // Get-ExecutionPolicy -List emits PSObjects with Scope and ExecutionPolicy note properties
             // set to expected values, so we must sift through those.
+
             ExecutionPolicy policyToSet = ExecutionPolicy.Bypass;
-            for (int i = policies.Count - 1; i >= 0; i--)
+            var currentUserPolicy = (ExecutionPolicy)policies[policies.Count - 2].Members["ExecutionPolicy"].Value;
```
SeeminglyScience: I like this, can we rely on the order to remain consistent though?

rjmholt: I [*think* so](https://github.com/PowerShell/PowerShell/blob/69bf7043c0c7471679265778e12053c1c40b0000/src/System.Management.Automation/security/SecuritySupport.cs#L121-L133)

rjmholt: The other alternative is to look for user scope and then to look for machine scope

rjmholt: But I don't think there's any plan to update anything around execution policy

PR #1394, opened 2020-11-24
fix #1393: Always use local help to return Full help

### LINKED ISSUE #1393: Get-Help always launches online help if user has the -Online parameter defaulted to $true
I originally [posted this bug in the VS Code PowerShell extension GitHub repo](https://github.com/PowerShell/vscode-powershell/issues/3071), but after digging through the code found that the bug was actually in this repository.

Basically the issue is that if the user has defaulted the `Get-Help` cmdlet's `-Online` parameter to $true, such as by putting this line in their PowerShell Profile:

```powershell
$PSDefaultParameterValues.Add("Get-Help:Online", $true)
```

Then `Get-Help` will always launch a browser window. This causes an issue because VS Code (and other editors) use the `getCommandHelp` provider to provide the text for intellisense and tooltips. So whenever the editor would normally show intellisense or a tooltip, it instead opens a browser window to the online help page of the cmdlet. You can see this demonstrated in this gif.

PowerShellHelpAutomaticallyOpens

deadlydog:
If the user has overridden the Get-Help -Online parameter to always be true, then the ShowHelpHandler will always launch a new browser window to the cmdlet help.
Instead, we want it to return the full cmdlet text as intended.

fixes #1393 
fixes https://github.com/PowerShell/vscode-powershell/issues/3071

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Utilities/CommandHelpers.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Utilities/CommandHelpers.cs
@@ -139,6 +139,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
                 // We use .Name here instead of just passing in commandInfo because
                 // CommandInfo.ToString() duplicates the Prefix if one exists.
                 .AddParameter("Name", commandInfo.Name)
+                .AddParameter("Online", false)
                 .AddParameter("ErrorAction", "Ignore");
 
             var results = await powerShellContext.ExecuteCommandAsync<PSObject>(command, sendOutputToHost: false, sendErrorToHost: false).ConfigureAwait(false);


TylerLeonhardt:
I'm curious... does that actually fix your issue?

I would have guessed that you needed to add it here:
https://github.com/PowerShell/PowerShellEditorServices/blob/7a828a0147e62cddf5f54f7f675fdc4c7f947c7a/src/PowerShellEditorServices/Services/PowerShellContext/Utilities/CommandHelpers.cs#L141

TylerLeonhardt:
What you've changed is what gets run by the Command Explorer added by @corbob which only gets fired if you hit this little question mark:

image

TylerLeonhardt:
Also just want to say, thank you so much for contributing. I love it when folks open issues and then submit PRs! You're always welcome here!

deadlydog:
Thanks for the quick response on this @TylerLeonhardt :)

You may very well be right that I applied the fix in the wrong spot. I read the[ Development docs on the ReadMe](https://github.com/PowerShell/PowerShellEditorServices#development) and [the Contribution Guidelines](http://powershell.github.io/PowerShellEditorServices/CONTRIBUTING.html), but neither of them told me how to test the changes. I was able to build the project file as the guide suggested, but wasn't sure how to test the changes. Perhaps there are docs on how to do that and I just overlooked them?

I was also hoping to add some unit tests around the new functionality, but could not get the test projects to build. This is what I tried:
- In VS 2019: The build fails trying to build the `PowerShellEditorServices.TestHost` project for net461 and netcoreapp2.1 saying
    > CS0234	The type or namespace name 'Protocol' does not exist in the namespace 'Microsoft.PowerShell.EditorServices' (are you missing an assembly reference?)	PowerShellEditorServices.Test.Host (net461), PowerShellEditorServices.Test.Host (netcoreapp2.1)	C:\dev\Git\3rdParty\PowerShellEditorServices\test\PowerShellEditorServices.Test.Host\DebugAdapterTests.cs	6	Active

- In my local VS Code: The `Run Build Task` succeeds, but the `Run Test Task` fails with a bunch of similar errors to:
    > C:\dev\Git\3rdParty\PowerShellEditorServices\test\PowerShellEditorServices.Test.Host\DebugAdapterTests.cs(6,43): error CS0234: The type or namespace name 'Protocol' does not exist in the namespace 'Microsoft.PowerShell.EditorServices' (are you missing an assembly reference?) [C:\dev\Git\3rdParty\PowerShellEditorServices\test\PowerShellEditorServices.Test.Host\PowerShellEditorServices.Test.Host.csproj]

- And in the VS Code dev container: The `Run Build Task` fails with:
    > \#\#\# Installation complete for version .                                                                                                                                .NET installation complete                                                                                                                                             ERROR: Cannot find path '/workspaces/PowerShellEditorServices/.dotnet/dotnet' because it does not exist.                                                               At /workspaces/PowerShellEditorServices/PowerShellEditorServices.build.ps1:116 char:25                                                                                 +     $script:dotnetExe = Resolve-Path $dotnetExePath   

I do have dotnet.exe on my Windows PATH though, and have v 5.0.100 of the .Net SDK installed.

I figured somebody had broken the tests in the `master` branch and that the dev container was still a work in progress. Maybe that's a bad assumption though?

That said, I do believe you're right though that I fixed the wrong code. Although, I'm wondering if we actually want the fix applied to both places. I'll update the PR and let you comment on it. Thanks Tyler!

TylerLeonhardt:
The easiest way to test your change manually is by using this guide:
https://github.com/PowerShell/vscode-powershell/blob/master/docs/development.md

That will start a local build of the vscode extension with your local build of PSES.

deadlydog:
Awesome, that would have been useful. Once I followed the code from the vscode-powershell repo to this one, I never went back to it. I kind of assumed there'd be a stand-alone way to test this git repo. This would be useful info to have on the ReadMe 😉

TylerLeonhardt:
>  Still, this would be useful info to have on the ReadMe 😉

Done:
https://github.com/PowerShell/PowerShellEditorServices/blob/master/README.md#development

deadlydog:
Following that doc I was able to test my changes :) 

1. I re-added `$PSDefaultParameterValues.Add("Get-Help:Online", $true)` to my PowerShell Profile
1. Built the PowerShellEditorServices from the `master` branch
1. Ran `code --extensionDevelopmentPath="c:\path\to\vscode-powershell"` to confirm I could reproduce the issue, both from intellisense and tooltips.
1. Switched the PowerShellEditorServices to my branch and rebuilt
1. Ran `code --extensionDevelopmentPath="c:\path\to\vscode-powershell"` again and confirmed that the intellisense and tooltips now show up properly without popping open a new browser window.

I also tested clicking on the Help icon in the PowerShell Command Explorer that you mentioned, but it seems it always open using the Online help, even when the `-Online` parameter is not defaulted to true. Since my original "fix" didn't seem to make any difference, I just removed it.

So this PR now has only the change that actually addresses the issue I reported, and I've confirmed it does indeed solve the problem.

TylerLeonhardt (approved):
LGTM Given the side effect of this change, I can't think of any test that could be written with a reasonable amount of work so we can skip that here.

Will merge tomorrow!

PR #1410, opened 2021-02-03
Fix publish step

andschwa:
The template allows us to always copy so we can depend on it for publish.


andschwa:
Forgot about this, didn't know it would fail until the auto-build kicked off.

TravisEz13 (approved):

rjmholt (approved):

PR #1411, opened 2021-02-13
Fix New-EditorFile failing when no Editor window open

corbob:
Fixes PowerShell/vscode-powershell#3180

This feels janky, but short of moving the context gathering later and doing a bunch of work we can't complete because of Temp session I'm not sure of a better way.

I've also updated the help to reference the correct filename.

### DIFF
--- a/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
@@ -134,6 +134,7 @@ function New-EditorFile {
             $psEditor.Workspace.NewFile()
             $editorContext = $psEditor.GetEditorContext()
         }
+
         if (!$editorContext) {
             return
         }
--- a/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
@@ -125,7 +125,15 @@ function New-EditorFile {
     end {
         # If editorContext is null, then we're in a Temp session and
         # this cmdlet won't work so return early.
-        $editorContext = $psEditor.GetEditorContext()
+        try {
+            $editorContext = $psEditor.GetEditorContext()
+        }
+        catch {
+            # If there's no editor, this throws an error. Create a new file, and grab the context here.
+            # This feels really hacky way to do it, but not sure if there's another way to detect editor context...
+            $psEditor.Workspace.NewFile()
+            $editorContext = $psEditor.GetEditorContext()
+        }
         if (!$editorContext) {
             return
         }
--- a/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
@@ -86,10 +86,10 @@ function Unregister-EditorCommand {
     Creates and opens a new foo.ps1 in your editor
 .EXAMPLE
     PS > Get-Process | New-EditorFile proc.txt
-    Creates and opens a new foo.ps1 in your editor with the contents of the call to Get-Process
+    Creates and opens a new proc.txt in your editor with the contents of the call to Get-Process
 .EXAMPLE
     PS > Get-Process | New-EditorFile proc.txt -Force
-    Creates and opens a new foo.ps1 in your editor with the contents of the call to Get-Process. Overwrites the file if it already exists
+    Creates and opens a new proc.txt in your editor with the contents of the call to Get-Process. Overwrites the file if it already exists
 .INPUTS
     Path
     an array of files you want to open in your editor


module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
```diff
@@ -125,7 +125,15 @@ function New-EditorFile {
...
-        $editorContext = $psEditor.GetEditorContext()
+        try {
+            $editorContext = $psEditor.GetEditorContext()
+        }
+        catch {
+            # If there's no editor, this throws an error. Create a new file, and grab the context here.
+            # This feels really hacky way to do it, but not sure if there's another way to detect editor context...
+            $psEditor.Workspace.NewFile()
+            $editorContext = $psEditor.GetEditorContext()
+        }
```
rjmholt, line 136:
```suggestion
        }

```

rjmholt (approved):

module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
```diff
@@ -125,7 +125,15 @@ function New-EditorFile {
     end {
         # If editorContext is null, then we're in a Temp session and
         # this cmdlet won't work so return early.
-        $editorContext = $psEditor.GetEditorContext()
+        try {
+            $editorContext = $psEditor.GetEditorContext()
+        }
+        catch {
+            # If there's no editor, this throws an error. Create a new file, and grab the context here.
+            # This feels really hacky way to do it, but not sure if there's another way to detect editor context...
```
andschwa, line 133: This definitely appears like it would work, so I don't mind the approach. I would like to know if there's a way to test if an editor context exists (if so we can avoid exception handling), and if creating a new file on the workspace is the best way to generate a new editor context. If this is the only approach, then I feel like our interface is lacking because as someone new to the code, I would have expected something like:

* `GetEditorContext()` (which exists)
* `HasEditorContext()` (essentially says if the above would throw)
* `NewEditorContext()` (to create one, instead of relying what I presume is a side-effect of `Workspace.NewFile()`)

rjmholt:
Ah ok so @andschwa's comment has reminded me of the deeper nature of this API, and that it's essentially an old API that had to undergo some changes when we migrated to Omnisharp.

Here's the actual implementation of the method that gets called:

https://github.com/PowerShell/PowerShellEditorServices/blob/b243fe3dd3baf88399b370437a827778b86e24ad/src/PowerShellEditorServices/Services/PowerShellContext/EditorOperationsService.cs#L34-L49

Some things we should look at here are:

- The full error message
- Whether we should change the synchronous call from using `.Result` to `.GetAwaiter().GetResult()` (I think so)
- How we could update this API to make this scenario more seamless

rjmholt:
@corbob what does the full error message look like here?

rjmholt:
In particular wrt @andschwa's proposal, my counter proposal would just be for `GetEditorContext()` to create a new context if need be (or return null if there's no language server)

corbob:
> @corbob what does the full error message look like here?

Here's a screenshot for color coding, followed by the copy/paste of the error:

image


```
┏[corbob]
┖[E:]> $pseditor.GetEditorContext()

MethodInvocationException: Exception calling "GetEditorContext" with "0" argument(s): "One or more errors occurred. (Internal error. )"
┏[corbob]
┖[E:]> $Error| select *


PSMessageDetails      :
Exception             : System.Management.Automation.MethodInvocationException: Exception calling "GetEditorContext" with "0" argument(s): "One or more errors occurred. (Internal error. )"
                         ---> System.AggregateException: One or more errors occurred. (Internal error. )
                         ---> OmniSharp.Extensions.JsonRpc.Server.InternalErrorException: Internal error.
                           at OmniSharp.Extensions.JsonRpc.ResponseRouter.ResponseRouterReturnsImpl.Returning[TResponse](CancellationToken cancellationToken)
                           at Microsoft.PowerShell.EditorServices.Services.EditorOperationsService.GetEditorContextAsync() in
                        D:\a\1\s\src\PowerShellEditorServices\Services\PowerShellContext\EditorOperationsService.cs:line 41
                           --- End of inner exception stack trace ---
                           at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
                           at CallSite.Target(Closure , CallSite , Object )
                           --- End of inner exception stack trace ---
                           at System.Management.Automation.ExceptionHandlingOps.ConvertToMethodInvocationException(Exception exception, Type typeToThrow, String methodName, Int32 numArgs, MemberInfo 
                        memberInfo)
                           at CallSite.Target(Closure , CallSite , Object )
                           at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0)
                           at System.Management.Automation.Interpreter.DynamicInstruction`2.Run(InterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
TargetObject          :
CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
FullyQualifiedErrorId : AggregateException
ErrorDetails          :
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at <ScriptBlock>, <No file>: line 1
PipelineIterationInfo : {}
```

rjmholt:
> `---> OmniSharp.Extensions.JsonRpc.Server.InternalErrorException: Internal error.`

Well that's not very helpful.

Ok my proposal for now is that we move the logic in the script inside the API call so that the file is created internally to it

andschwa:
Thanks @corbob!

PR #1415, opened 2021-02-22
Fix configuration processing to ensure that profiles are loaded

rjmholt:


andschwa (approved):

rjmholt:
So the test failures are due to help not being installed on Ubuntu I think. Not sure why that is

PR #1418, opened 2021-02-25
Update CHANGELOG for v2.3.0

andschwa:


PowerShellEditorServices.Common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.3.0</VersionPrefix>
-    <VersionSuffix>preview.4</VersionSuffix>
+    <VersionPrefix>2.4.0</VersionPrefix>
+    <VersionSuffix>preview</VersionSuffix>
```
rjmholt, line 4: I think you just want to remove this entry no?

rjmholt: Unless this is the post-release bump, in which case I'd call it `preview.1`

andschwa: No, this is the commit after the tag we're back to 2.4.0-preview.

rjmholt: I don't follow -- are you saying it's not the post-release version bump?

In any case, we've never had a prerelease tag that didn't match `preview\.\d+`

andschwa: Oi, our messages were crossed. Fixed.

andschwa:
Ping @rjmholt for a review.

rjmholt (approved):

PR #1424, opened 2021-03-12
Update to use OmniSharp 0.19.0

andschwa:
This updates our base dependency, OmniSharp, from 0.18.3 to 0.19.0, which included many API changes. The approach taken to reduce code was the replacement of interface-implementations with ABC derivations, providing a few things for free for each of the derived classes, and is more inline with the expectations of the library developers. There are a couple classes that implement multiple interfaces, and have been left TODO for a switch to an ABC implementation. Furthermore, the C# language specification needed to be updated because of OmniSharp's use of new language features (namely `record`s), and I found that our xUnit package was misaligned.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ThreadsHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ThreadsHandler.cs
@@ -16,13 +16,8 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
         {
             return Task.FromResult(new ThreadsResponse
             {
-                // TODO: What do I do with these?
-                Threads = new Container<OmniSharp.Extensions.DebugAdapter.Protocol.Models.Thread>(
-                    new OmniSharp.Extensions.DebugAdapter.Protocol.Models.Thread
-                    {
-                        Id = 1,
-                        Name = "Main Thread"
-                    })
+                // TODO: This is an empty container of threads...do we need to make a thread?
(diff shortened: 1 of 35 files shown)


andschwa:
Working on getting this put together. FYI @rjmholt and @david-driscoll

andschwa:
The PSES tests don't pass, but it builds! The extension tests pass, I think.

andschwa:
This is ready @rjmholt!

rjmholt (approved):

andschwa:
Supersedes dependabot's PRs: #1419, #1421, and #1422.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ThreadsHandler.cs
```diff
@@ -16,13 +16,8 @@ public Task<ThreadsResponse> Handle(ThreadsArguments request, CancellationToken
...
             {
-                // TODO: What do I do with these?
-                Threads = new Container<OmniSharp.Extensions.DebugAdapter.Protocol.Models.Thread>(
-                    new OmniSharp.Extensions.DebugAdapter.Protocol.Models.Thread
-                    {
-                        Id = 1,
-                        Name = "Main Thread"
-                    })
+                // TODO: This is an empty container of threads...do we need to make a thread?
+                Threads = new Container<System.Threading.Thread>()
```
andschwa, line 20: I thought this was weird, but we already didn't really know what to do with it (see "What do I do with these?")

PR #1426, opened 2021-03-18
Bump CI images and enable tests on Apple M1

andschwa:
Pulled this out of #1424 to get it in quicker. Also deletes a confusing folder of long-dead code.


PowerShellEditorServices.build.ps1
```diff
@@ -19,19 +19,19 @@ param(
 #Requires -Modules @{ModuleName="InvokeBuild";ModuleVersion="3.2.1"}
 
 $script:IsUnix = $PSVersionTable.PSEdition -and $PSVersionTable.PSEdition -eq "Core" -and !$IsWindows
+$script:IsRosetta = $script:IsUnix -and (sysctl -n sysctl.proc_translated) -eq 1 # Mac M1
```
rjmholt, line 22: `$IsMacOS` might be more specific here -- otherwise I'd expect this command to fail on Linux

PowerShellEditorServices.build.ps1
```diff
@@ -19,19 +19,19 @@ param(
 #Requires -Modules @{ModuleName="InvokeBuild";ModuleVersion="3.2.1"}
 
 $script:IsUnix = $PSVersionTable.PSEdition -and $PSVersionTable.PSEdition -eq "Core" -and !$IsWindows
```
rjmholt, line 21: Honestly this test is silly, I'd change it to:

```powershell
$script:IsUnix = $IsLinux -or $IsMacOS
```

Or:

```powershell
$script:IsUnix = $IsWindows -eq $false
```

rjmholt (approved):

PowerShellEditorServices.build.ps1
rjmholt:
```suggestion
$script:IsRosetta = $IsMacOS -and (sysctl -n sysctl.proc_translated) -eq 1 # Mac M1
```

PowerShellEditorServices.build.ps1
andschwa: Agreed. Also `Unix` is wrong, it's just `*nix` so I changed to `IsNix`. Neither of these platforms are Unix.

PowerShellEditorServices.build.ps1
andschwa: Good catch!

PR #1427, opened 2021-03-21
Add '$' as trigger character for completion

MartinGC94:
Fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1176 so IntelliSense gets triggered automatically when writing variables.


andschwa (approved):
While I'm not seeing any reason not to take this change, I went ahead and tested locally and I already get completions on `$`. Then I saw you're using the LSP server via a client in Vim, and so I'm chalking this up to some weirdness with VS Code asking for completions anyway. Thanks for the PR!

MartinGC94:
In VS code quickSuggestions are enabled by default which is why $ triggers IntelliSense in VS code out of the box. If you disable them the Powershell specific triggers still start IntelliSense. I don't know if that's working as intended, but it's definitely the behavior I want because quickSuggestions makes IntelliSense a bit too aggressive IMO.

PR #1473, opened 2021-05-07
Delete deprecated files

andschwa:
* The `launch.json` file is unused
* The log analyzer is no longer relevant
* The signing XML was replaced with an ADO task
* The release scripts are unused
* The copyright header script will be replaced


andschwa:
@rjmholt You *know* how much I don't like dead code hanging around!

rjmholt (approved):

PR #1474, opened 2021-05-11
Add more logging and internal documentation

andschwa:
Just writing up what I've learned and odd stuff I've spotted while reading through the code. Also added at least a bit more logging to help debug https://github.com/PowerShell/vscode-powershell/issues/3306.

PR #1475, opened 2021-05-12
WIP: Implement a safe wrapper around ILanguageServerAdapter to ensure messages are only sent after initialization

rjmholt:


andschwa:
> So it turns out this creates a deadlock here:
> 
> https://github.com/rjmholt/PowerShellEditorServices/blob/1d3e9da884bdd91ea3d33f88035b4a2a3e1d5824/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs#L1888-L1893
> 
> The fix is in [1d3e9da](https://github.com/rjmholt/PowerShellEditorServices/commit/1d3e9da884bdd91ea3d33f88035b4a2a3e1d5824)

So with the additional logging in #1474 I've been able to rule out the `executionStatusChanged` notification as the cause of the deadlock. Still suspecting the `runspaceChanged` notification, but still testing...

andschwa:
The only news I really have is that when it _doesn't_ repro, I never see the "runspace changed" notification. So our suspicion is half-proved?

andschwa:
Wait, @rjmholt I have info! When I get the issue to repro, it is _always_ accompanied by:

```
Unexpected notification workspace/didChangeConfiguration {"Method":"workspace/didChangeConfiguration","Params":[[[[[[[[[]],[[]]]],[[]],[[]],[[]],[[]],[[]],[[[[]],[[]]]],[[]],[[]],[[]],[[[[]]]],[[]],[[]],[[[[]],[[]]]],[[[[]],[[]]]],[[[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]]]],[[[[]],[[]],[[]],[[]],[[]]]],[[[[]]]],[[[[]],[[]],[[]],[[]],[[]]]],[[[[]],[[]],[[]]]],[[[[]],[[]]]]]],[[[[[[]]]],[[[[]],[[]],[[]],[[]],[[]]]],[[[[]],[[]]]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[[[]],[[]],[[]],[[]]]],[[]],[[]],[[]],[[]],[[]],[[[[]]]]]],[[[[[[]],[[]],[[]]]],[[]],[[]],[[]],[[]],[[]],[[[[]],[[]],[[null]]]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[]],[[[[]],[[]],[[]]]],[[]]]]]]],"TraceParent":null,"TraceState":null,"$type":"Notification"}
```

andschwa:
Thanks for this Rob! Looks like we found the root issue elsewhere though, closing...

PR #1490, opened 2021-05-26
Remove unnecessary exclusions

MartinGC94:
Now that IntelliSense no longer imports modules automatically (see https://github.com/PowerShell/PowerShellEditorServices/pull/1484 ) we can remove the exclusion list meant to avoid certain modules from being imported unintentionally.


rjmholt:
If this is just for intellisense, then it makes sense.

But we have to do everything we can to avoid loading those modules generally, since all but the most recent version of PackageManagement have a nasty bug that deadlocks stuff.

MartinGC94:
I see. I thought it was just an IntelliSense issue but I can see that it's called in 3 other places so I guess the exclusions are still needed.

PR #1494, opened 2021-06-04
Work around `dotnet publish` bug

andschwa:
This fixes https://github.com/PowerShell/vscode-powershell/issues/3395


andschwa:
I kinda think I've hit the same issue as someone else hit with `msbuild` (which if I remember correctly is what `dotnet` became a port of): https://github.com/dotnet/standard/issues/1334

andschwa:
I also tried playing around with the various options like `--self-contained`, since this _is_ a library and we moved from .NET 2.1 to 3.1+ and that behavior changed, but `dotnet publish` still never included `netstandard.dll` once the CLI was on 6.0.

andschwa:
Linking regression report upstream: https://github.com/dotnet/sdk/issues/18101

rjmholt (approved):

Issue #1502, opened 2021-06-17
NullReferenceException when $profile doesn’t exist


dkattan:
https://github.com/PowerShell/PowerShellEditorServices/blob/f5006ae3748985a0ed111f08c0408dfcea562ff2/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs#L349

Can we get a null check here to prevent profile loading if the $profile value is null?

SeeminglyScience:
Explicit casts do not throw on null in C#.  Are you sure that's where the exception you're seeing is coming from?

rjmholt:
> Can we get a null check here

AFAIK you're the only person to report this, and we'd be happy to accept a PR

dkattan:
> Explicit casts do not throw on null in C#. Are you sure that's where the exception you're seeing is coming from?

You are correct, it is actually on line 382.

![image](https://user-images.githubusercontent.com/1424395/122570137-162cfa80-d011-11eb-9b7e-e06a3097a9ba.png)


> > Can we get a null check here
> 
> AFAIK you're the only person to report this, and we'd be happy to accept a PR

I'm working on a merge request now.

PR #1504, opened 2021-06-18
Bugfix/1503 send feature changes telemetry throws exception when properties of incoming settings are null

### LINKED ISSUE #1503: SendFeatureChangesTelemetry throws exception when properties of incomingSettings are null
image

I'll get a merge request created with all these tedious bugfixes so we don't need to go back and forth.

dkattan:
Fixes #1503  and #1502

### DIFF
--- a/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
+++ b/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
@@ -378,7 +378,8 @@ namespace Microsoft.PowerShell.EditorServices.Commands
         private string GetProfilePathFromProfileObject(PSObject profileObject, ProfileUserKind userKind, ProfileHostKind hostKind)
         {
             string profilePathName = $"{userKind}{hostKind}";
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
```diff
@@ -378,7 +378,8 @@ private EditorServicesConfig CreateConfigObject()
         private string GetProfilePathFromProfileObject(PSObject profileObject, ProfileUserKind userKind, ProfileHostKind hostKind)
         {
             string profilePathName = $"{userKind}{hostKind}";
-
+            if (profileObject is null)
+                return null;
```
andschwa, line 382:
```suggestion
            if (profileObject is null)
            {
                return null;
            }
```

andschwa: Just a style nitpick.

andschwa (approved):

PR #1508, opened 2021-06-23
Bump version to `v2.4.5`

andschwa:
We only run builds on ADO with the intent to release, therefore we need to just
always sign them. Otherwise the compliance scripts' default behavior is to sign
when the build is manually triggered, which is not our workflow. We push a
release branch and expect that automatic build to result in signing.


andschwa:
Fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1509

andschwa:
Ugh, our flaky tests are becoming flakier.

rjmholt (approved):

andschwa:
Heh @rjmholt I missed a spot, there are two places we need `shouldSign: true`. Have to run the internal CI to know if it works 😖

rjmholt (approved):

PR #1513, opened 2021-06-25
Import `InvokePesterStub.ps1` from `vscode-powershell` (with history)

andschwa:
I used `git filter-repo` to export the history and adjust the path of this script, so that I could then use `git merge --allow-unrelated-history` to merge said script in with its development history intact.

### DIFF
--- a/module/PowerShellEditorServices/InvokePesterStub.ps1
+++ b/module/PowerShellEditorServices/InvokePesterStub.ps1
@@ -1,0 +1,192 @@
+#!/usr/bin/env pwsh
+
+<#
+.SYNOPSIS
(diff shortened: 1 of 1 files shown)


andschwa:
Ok, verified that _this_ build signs it properly:

```
| InvokePesterStub.ps1      | Pass   | 6.66KB   | 21.56          | 0.87        | 0.48        | 17.24         | 0           | 
```

andschwa:
Ok! All confirmed. The file is properly signed by this build, then incorporated into the module, which then uses it from the correct path and the "Run Pester Tests" command works as expected. @rjmholt

rjmholt (approved):

PR #1459, opened 2021-04-23
WIP: Reimplement the way PowerShell is run and hosted in PSES with a dedicated pipeline thread consumer

rjmholt:

PR #1514, opened 2021-06-25
Load only bundled `PSReadLine`

### LINKED ISSUE #1493: Lock bundled modules (especially PSReadLine)
Per https://github.com/PowerShell/vscode-powershell/issues/3385 users can encounter incompatibility issues by installing newer versions of our dependencies since we do not lock PSES to the bundled version. This is unfortunate and causes more issues than we believe are fixed by picking up the latest installed version. We are going to change PSES to only load the bundled module so that we, the developers, can manually update and test our dependencies. In dire scenarios, users _can_ still just overwrite the bundled module in the extension folder, and at that point will either be following our specific instructions for a workaround or going outside our supported scenarios.

https://github.com/PowerShell/PowerShellEditorServices/blob/32f318ec0e80ff1786660fc5920b8667a93324e9/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs#L27-L43

andschwa:
This definitely would resolve #1493, but @rjmholt I have a hunch there's a better way to load a bundled dependency. That is: I think we were using this inline script only so we could support loading any arbitrary version of PSReadLine, and since we're deprecating that behavior, there's _probably_ a way to just straight up load it into the session without executing some script like this.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
@@ -22,6 +22,13 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
             "..",
             "..",
             "..",
+#if TEST
+            // When using xUnit (dotnet test) the assemblies are deployed to the
+            // test project folder, invalidating our relative path assumption.
+            "..",
+            "..",
+            "module",
+#endif
             "PSReadLine");
 
         private static readonly Lazy<CmdletInfo> s_lazyInvokeReadLineForEditorServicesCmdletInfo = new Lazy<CmdletInfo>(() =>
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
@@ -24,21 +24,13 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
             "..",
             "PSReadLine");
 
+        // TODO: Does this have to be done in an inline script?
         private static readonly string ReadLineInitScript = $@"
             [System.Diagnostics.DebuggerHidden()]
             [System.Diagnostics.DebuggerStepThrough()]
             param()
             end {{
-                $module = Get-Module -ListAvailable PSReadLine |
-                    Where-Object {{ $_.Version -ge '2.0.2' }} |
-                    Sort-Object -Descending Version |
-                    Select-Object -First 1
-                if (-not $module) {{
-                    Import-Module '{_psReadLineModulePath.Replace("'", "''")}'
-                    return [Microsoft.PowerShell.PSConsoleReadLine]
-                }}
-
-                Import-Module -ModuleInfo $module
+                Import-Module '{_psReadLineModulePath.Replace("'", "''")}'
                 return [Microsoft.PowerShell.PSConsoleReadLine]
             }}";
 
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
@@ -24,16 +24,6 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
             "..",
             "PSReadLine");
 
-        // TODO: Does this have to be done in an inline script?
-        private static readonly string ReadLineInitScript = $@"
-            [System.Diagnostics.DebuggerHidden()]
-            [System.Diagnostics.DebuggerStepThrough()]
-            param()
-            end {{
-                Import-Module '{_psReadLineModulePath.Replace("'", "''")}'
-                return [Microsoft.PowerShell.PSConsoleReadLine]
-            }}";
-
         private static readonly Lazy<CmdletInfo> s_lazyInvokeReadLineForEditorServicesCmdletInfo = new Lazy<CmdletInfo>(() =>
         {
             var type = Type.GetType("Microsoft.PowerShell.EditorServices.Commands.InvokeReadLineForEditorServicesCommand, Microsoft.PowerShell.EditorServices.Hosting");
@@ -89,14 +79,15 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
             using (var pwsh = PowerShell.Create())
             {
                 pwsh.Runspace = runspace;
-                var psReadLineType = pwsh
-                    .AddScript(ReadLineInitScript, useLocalScope: true)
-                    .Invoke<Type>()
-                    .FirstOrDefault();
+                pwsh.AddCommand("Microsoft.PowerShell.Core\\Import-Module")
+                    .AddParameter("Name", _psReadLineModulePath)
+                    .Invoke();
+
+                var psReadLineType = Type.GetType("Microsoft.PowerShell.PSConsoleReadLine, Microsoft.PowerShell.PSReadLine2");
 
                 if (psReadLineType == null)
                 {
-                    logger.LogWarning("PSReadLine unable to be loaded: {Reason}", pwsh.HadErrors ? pwsh.Streams.Error[0].ToString() : "<Unknown reason>");
+                    logger.LogWarning("PSConsoleReadline type not found: {Reason}", pwsh.HadErrors ? pwsh.Streams.Error[0].ToString() : "<Unknown reason>");
                     return false;
                 }
 
@@ -109,7 +100,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
                     // The Type we got back from PowerShell doesn't have the members we expected.
                     // Could be an older version, a custom build, or something a newer version with
                     // breaking changes.
-                    logger.LogWarning("PSReadLine unable to be loaded: {Reason}", e);
+                    logger.LogWarning("PSReadLineProxy unable to be initialized: {Reason}", e);
                     return false;
                 }
             }
--- a/test/PowerShellEditorServices.Test/Session/PowerShellContextTests.cs
+++ b/test/PowerShellEditorServices.Test/Session/PowerShellContextTests.cs
@@ -1,17 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.Logging.Abstractions;
-using Microsoft.PowerShell.EditorServices.Services;
-using Microsoft.PowerShell.EditorServices.Services.PowerShellContext;
-using Microsoft.PowerShell.EditorServices.Test.Shared;
-using Microsoft.PowerShell.EditorServices.Utility;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Management.Automation;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.PowerShell.EditorServices.Services;
+using Microsoft.PowerShell.EditorServices.Services.PowerShellContext;
+using Microsoft.PowerShell.EditorServices.Test.Shared;
+using Microsoft.PowerShell.EditorServices.Utility;
 using Xunit;
 
 namespace Microsoft.PowerShell.EditorServices.Test.Console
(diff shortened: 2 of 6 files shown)


andschwa:
The only other way I can see to do this is with the `InitialSessionState.ImportPSModule` method, which we don't have access to at this point because we only have the runspace (which is created after the initial session state).

dkattan:
Hey guys, could this potentially be why my tests are failing for my constrained runspace support branch?

https://github.com/PowerShell/PowerShellEditorServices/pull/1507/checks?check_run_id=2907295896

When CanLaunchScriptWithNoBreakpointsAsync runs, it throws
```
Module 'PackageManagement' is in currently in use or you don't have the required permissions.
```

Which led me to [this issue](https://github.com/OneGet/oneget/issues/344#issuecomment-783482503) that points to PSReadLine as a potential culprit, which then led me here.





I'm unable to reproduce the issue locally, which is infuriating, but I see

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -98,7 +80,9 @@ internal class PSReadLinePromptContext : IPromptContext
             {
                 pwsh.Runspace = runspace;
                 var psReadLineType = pwsh
-                    .AddScript(ReadLineInitScript, useLocalScope: true)
+                    .AddCommand("Import-Module")
```
dkattan, line 83: We can't assume that Import-Module will be available in all runspaces. I think we should centralize all module loading to PowerShellContextService.Create();

If you check out my changes in https://github.com/PowerShell/PowerShellEditorServices/pull/1507 you'll see that I conditionally load modules using either Import-Module or InitialSessionState.ImportPSModulesFromPath depending on whether or not the runspace has a FileSystem provider. We already know at that point whether or not PSReadline will be required based on hostStartupInfo.ConsoleReplEnabled, so I say we load it there instead.

dkattan: On further inspection, I realized that this PowerShell instance is temporary for getting the type information for the proxy so whatever I load with PowerShellContextService.Create(); won't show up here. Carry on.

dkattan: Turns out you guys are assigning the runspace to the newly created PowerShell, therefore it _is_ inheriting the language mode/modules so I stand by my initial assertion of centrally locating all modules.

In case it is unclear, in order to load modules that require file system access into a constrained runspace that has no FileSystem provider, you have to use InitialSessionState.ImportPSModulesFromPath, Import-Module won't work.

dkattan: Also, are you sure this actually worked? 
image

dkattan: Here's how I think we should do it:

```c#
var assemblies = AppDomain.CurrentDomain.GetAssemblies();
                var psReadLineType = assemblies.FirstOrDefault(a => a.FullName.Contains("Microsoft.PowerShell.PSReadLine2"))?.ExportedTypes?.FirstOrDefault(a => a.Name == "PSConsoleReadLine");

                if (psReadLineType == null)
                {
                    logger.LogWarning("PSReadLine unable to be loaded: {Reason}", pwsh.HadErrors ? pwsh.Streams.Error[0].ToString() : "<Unknown reason>");
                    return false;
                }

                try
                {
                    readLineProxy = new PSReadLineProxy(psReadLineType, logger);
                }
                catch (InvalidOperationException e)
                {
                    // The Type we got back from PowerShell doesn't have the members we expected.
                    // Could be an older version, a custom build, or something a newer version with
                    // breaking changes.
                    logger.LogWarning("PSReadLine unable to be loaded: {Reason}", e);
                    return false;
                }
```

dkattan: Actually, now that we're not using PowerShell to resolve it, we don't need to pass runspace in anymore at all, it could be

```c#
internal static bool TryGetPSReadLineProxy(
    ILogger logger,
    out PSReadLineProxy readLineProxy)
{
    readLineProxy = null;
    logger.LogTrace("Attempting to load PSReadLine");
    var psReadLineAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName.Contains("Microsoft.PowerShell.PSReadLine2"));
    if(psReadLineAssembly is null)
    {
        logger.LogWarning("Microsoft.PowerShell.PSReadLine2 not found in loaded assemblies");
        return false;
    }
    var psReadLineType = psReadLineAssembly?.ExportedTypes?.FirstOrDefault(a => a.Name == "PSConsoleReadLine");

    if (psReadLineType == null)
    {
        logger.LogWarning("PSConsoleReadLine type not found in Microsoft.PowerShell.PSReadLine2");
        return false;
    }

    try
    {
        readLineProxy = new PSReadLineProxy(psReadLineType, logger);
    }
    catch (InvalidOperationException e)
    {
        // The Type we got back from PowerShell doesn't have the members we expected.
        // Could be an older version, a custom build, or something a newer version with
        // breaking changes.
        logger.LogWarning("PSReadLine unable to be loaded: {Reason}", e);
        return false;
    }

    return true;
}
```

andschwa: > Also, are you sure this actually worked?

No I'm not, I only know the tests passed. I needed to put this through some manual testing still 😆

andschwa:
I meant to open this PR as a draft.

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
dkattan: I found a few other rough spots in my testing over the weekend, I have my own pull request with the changes:
https://github.com/PowerShell/PowerShellEditorServices/pull/1516

Turns out we weren't at all honoring the path provided by the -BundledModulesPath param

dkattan: Ah you already found it! Now if I can only figure out why the tests never pass when I create pull requests...

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -98,7 +80,9 @@ internal class PSReadLinePromptContext : IPromptContext
...
+                        .AddParameter("Name", _psReadLineModulePath.Replace("'", "''"))
```
rjmholt, line 84: I wouldn't have thought the escaping here was needed, since it's already being defined as the value associated with the parameter, although admittedly I don't know for sure.

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
rjmholt: Just trying to get the assembly itself will fail since it's not yet loaded. Also PSES surfaces PSReadLine commands in the terminal, so we have to import the module properly.

dkattan:
@rjmholt My proposed changes assume that the module has been loaded earlier in PowerShellContextService.Create

andschwa:
Ok, not it's tested and ready.

andschwa:
@dkattan, @rjmholt and I discussed this in a call earlier: we'll take this PR for now, split #1516 into 1) fixing bundledModulePath and 2) importing PSReadLine earlier and centrally with `importPSModule`, merge 1) ASAP, and robustly test 2), which will setup for taking in #1507 after #1459.

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -98,7 +80,10 @@ internal class PSReadLinePromptContext : IPromptContext
             {
                 pwsh.Runspace = runspace;
                 var psReadLineType = pwsh
-                    .AddScript(ReadLineInitScript, useLocalScope: true)
+                    .AddCommand("Microsoft.PowerShell.Core\\Import-Module")
+                        .AddParameter("Name", _psReadLineModulePath)
+                    .AddStatement()
+                    .AddScript("[Microsoft.PowerShell.PSConsoleReadLine]", true)
```
rjmholt, line 86: So reading through this again, it's not strictly required to do this part in PowerShell.

We could do it from C# instead like:

```csharp
pwsh
    .AddCommand("Microsoft.PowerShell.Core\\Import-Module")
    .AddParameter("Name", _psReadLineModulePath)
    .Invoke();

Type psReadLineType = Type.GetType("Microsoft.PowerShell.PSConsoleReadline");
```

andschwa: Heh, I think I like that more, yeah. But are we just yak shaving?

rjmholt: Yeah this is a bit of bike shedding here, but there are some advantages to the latter form. The main one being performance is better (this code being on the startup path).

andschwa: Not gonna lie, I dig the GitHub integration in VS Code.

andschwa: Ok so doing that seems to cause the console to not have PSReadLine initialized. Like, it's imported, but not being used. Not sure the right term for that.

andschwa: Strange:

```
2021-07-01 10:54:54.222 -07:00 [INF] Attempting to load PSReadLine
2021-07-01 10:54:54.378 -07:00 [VRB] Loaded into load context "Default" System.Runtime.Loader.DefaultAssemblyLoadContext #0: Microsoft.PowerShell.PSReadLine2, Version=2.1.0.0, Culture=neutral, PublicKeyToken=null
2021-07-01 10:54:54.380 -07:00 [VRB] Loaded into load context "Default" System.Runtime.Loader.DefaultAssemblyLoadContext #0: mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
2021-07-01 10:54:54.466 -07:00 [WRN] PSReadLine unable to be loaded: <Unknown reason>
```

andschwa: But `Get-Module` lists it.

andschwa: Ok so we got that working, but adding a test for this is proving soooo difficult.

rjmholt (approved):

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -22,26 +22,15 @@ internal class PSReadLinePromptContext : IPromptContext
             "..",
             "..",
             "..",
+#if TEST
```
andschwa, line 25: OHH @rjmholt this won't work because we need it defined during test compilation but for this project. Ugh how on earth do we do that?

andschwa: Found the answer https://stackoverflow.com/questions/59249847/passing-define-constants-to-dotnet-build-kills-target-framework-symbols

test/PowerShellEditorServices.Test/Session/PowerShellContextTests.cs
```diff
@@ -143,6 +147,17 @@ public async Task CanResolveAndLoadProfilesForHostId()
             Assert.Equal(expectedString, result.FirstOrDefault(), true);
         }
 
+        [Trait("Category", "PSReadLine")]
+        [SkippableFact]
+        public async Task CanGetPSReadLineProxy()
+        {
+            Skip.If(IsWindows, "This test doesn't work on Windows for some reason.");
```
andschwa, line 154: @rjmholt For later debugging 😭

PR #1516, opened 2021-06-28
Honor BundledModulesPath when loading modules in PowerShellContextService

dkattan:

PR #1522, opened 2021-07-06
Run new PSReadLine test on Windows

andschwa:
Finishing #1514. Also replaces #1516.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
@@ -11,6 +11,7 @@ using System.Management.Automation.Host;
 using System.Management.Automation.Remoting;
 using System.Management.Automation.Runspaces;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -280,6 +281,15 @@ namespace Microsoft.PowerShell.EditorServices.Services
             // that started PowerShell Editor Services. This is because the PowerShell Integrated Console
             // should have the same LanguageMode of whatever is set by the system.
             initialSessionState.LanguageMode = languageMode;
+
+            // We set the process scope's execution policy (which is really the runspace's scope) to
+            // Bypass so we can import our bundled modules. This is equivalent in scope to the CLI
+            // argument `-Bypass`, which (for instance) the extension passes. Thus we emulate this
+            // behavior for consistency such that unit tests can pass in a similar environment.
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                initialSessionState.ExecutionPolicy = ExecutionPolicy.Bypass;
+            }
 
             Runspace runspace = RunspaceFactory.CreateRunspace(psHost, initialSessionState);
 
--- a/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
@@ -33,10 +33,18 @@ namespace Microsoft.PowerShell.EditorServices.Services
     /// </summary>
     internal class PowerShellContextService : IHostSupportsInteractiveSession
     {
-        private static readonly string s_commandsModulePath = Path.GetFullPath(
-            Path.Combine(
-                Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
-                "../../Commands/PowerShellEditorServices.Commands.psd1"));
+        // This is a default that can be overriden at runtime by the user or tests.
+        private static string s_bundledModulePath = Path.GetFullPath(Path.Combine(
+                Path.GetDirectoryName(typeof(PowerShellContextService).Assembly.Location),
+                "..",
+                "..",
+                ".."));
+
+        private static string s_commandsModulePath => Path.GetFullPath(Path.Combine(
+            s_bundledModulePath,
+            "PowerShellEditorServices",
+            "Commands",
+            "PowerShellEditorServices.Commands.psd1"));
 
         private static readonly Action<Runspace, ApartmentState> s_runspaceApartmentStateSetter;
         private static readonly PropertyInfo s_writeStreamProperty;
@@ -190,9 +198,16 @@ namespace Microsoft.PowerShell.EditorServices.Services
             OmniSharp.Extensions.LanguageServer.Protocol.Server.ILanguageServerFacade languageServer,
             HostStartupInfo hostStartupInfo)
         {
+            var logger = factory.CreateLogger<PowerShellContextService>();
+
             Validate.IsNotNull(nameof(hostStartupInfo), hostStartupInfo);
 
-            var logger = factory.CreateLogger<PowerShellContextService>();
+            // Respect a user provided bundled module path.
+            if (Directory.Exists(hostStartupInfo.BundledModulePath))
+            {
+                logger.LogTrace($"Using new bundled module path: {hostStartupInfo.BundledModulePath}");
+                s_bundledModulePath = hostStartupInfo.BundledModulePath;
+            }
 
             bool shouldUsePSReadLine = hostStartupInfo.ConsoleReplEnabled
                 && !hostStartupInfo.UsesLegacyReadLine;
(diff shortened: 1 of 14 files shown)


andschwa:
Finally stole your code @dkattan

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -79,23 +84,37 @@ internal class PSReadLinePromptContext : IPromptContext
...
+                    return false;
+                }
+
                 var psReadLineType = Type.GetType("Microsoft.PowerShell.PSConsoleReadLine, Microsoft.PowerShell.PSReadLine2");
 
                 if (psReadLineType == null)
                 {
-                    logger.LogWarning("PSConsoleReadline type not found: {Reason}", pwsh.HadErrors ? pwsh.Streams.Error[0].ToString() : "<Unknown reason>");
-                    return false;
+                    logger.LogTrace("PSConsoleReadline type not found using Type.GetType(), searching all loaded assemblies...");
```
andschwa, line 109: @rjmholt This should only happen in the case that it needs to. When `Type.GetType` works, then this logic is skipped. So it feels fine, though I still wish I knew _why_ it was necessary.

test/PowerShellEditorServices.Test/Session/PowerShellContextTests.cs
```diff
@@ -148,14 +147,14 @@ public async Task CanResolveAndLoadProfilesForHostId()
         }
 
         [Trait("Category", "PSReadLine")]
-        [SkippableFact]
-        public async Task CanGetPSReadLineProxy()
+        [Fact]
+        public void CanGetPSReadLineProxy()
         {
-            Skip.If(IsWindows, "This test doesn't work on Windows for some reason.");
```
andschwa, line 154: For so many reasons 😆

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
dkattan: Not saying it's "right" but I keep seeing code similar to this come up when I research the issue. I think we will get to the bottom of it as we continue to refactor but this works.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -2129,75 +2144,6 @@ private static string GetStringForPSCommand(PSCommand psCommand)
             return stringBuilder.ToString();
         }
 
-        private void SetExecutionPolicy()
```
dkattan, line 2132: Huge fan of removing this and relying on the initialSessionState. I look forward to bringing the entire InitialSessionState in instead of just the LanguageMode so we more accurately create runspaces similar to the parent.

dkattan (approved):
Looks good!

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -32,10 +33,18 @@ namespace Microsoft.PowerShell.EditorServices.Services
     /// </summary>
     internal class PowerShellContextService : IHostSupportsInteractiveSession
     {
-        private static readonly string s_commandsModulePath = Path.GetFullPath(
-            Path.Combine(
+        // This is a default that can be overriden at runtime by the user or tests.
+        private static string s_bundledModulePath = Path.GetFullPath(Path.Combine(
                 Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
```
andschwa, line 38: Wait @rjmholt is this going to be accurate?

andschwa: `GetExecutingAssembly` seems to be PowerShell itself, so this becomes `$psHome` (and isn't caught by tests because xUnit _forces_ us to use a distinct module path).

andschwa: How could this have even worked in the past?! I don't think it did.

andschwa: Hm, no, my VS Code updated and it got funky.

andschwa:
@rjmholt I've been looking at `Start-EditorServices` and the use of `-BundledModulesPath` there and things seem...fishy. Let's look tomorrow if you have time, because this PR as is does not respect the user settings. So we haven't quite fixed that yet.

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -87,15 +74,31 @@ internal class PSReadLinePromptContext : IPromptContext
...
                 {
-                    logger.LogWarning("PSConsoleReadline type not found: {Reason}", pwsh.HadErrors ? pwsh.Streams.Error[0].ToString() : "<Unknown reason>");
-                    return false;
+                    // NOTE: For some reason `Type.GetType(...)` can fail to find the type,
+                    // and in that case, this search through the `AppDomain` for some reason will succeed.
+                    // It's slower, but only happens when needed.
+                    logger.LogTrace("PSConsoleReadline type not found using Type.GetType(), searching all loaded assemblies...");
+                    var allAssemblies = AppDomain.CurrentDomain.GetAssemblies();
+                    var assemblies = allAssemblies.FirstOrDefault(a => a.FullName.Contains("Microsoft.PowerShell.PSReadLine2"));
+                    psReadLineType = assemblies?.ExportedTypes?.FirstOrDefault(a => a.FullName == "Microsoft.PowerShell.PSConsoleReadLine");
```
rjmholt, line 96:
```suggestion
                    Type psrlType = AppDomain.CurrentDomain
                        .GetAssemblies()
                        .FirstOrDefault(asm => asm.GetName().Name.Equals("Microsoft.PowerShell.PSReadLine2"))
                        ?.ExportedTypes
                        ?.FirstOrDefault(type => type.FullName.Equals("Microsoft.PowerShell.PSConsoleReadLine"));
```

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
rjmholt: My suspicion is that this always worked in script because it always runs on the pipeline thread and through PowerShell's own type search logic. When we sift through the whole AppDomain (which is kind of a dead concept) for a type, we ignore assembly load contexts, and so that might be why the direction reflection query fails but the search succeeds. Just a theory though. An ALC shouldn't be thread-dependent though and I don't think anyone is using something like `EnterContextualReflection()`

rjmholt (approved):

andschwa:
So to actually assert that we now respect `BundledModulePath` threw me on a bit of a loop (thanks @rjmholt for the help). E.g. with this JSON setting:

```json
"powershell.developer.bundledModulesPath": "/Users/andschwa/Downloads/PowerShellEditorServices"
```

Two things have to be true: 1. the extension must be run in the VS Code debugger (such that "developer" is honored) and 2. a previous dev was "smart" and added silent fallback logic to use the extension's path if the supplied `bundledModulesPath` did not contain the actual modules, so it can only be tested with a perfect layout. I had originally set it to just `/Users/andschwa/` expecting a _failure_ instead of a silent ignoring of the path. It's fine, just unexpected and something I had to learn.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -281,6 +297,15 @@ public static Runspace CreateRunspace(PSHost psHost, PSLanguageMode languageMode
             // should have the same LanguageMode of whatever is set by the system.
             initialSessionState.LanguageMode = languageMode;
 
+            // We set the process scope's execution policy (which is really the runspace's scope) to
+            // Bypass so we can import our bundled modules. This is equivalent in scope to the CLI
+            // argument `-Bypass`, which (for instance) the extension passes. Thus we emulate this
+            // behavior for consistency such that unit tests can pass in a similar environment.
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                initialSessionState.ExecutionPolicy = ExecutionPolicy.Bypass;
```
andschwa, line 306: @dkattan So we resurrected the `SetExecutionPolicy()` function because on closer inspection, the actual issue was that in the _unit test_ scenario we weren't having `Bypass` set (because the codebase's expectation is that the startup script is run with `-ExecutionPolicy Bypass`, like what the extension does), and the `SetExecutionPolicy()` actually sets the execution policy _back_ to the original inherited scoped policy, not the process policy. That is, it override's the process/runspace's policy after everything is loaded to be equivalent to the user's machine/user policy.

andschwa: @rjmholt I might rename that function `ResetExecutionPolicy()` 😆

dkattan: That works. I'm a fan of removing large chunks of code, as long as it doesn't change expected functionality :D

PR #1523, opened 2021-07-13
Pass InitialSessionState instead of LanguageMode

dkattan:


andschwa:
I like, I like. We should move the setting of execution policy to bypass in the ISS to happen _only_ in the test factory, since it shouldn't be necessary otherwise.

dkattan:
> I like, I like. We should move the setting of execution policy to bypass in the ISS to happen _only_ in the test factory, since it shouldn't be necessary otherwise.

Done! All tests pass.

rjmholt:
Closing in favour of https://github.com/PowerShell/PowerShellEditorServices/pull/1526

andschwa:
Thanks @dkattan!

PR #1526, opened 2021-07-20
Initialize runspaces with `InitialSessionState` object

andschwa:
Instead of passing values for `LanguageMode` and `ExecutionPolicy` as
fields that are set on an `InitialSessionState` object that's eventually
created, we now pass a fully initialized `InitialSessionState` object
from the start.

We also rename `SetExecutionPolicy` to `RestoreExecutionPolicy` to
reflect what it actually does: restores the user's policy after we
overrode it with `Bypass` while initializing the integrated console's
runspace. We override it because we need to be able to load `PSReadLine`
etc. without policy issues.

Finally, we fix the EditorConfig setting to enforce spaces after
control-flow keywords (such as `if`) because, while our codebase is
inconsistent, it is our preferred style going forward.

This replaces #1523.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,7 +11,7 @@ insert_final_newline = true
 indent_size = 4
 trim_trailing_whitespace = true
 csharp_space_before_open_square_brackets = true
-csharp_space_after_keywords_in_control_flow_statements = false
+csharp_space_after_keywords_in_control_flow_statements = true
 
 [*.{json}]
 indent_size = 2
--- a/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
+++ b/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
@@ -349,7 +349,16 @@ namespace Microsoft.PowerShell.EditorServices.Commands
             var profile = (PSObject)GetVariableValue("profile");
 
             var hostInfo = new HostInfo(HostName, HostProfileId, HostVersion);
-            var editorServicesConfig = new EditorServicesConfig(hostInfo, Host, SessionDetailsPath, bundledModulesPath, LogPath)
+
+            var initialSessionState = Runspace.DefaultRunspace.InitialSessionState;
+            initialSessionState.LanguageMode = Runspace.DefaultRunspace.SessionStateProxy.LanguageMode;
+
+            var editorServicesConfig = new EditorServicesConfig(
+                hostInfo,
+                Host,
+                SessionDetailsPath,
+                bundledModulesPath,
+                LogPath)
             {
                 FeatureFlags = FeatureFlags,
                 LogLevel = LogLevel,
@@ -357,7 +366,7 @@ namespace Microsoft.PowerShell.EditorServices.Commands
                 AdditionalModules = AdditionalModules,
                 LanguageServiceTransport = GetLanguageServiceTransport(),
                 DebugServiceTransport = GetDebugServiceTransport(),
-                LanguageMode = Runspace.DefaultRunspace.SessionStateProxy.LanguageMode,
+                InitialSessionState = initialSessionState,
                 ProfilePaths = new ProfilePathConfig
                 {
                     AllUsersAllHosts = GetProfilePathFromProfileObject(profile, ProfileUserKind.AllUsers, ProfileHostKind.AllHosts),
--- a/src/PowerShellEditorServices/Hosting/HostStartupInfo.cs
+++ b/src/PowerShellEditorServices/Hosting/HostStartupInfo.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections.Generic;
-using System.Management.Automation;
 using System.Management.Automation.Host;
+using System.Management.Automation.Runspaces;
 
 namespace Microsoft.PowerShell.EditorServices.Hosting
 {
@@ -92,10 +92,10 @@ namespace Microsoft.PowerShell.EditorServices.Hosting
         public string LogPath { get; }
 
         /// <summary>
-        /// The language mode inherited from the orginal PowerShell process.
-        /// This will be used when creating runspaces so that we honor the same language mode.
+        /// The InitialSessionState will be inherited from the orginal PowerShell process. This will
+        /// be used when creating runspaces so that we honor the same InitialSessionState.
         /// </summary>
-        public PSLanguageMode LanguageMode { get; }
+        public InitialSessionState InitialSessionState { get; }
 
         /// <summary>
         /// The minimum log level of log events to be logged.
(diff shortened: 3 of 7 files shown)


.editorconfig
```diff
@@ -11,7 +11,7 @@ insert_final_newline = true
 indent_size = 4
 trim_trailing_whitespace = true
 csharp_space_before_open_square_brackets = true
-csharp_space_after_keywords_in_control_flow_statements = false
+csharp_space_after_keywords_in_control_flow_statements = true
```
andschwa, line 14: @rjmholt I have no idea why this was `false` previously.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -274,38 +276,17 @@ public RunspaceDetails CurrentRunspace
...
         /// </summary>
         /// <param name="psHost">The PSHost that will be used for this Runspace.</param>
-        /// <param name="languageMode">The language mode inherited from the orginal PowerShell process. This will be used when creating runspaces so that we honor the same language mode.</param>
+        /// <param name="initialSessionState">This will be used when creating runspaces so that we honor the same InitialSessionState.</param>
         /// <returns></returns>
-        public static Runspace CreateRunspace(PSHost psHost, PSLanguageMode languageMode)
+        public static Runspace CreateRunspace(PSHost psHost, InitialSessionState initialSessionState)
         {
-            InitialSessionState initialSessionState;
-            if (Environment.GetEnvironmentVariable("PSES_TEST_USE_CREATE_DEFAULT") == "1") {
```
andschwa, line 289: @rjmholt FYI this code is gone but I don't believe we were using it for anything at the moment.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -434,10 +415,9 @@ public static Runspace CreateRunspace(PSHost psHost, PSLanguageMode languageMode
                 this.PromptContext = new LegacyReadLineContext(this);
             }
 
-            if (VersionUtils.IsWindows)
-            {
-                this.SetExecutionPolicy();
-            }
+            // Finally, restore the runspace's execution policy to the user's policy instead of
+            // Bypass.
+            this.RestoreExecutionPolicy();
```
andschwa, line 420: @rjmholt I moved the "if windows" to "if not windows" inside the function itself.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -2152,9 +2132,20 @@ private static string GetStringForPSCommand(PSCommand psCommand)
             return stringBuilder.ToString();
         }
 
-        private void SetExecutionPolicy()
+        /// <summary>
+        /// This function restores the execution policy for the process by examining the user's
+        /// execution policy hierarchy. We do this because the process policy will always be set to
+        /// Bypass when initializing our runspaces.
+        /// </summary>
+        internal void RestoreExecutionPolicy()
```
andschwa, line 2140: Thanks for renaming this @dkattan!

test/PowerShellEditorServices.Test/PowerShellContextFactory.cs
```diff
@@ -40,6 +41,16 @@ internal static class PowerShellContextFactory
...
             PowerShellContextService powerShellContext = new PowerShellContextService(logger, null, isPSReadLineEnabled: false);
+            var initialSessionState = InitialSessionState.CreateDefault();
+            // We set the process scope's execution policy (which is really the runspace's scope) to
+            // `Bypass` so we can import our bundled modules. This is equivalent in scope to the CLI
+            // argument `-ExecutionPolicy Bypass`, which (for instance) the extension passes. Thus
+            // we emulate this behavior for consistency such that unit tests can pass in a similar
+            // environment.
+            if (VersionUtils.IsWindows)
+            {
+                initialSessionState.ExecutionPolicy = ExecutionPolicy.Bypass;
```
andschwa, line 52: Yay! This if finally just in the test path.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
andschwa: @dkattan This still needs to be called because it restores the process scope policy when it was set on the CLI (such as by the extension).

dkattan: I think I learned this the hard way in my constrained runspace branch. I makes a lot more sense now that it is called RestoreExecutionPolicy.

dkattan (approved):

.editorconfig
rjmholt: Not sure either -- might be an autogenerated setting from a long time ago

src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
```diff
@@ -349,15 +349,24 @@ private EditorServicesConfig CreateConfigObject()
             var profile = (PSObject)GetVariableValue("profile");
 
             var hostInfo = new HostInfo(HostName, HostProfileId, HostVersion);
-            var editorServicesConfig = new EditorServicesConfig(hostInfo, Host, SessionDetailsPath, bundledModulesPath, LogPath)
+
+            var initialSessionState = Runspace.DefaultRunspace.InitialSessionState;
```
rjmholt, line 353:
```suggestion
            InitialSessionState initialSessionState = Runspace.DefaultRunspace.InitialSessionState;
```

src/PowerShellEditorServices/Hosting/HostStartupInfo.cs
```diff
@@ -135,7 +135,7 @@ public sealed class HostStartupInfo
         /// <param name="currentUsersProfilePath">The path to the user specific profile.</param>
         /// <param name="featureFlags">Flags of features to enable.</param>
         /// <param name="additionalModules">Names or paths of additional modules to import.</param>
-        /// <param name="languageMode">The language mode inherited from the orginal PowerShell process. This will be used when creating runspaces so that we honor the same language mode.</param>
+        /// <param name="initialSessionState">The language mode inherited from the orginal PowerShell process. This will be used when creating runspaces so that we honor the same initialSessionState including allowed modules, cmdlets and language mode.</param>
```
rjmholt, line 138: The initial part of this description is no longer accurate

rjmholt (approved):

src/PowerShellEditorServices/Hosting/HostStartupInfo.cs
andschwa: Oh dang, I thought I double-checked all of these. Ah well. For the documentation project.

test/PowerShellEditorServices.Test/PowerShellContextFactory.cs
```diff
@@ -40,6 +41,16 @@ internal static class PowerShellContextFactory
         public static PowerShellContextService Create(ILogger logger)
         {
             PowerShellContextService powerShellContext = new PowerShellContextService(logger, null, isPSReadLineEnabled: false);
+            var initialSessionState = InitialSessionState.CreateDefault();
```
andschwa, line 44: Wait, @rjmholt @dkattan shouldn't this have been `CreateDefault2()`???

PR #1532, opened 2021-07-30
Make `ExecuteCommandAsync` cancellable 

andschwa:
This hasn't solved the problem where the PowerShell debugger and the VS Code debugger somehow disconnect from each other when `System.Windows.Forms` is added to the PowerShell session, but it does improve the situation somewhat. The debugger can be stopped now (though it cannot be restarted yet).

PR #1533, opened 2021-08-02
Enable and fix many .NET Code Analysis warnings

andschwa:
Pulling this out of #1532, enabling and fixing the code analysis warnings that matter to us one-by-one, and treating them as build errors. I should note that there are many other warnings we're ignoring for now and should fix eventually.

### DIFF
--- a/src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
+++ b/src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
@@ -28,40 +28,31 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
             this.languageServer = languageServer;
         }
 
-        internal async Task CreateAsync()
-        {
-            await languageServer.SendRequestAsync(
+        internal Task CreateAsync() =>
+            languageServer.SendRequestAsync(
                 NewCustomViewRequest.Method,
                 new NewCustomViewRequest
                 {
                     Id = this.Id,
                     Title = this.Title,
                     ViewType = this.ViewType,
-                }
-            );
-        }
+                });
 
-        public async Task Show(ViewColumn viewColumn)
-        {
-            await languageServer.SendRequestAsync(
+        public Task Show(ViewColumn viewColumn) =>
+            languageServer.SendRequestAsync(
                 ShowCustomViewRequest.Method,
                 new ShowCustomViewRequest
                 {
                     Id = this.Id,
                     ViewColumn = viewColumn
-                }
-            );
-        }
+                });
 
-        public async Task Close()
-        {
-            await languageServer.SendRequestAsync(
+        public Task Close() =>
+            languageServer.SendRequestAsync(
                 CloseCustomViewRequest.Method,
                 new CloseCustomViewRequest
                 {
                     Id = this.Id,
-                }
-            );
-        }
+                });
     }
 }
--- a/src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
+++ b/src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
@@ -21,9 +21,8 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
         {
         }
 
-        public async Task SetContentAsync(string htmlBodyContent)
-        {
-            await languageServer.SendRequestAsync(
+        public Task SetContentAsync(string htmlBodyContent) =>
+            languageServer.SendRequestAsync(
                 SetHtmlContentViewRequest.Method,
                 new SetHtmlContentViewRequest
                 {
@@ -31,31 +30,24 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
                     HtmlContent = new HtmlContent { BodyContent = htmlBodyContent }
                 }
             );
-        }
 
-        public async Task SetContentAsync(HtmlContent htmlContent)
-        {
-            HtmlContent validatedContent =
-                new HtmlContent()
-                {
-                    BodyContent = htmlContent.BodyContent,
-                    JavaScriptPaths = this.GetUriPaths(htmlContent.JavaScriptPaths),
-                    StyleSheetPaths = this.GetUriPaths(htmlContent.StyleSheetPaths)
-                };
-
-            await languageServer.SendRequestAsync(
+        public Task SetContentAsync(HtmlContent htmlContent) =>
+            languageServer.SendRequestAsync(
                 SetHtmlContentViewRequest.Method,
                 new SetHtmlContentViewRequest
                 {
                     Id = this.Id,
-                    HtmlContent = validatedContent
+                    HtmlContent = new HtmlContent()
+                    {
+                        BodyContent = htmlContent.BodyContent,
+                        JavaScriptPaths = this.GetUriPaths(htmlContent.JavaScriptPaths),
+                        StyleSheetPaths = this.GetUriPaths(htmlContent.StyleSheetPaths)
+                    }
                 }
             );
-        }
 
-        public async Task AppendContentAsync(string appendedHtmlBodyContent)
-        {
-            await languageServer.SendRequestAsync(
+        public Task AppendContentAsync(string appendedHtmlBodyContent) =>
+            languageServer.SendRequestAsync(
                 AppendHtmlContentViewRequest.Method,
                 new AppendHtmlContentViewRequest
                 {
@@ -63,7 +55,6 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
                     AppendedHtmlBodyContent = appendedHtmlBodyContent
                 }
             );
-        }
 
         private string[] GetUriPaths(string[] filePaths)
         {
--- a/src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
+++ b/src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
@@ -40,8 +40,8 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
                     HtmlContent = new HtmlContent()
                     {
                         BodyContent = htmlContent.BodyContent,
-                        JavaScriptPaths = this.GetUriPaths(htmlContent.JavaScriptPaths),
-                        StyleSheetPaths = this.GetUriPaths(htmlContent.StyleSheetPaths)
+                        JavaScriptPaths = HtmlContentView.GetUriPaths(htmlContent.JavaScriptPaths),
+                        StyleSheetPaths = HtmlContentView.GetUriPaths(htmlContent.StyleSheetPaths)
                     }
                 }
             );
@@ -56,7 +56,7 @@ namespace Microsoft.PowerShell.EditorServices.VSCode.CustomViews
                 }
             );
 
-        private string[] GetUriPaths(string[] filePaths)
+        private static string[] GetUriPaths(string[] filePaths)
         {
             return
                 filePaths?
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
@@ -75,7 +75,7 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
 #pragma warning disable CS4014
             // Trigger the clean up of the debugger. No need to wait for it.
-            Task.Run(_psesDebugServer.OnSessionEnded);
+            Task.Run(_psesDebugServer.OnSessionEnded, cancellationToken);
 #pragma warning restore CS4014
 
             return new DisconnectResponse();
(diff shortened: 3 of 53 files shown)


src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
```diff
@@ -38,7 +38,7 @@ internal async Task CreateAsync()
                     Title = this.Title,
                     ViewType = this.ViewType,
                 }
-            );
+            ).ConfigureAwait(false);
```
rjmholt, line 41: We should just return the result here rather than await it, and make the method non-`async`

src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
```diff
@@ -50,7 +50,7 @@ public async Task Show(ViewColumn viewColumn)
                     Id = this.Id,
                     ViewColumn = viewColumn
                 }
-            );
+            ).ConfigureAwait(false);
```
rjmholt, line 53: Ditto

src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
```diff
@@ -61,7 +61,7 @@ public async Task Close()
                 {
                     Id = this.Id,
                 }
-            );
+            ).ConfigureAwait(false);
```
rjmholt, line 64: Ditto

src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
```diff
@@ -30,7 +30,7 @@ public async Task SetContentAsync(string htmlBodyContent)
                     Id = this.Id,
                     HtmlContent = new HtmlContent { BodyContent = htmlBodyContent }
                 }
-            );
+            ).ConfigureAwait(false);
```
rjmholt, line 33: Ditto

src/PowerShellEditorServices.VSCode/CustomViews/HtmlContentView.cs
```diff
@@ -50,7 +50,7 @@ public async Task SetContentAsync(HtmlContent htmlContent)
                     Id = this.Id,
                     HtmlContent = validatedContent
                 }
-            );
+            ).ConfigureAwait(false);
```
rjmholt, line 53: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -485,7 +485,7 @@ private void CleanupRunspace(RunspaceDetails runspaceDetails)
         /// <returns>A RunspaceHandle instance that gives access to the session's runspace.</returns>
         public Task<RunspaceHandle> GetRunspaceHandleAsync()
         {
-            return this.GetRunspaceHandleImplAsync(CancellationToken.None, isReadLine: false);
+            return this.GetRunspaceHandleImplAsync(false, CancellationToken.None);
```
rjmholt, line 488: Would still be helpful to name the boolean here

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
```diff
@@ -497,7 +497,7 @@ public Task<RunspaceHandle> GetRunspaceHandleAsync()
         /// <returns>A RunspaceHandle instance that gives access to the session's runspace.</returns>
         public Task<RunspaceHandle> GetRunspaceHandleAsync(CancellationToken cancellationToken)
         {
-            return this.GetRunspaceHandleImplAsync(cancellationToken, isReadLine: false);
+            return this.GetRunspaceHandleImplAsync(false, cancellationToken);
```
rjmholt, line 500: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/Session/InvocationEventQueue.cs
```diff
@@ -113,7 +113,7 @@ internal async Task InvokeOnPipelineThreadAsync(Action<PowerShell> invocationAct
         {
             var request = new InvocationRequest(pwsh =>
             {
-                using (_promptNest.GetRunspaceHandle(CancellationToken.None, isReadLine: false))
+                using (_promptNest.GetRunspaceHandle(false, CancellationToken.None))
```
rjmholt, line 116: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -177,12 +177,12 @@ public void AbortReadLine()
 
         public void WaitForReadLineExit()
         {
-            using (_promptNest.GetRunspaceHandle(CancellationToken.None, isReadLine: true))
+            using (_promptNest.GetRunspaceHandle(true, CancellationToken.None))
```
rjmholt, line 180: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs
```diff
@@ -177,12 +177,12 @@ public void AbortReadLine()
...
             { }
         }
 
         public async Task WaitForReadLineExitAsync() {
-            using (await _promptNest.GetRunspaceHandleAsync(CancellationToken.None, isReadLine: true).ConfigureAwait(false))
+            using (await _promptNest.GetRunspaceHandleAsync(true, CancellationToken.None).ConfigureAwait(false))
```
rjmholt, line 185: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/Session/PromptNest.cs
```diff
@@ -264,10 +264,10 @@ internal RunspaceHandle GetRunspaceHandle(CancellationToken cancellationToken, b
             // is in process.
             if (isReadLine && !_powerShellContext.IsCurrentRunspaceOutOfProcess())
             {
-                GetRunspaceHandleImpl(cancellationToken, isReadLine: false);
+                GetRunspaceHandleImpl(false, cancellationToken);
```
rjmholt, line 267: Ditto

src/PowerShellEditorServices/Services/PowerShellContext/Session/PromptNest.cs
```diff
@@ -295,10 +295,10 @@ internal async Task<RunspaceHandle> GetRunspaceHandleAsync(CancellationToken can
             // is in process.
             if (isReadLine && !_powerShellContext.IsCurrentRunspaceOutOfProcess())
             {
-                await GetRunspaceHandleImplAsync(cancellationToken, isReadLine: false).ConfigureAwait(false);
+                await GetRunspaceHandleImplAsync(false, cancellationToken).ConfigureAwait(false);
```
rjmholt, line 298: Ditto

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
```diff
@@ -62,7 +62,7 @@ internal static class AstOperations
             ILogger logger,
             CancellationToken cancellationToken)
         {
-            if (!s_completionHandle.Wait(0))
+            if (!s_completionHandle.Wait(0, cancellationToken))
```
rjmholt, line 65: This shouldn't need the cancellation passed through, since it's a `Wait(0)` -- the token should never get a chance to cancel

andschwa: Tell that to the compiler that complained.

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
andschwa: I don't see how, there's one argument; hover over the function for its parameter name.

src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
andschwa: Why's that? And also, can we do it separately from fixing compiler warnings?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
```diff
@@ -75,7 +75,7 @@ public async Task<DisconnectResponse> Handle(DisconnectArguments request, Cancel
 
 #pragma warning disable CS4014
             // Trigger the clean up of the debugger. No need to wait for it.
-            Task.Run(_psesDebugServer.OnSessionEnded);
+            Task.Run(_psesDebugServer.OnSessionEnded, cancellationToken);
```
andschwa, line 78: @rjmholt Should this one be explicitly `CancellationToken.None`? Presumably we probably don't want to the debug server's clean up to be cancelled.

andschwa: But that said if it's cancelled here then we don't return the `DisconnectResponse` so I suppose it would get cancelled again?

src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
rjmholt: > Why's that?

With an `async` method, the compiler transforms the method body into an enumerator, and `await` must be called, adding overhead in terms of task waiting and method execution. But in trivial cases, we're actually just creating an async task around awaiting another task. In those cases, the caller is going to have to await anyway, so may as well simply pass the task through. Compare this:

```csharp
        internal Task CreateAsync() =>
            languageServer.SendRequestAsync(
                NewCustomViewRequest.Method,
                new NewCustomViewRequest
                {
                    Id = this.Id,
                    Title = this.Title,
                    ViewType = this.ViewType,
                });
```

> can we do it separately from fixing compiler warnings?

Technically this is also a valid fix for this warning and just as simple

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
rjmholt: > hover over the function for its parameter name.

Can't do that in GitHub. Ideally we wouldn't have boolean parameters, but labelling them definitely helps.

src/PowerShellEditorServices/Services/TextDocument/TokenOperations.cs
```diff
@@ -14,9 +14,6 @@ namespace Microsoft.PowerShell.EditorServices.Services.TextDocument
     /// </summary>
     internal static class TokenOperations
     {
-        // Region kinds to align with VSCode's region kinds
-        private const string RegionKindComment = "comment";
-        private const string RegionKindRegion = "region";
```
andschwa, line 19: These unused fields smell like olllld code to me.

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
rjmholt: I suppose it can't hurt here, but I don't like the idea of misleading code -- the cancellation token definitely has no utility here

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
rjmholt: Yeah I think `None` seems reasonable

src/PowerShellEditorServices/Services/PowerShellContext/PowerShellContextService.cs
andschwa: Already fixed 😉 (Also GitHub is getting better at that! I think the Rich Code Nav is what enables it, and I need to get that setup for this repo)

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
andschwa: We could pass `CancellationToken.None` instead (which says explicitly this won't be cancelled).

src/PowerShellEditorServices.VSCode/CustomViews/CustomViewBase.cs
andschwa: Ah I see.

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
rjmholt: Nah I think what you've got works fine

rjmholt (approved):

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
andschwa: Oops forgot about this 🤷

PR #1542, opened 2021-08-13
Add regression test for `System.Windows.Forms` bug

andschwa:
This PR will fail until we can update OmniSharp and so include https://github.com/OmniSharp/csharp-language-server-protocol/pull/633, which fixes the bug this regression test covers. When that happens, this PR will fix:

* https://github.com/PowerShell/vscode-powershell/issues/3410
* https://github.com/PowerShell/vscode-powershell/issues/3396
* https://github.com/PowerShell/vscode-powershell/issues/3394 (minus a separate issue we're still working out)
* https://github.com/PowerShell/vscode-powershell/issues/3415 (probably anyway, still waiting to hear back from the original author)

### DIFF
--- a/test/PowerShellEditorServices.Test/PowerShellEditorServices.Test.csproj
+++ b/test/PowerShellEditorServices.Test/PowerShellEditorServices.Test.csproj
@@ -34,9 +34,7 @@
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
     <PackageReference Include="Xunit.SkippableFact" Version="1.4.13" />
-    <!-- TODO: Update to 0.19.3 when available -->
-    <!-- <PackageReference Include="OmniSharp.Extensions.LanguageServer" Version="0.19.2" /> -->
-    <ProjectReference Include="..\..\..\csharp-language-server-protocol\src\Server\Server.csproj" />
+    <PackageReference Include="OmniSharp.Extensions.LanguageServer" Version="0.19.3" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


andschwa:
Whelp, it's _passing_, which isn't right!

andschwa:
Ok, I need to actually write this test in the style of:

https://github.com/PowerShell/PowerShellEditorServices/blob/00c78242e17a7265706a0fa07b67770709a670ac/test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs#L168

Because it's the OmniSharp integration that is the regression to assert.

andschwa:
Ok, @rjmholt that build failure _looks_ like the expected hang to me, I only wish I could see that locally.

andschwa:
Look at that! Regression test hung before the update, and doesn't after!!!

test/PowerShellEditorServices.Test/PowerShellEditorServices.Test.csproj
```diff
@@ -34,8 +34,7 @@
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
     <PackageReference Include="Xunit.SkippableFact" Version="1.4.13" />
-    <PackageReference Include="OmniSharp.Extensions.LanguageServer" Version="0.19.2" />
-    <DotNetCliToolReference Include="dotnet-xunit" Version="2.4.0-beta.1.build3958" />
```
andschwa, line 38: Not sure what on earth this `DotNetCliToolReference` was for, it doesn't seem needed any more.

rjmholt (approved):

andschwa:
@rjmholt I dropped this back to OmniSharp v0.19.3 as it's all that's on NuGet. v0.19.4 will come eventually, we'll just let Dependabot update it then.

PR #1546, opened 2021-08-18
Explicitly disable implicit namespace imports

### LINKED ISSUE #1544: Revert language version fix and find another fix?
In #1541 we updated the language version from 9.0 to 10.0 to workaround a bug with .NET 6 Preview 7; however, this isn't the greatest workaround because it requires everyone building PSES to update their dotnet:

```
CSC : error CS1617: Invalid option '10.0' for /langversion. Use '/langversion:?' to list supported values. [/Users/andschwa/src/PowerShellEditorServices/src/PowerShellEditorServices/PowerShellEditorServices.csproj]
```

We didn't realize this because the CI _always_ comes with an up-to-date dotnet.

The simply solution is for PSES devs to just delete `.dotnet` and let the build script install the latest, but perhaps we should fix it another way.

andschwa:
This also reverts us back to language version 9.0.

This reverts commit e29979c369efdcb645a681403ae5d380076a85c8.

Fixes #1544.

### DIFF
--- a/PowerShellEditorServices.Common.props
+++ b/PowerShellEditorServices.Common.props
@@ -4,7 +4,7 @@
     <VersionSuffix></VersionSuffix>
     <Company>Microsoft</Company>
     <Copyright>© Microsoft Corporation.</Copyright>
-    <LangVersion>10.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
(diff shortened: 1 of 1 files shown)


andschwa:
Also see https://github.com/dotnet/sdk/issues/19702

rjmholt (approved):

PR #1547, opened 2021-08-18
Fix creation of `InitialSessionState` to use `CreateDefault2()`

andschwa:
We erroneously changed our code to use `CreateDefault` instead of `CreateDefault2`. There existed old but defunct code for testing the use of `CreateDefault` instead, but it was never intentionally moved into production. The latter function loads only `Microsoft.PowerShell.Core`, leaving us to load the exact modules we intend to load ourselves, which is the safer behavior (and the previous behavior).


rjmholt (approved):

andschwa:
FYI @dkattan. I accidentally changed this when shepherding in #1526

PR #1555, opened 2021-08-19
Fix debugger regression where console needed input to start/continue

andschwa:
These needed to be explicitly `None` to indicate that they should not be cancelled (for various reasons). The actual bug fix was the removal of `shell.Stop()`. Fixes https://github.com/PowerShell/vscode-powershell/issues/3513.


andschwa:
Did not actually fix it 😢

rjmholt (approved):

PR #1569, opened 2021-09-01
Fix bug with `ExecuteScriptWithArgsAsync` when `script` is a command

andschwa:
This fixes https://github.com/PowerShell/vscode-powershell/issues/3537.


andschwa:
@rjmholt I'm just glad that we're starting to get all these expected behaviors (user scenarios) covered with tests 🙃

rjmholt (approved):

PR #1570, opened 2021-09-02
 Send `stopDebugger` notification when appropriate

andschwa:
This is the edge case where the debug server is running because it was started by PowerShell (and not by an LSP event), and we're no longer in the debugger within PowerShell, so since we own the state we need to stop the debug server too.

Along with https://github.com/PowerShell/vscode-powershell/pull/3542, this resolves one of the main issues brought up in https://github.com/PowerShell/vscode-powershell/issues/3522.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,7 @@ indent_size = 4
 trim_trailing_whitespace = true
 csharp_space_before_open_square_brackets = true
 csharp_space_after_keywords_in_control_flow_statements = true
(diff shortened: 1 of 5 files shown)


andschwa:
This needs some tests...

rjmholt (approved):

SeeminglyScience (approved):
LGTM

andschwa:
I'm going to merge so as to get it into a preview while I work on tests.

PR #1571, opened 2021-09-06
Implement `LengthInBufferCells` to fix ANSI formatting

### LINKED ISSUE #840: Table width not calculated correct when using VT100
Table width not calculated correct when using VT100.

https://docs.microsoft.com/en-us/powershell/wmf/5.1/console-improvements states "PowerShell will ignore certain VT100 formatting escape sequences when calculating table widths". This is true in "plain" PowerShell, but not in VSCode.

This example demonstrates the problem:

```
# Create object with VT100 escape sequences
$x = [PSCustomObject]@{
    A = 'NoColor'
    B = "$([char](27))[33;1mColor$([char](27))[0m"
    C = 'NoColor'
}

# Auto size the columns
$x

# Set column width to 10
$x | Format-Table -Property @{L='A';E={$_.A};W=10},@{L='B';E={$_.B};W=10},@{L='C';E={$_.C};W=10}

# Column width is only 4 - this does not work anywhere!
$x | Format-Table -Property @{L='A';E={$_.A};W=4},@{L='B';E={$_.B};W=4},@{L='C';E={$_.C};W=4}
```

This example...:
- Does not work in VSCode - the VT100 sequence characters is included in the character count for table width
- Works in "plain" PowerShell on Windows 10 (not the four wide example)
- Does not work in ISE (but $host.UI.SupportsVirtualTerminal is $false - so VT100 doesn't work at all here)

SeeminglyScience:
Without redirecting this call to `internalRawUI` any formatting with ANSI escapes will likely be truncated in the wrong spot.

## Before fix

image

## After fix

image

Fixes #840

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShellContext/Session/Host/TerminalPSHostRawUserInterface.cs
+++ b/src/PowerShellEditorServices/Services/PowerShellContext/Session/Host/TerminalPSHostRawUserInterface.cs
@@ -265,6 +265,53 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShellContext
             this.internalRawUI.SetBufferContents(origin, contents);
         }
 
+        /// <summary>
+        /// Determines the number of BufferCells a character occupies.
+        /// </summary>
+        /// <param name="source">
+        /// The character whose length we want to know.
+        /// </param>
+        /// <returns>
+        /// The length in buffer cells according to the original host
+        /// implementation for the process.
+        /// </returns>
+        public override int LengthInBufferCells(char source)
+        {
+            return this.internalRawUI.LengthInBufferCells(source);
+        }
+        /// <summary>
+        /// Determines the number of BufferCells a string occupies.
+        /// </summary>
+        /// <param name="source">
+        /// The string whose length we want to know.
+        /// </param>
+        /// <returns>
+        /// The length in buffer cells according to the original host
+        /// implementation for the process.
+        /// </returns>
+        public override int LengthInBufferCells(string source)
+        {
+            return this.internalRawUI.LengthInBufferCells(source);
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


andschwa (approved):
I wonder how we could test this 🤔

andschwa:
Thanks @SeeminglyScience!!!

andschwa:
_Just_ because it relates to ANSI escape stuff...any chance this could fix https://github.com/PowerShell/vscode-powershell/issues/3543? I don't know which escape character is being sent in that instance.

michal037:
> _Just_ because it relates to ANSI escape stuff...any chance this could fix [PowerShell/vscode-powershell#3543](https://github.com/PowerShell/vscode-powershell/issues/3543)? I don't know which escape character is being sent in that instance.

If you have any idea how to read it, I'll send you additional logs.

SeeminglyScience:
> I wonder how we could test this 🤔

Yeah that's a tough one.  Any test I can think of is too arbitrary since we can't really utilize `TerminalHost*` atm.  Even if we could all we could really do is verify that it just proxies the call.  Even trying to test that it returns the expected length seems a bit odd because then we're really testing `ConsoleHost*` instead.

> _Just_ because it relates to ANSI escape stuff...any chance this could fix [PowerShell/vscode-powershell#3543](https://github.com/PowerShell/vscode-powershell/issues/3543)? I don't know which escape character is being sent in that instance.

Probably not.  Kind of wonder if it might be something in [`termios`](https://github.com/SeeminglyScience/UnixConsoleEcho/blob/11a8886efdaf140514ab495266be87ed0f41312c/src/Native/Unix/disable_key_echo.cpp#L39) that needs to be set. A before and after of `stty --all` might be helpful there.

andschwa:
LMAO the body of this PR found an unhandled edge case in the release scripts and caused the number in the changelog for this entry to be `24977523` instead of `840`.

PR #1574, opened 2021-09-14
Fix debugging with PSES pipeline thread changes

rjmholt:


src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -103,26 +118,81 @@ private async Task LaunchScriptAsync(string scriptToLaunch)
...
+                    sb.Append('\'').Append(arg.Replace("'", "''")).Append('\'');
+                }
+                else
+                {
+                    sb.Append(arg);
+                }
+            }
+
+            return new PSCommand().AddScript(sb.ToString());
+        }
```
rjmholt, line 176: See https://github.com/PowerShell/PowerShell/issues/15285

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs
```diff
@@ -48,10 +48,11 @@ internal class DisconnectHandler : IDisconnectHandler
         public async Task<DisconnectResponse> Handle(DisconnectArguments request, CancellationToken cancellationToken)
         {
             _debugEventHandlerService.UnregisterEventHandlers();
-            if (_debugStateService.ExecutionCompleted == false)
+
+            if (!_debugStateService.ExecutionCompleted)
             {
                 _debugStateService.ExecutionCompleted = true;
-                _executionService.CancelCurrentTask();
+                _debugService.Abort();
```
rjmholt, line 55: There's a big question here still about ordering and what should happen.

Currently we call `UnregisterEventHandlers()` and then if execution is still going we call `Abort()` which stops a bunch of debug processes, or at least cancels out of a few of them.

But it's not clear we're handling *all* debug threads properly. We don't have cancellations set up for everything. It's also not clear whether unregistering things and *then* aborting debugging is the right order to do things in.

I'm reluctant to undertake more implementation here since it will only get more complex, but I think we should keep an eye on this. Basically my suspicion is that we might need to lock some things here, evaluate what's going on, and then ensure they've decided to cancel or not before releasing the lock.

src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
```diff
@@ -91,16 +93,25 @@ internal sealed class ExtensionService
...
-            {
-                pwsh.Runspace.SessionStateProxy.PSVariable.Set("psEditor", EditorObject);
-            }, representation: "Set PSEditor",  CancellationToken.None);
+            return ExecutionService.ExecuteDelegateAsync(
+                "Create $psEditorObject",
+                ExecutionOptions.Default,
+                CancellationToken.None,
+                (pwsh, cancellationToken) =>
+                {
+                    pwsh.Runspace.SessionStateProxy.PSVariable.Set("psEditor", EditorObject);
```
rjmholt, line 113: This variable name should be a constant

src/PowerShellEditorServices/Services/PowerShell/Console/ConsoleReadLine.cs
```diff
@@ -5,52 +5,46 @@
...
         #endregion
 
         #region Public Methods
 
-        public Task<string> ReadLineAsync(CancellationToken cancellationToken) => ReadLineAsync(isCommandLine: true, cancellationToken);
-
-        public string ReadLine() => ReadLineAsync(CancellationToken.None).GetAwaiter().GetResult();
-
-        public SecureString ReadSecureLine() => ReadSecureLineAsync(CancellationToken.None).GetAwaiter().GetResult();
+        public string ReadLine(CancellationToken cancellationToken)
```
rjmholt, line 44: The big change in this file is that there's no more async implementation. Since we're always calling this from our dedicated pipeline thread, synchronous is what we want. That also maps better to the underlying implementation. We're not blocking a threadpool worker just to call readline any more.

src/PowerShellEditorServices/Services/PowerShell/Console/ConsoleReadLine.cs
```diff
@@ -158,20 +142,19 @@ public async Task<SecureString> ReadSecureLineAsync(CancellationToken cancellati
...
-            return _executionService.ExecuteDelegateAsync(InvokePSReadLine, representation: "ReadLine", cancellationToken);
+            return ConsoleProxy.ReadKey(intercept: true, cancellationToken);
         }
 
         private string InvokePSReadLine(CancellationToken cancellationToken)
         {
+            cancellationToken.Register(() =>
+            {
+                Debug.WriteLine("PSRL CANCELLED");
+            });
```
rjmholt, line 155: Should take this out -- was for debugging purposes

src/PowerShellEditorServices/Services/PowerShell/Console/ConsoleReplRunner.cs
```diff
@@ -1,303 +0,0 @@
-﻿using Microsoft.Extensions.Logging;
```
rjmholt, line 1: This is the deleted REPL thread implemented

src/PowerShellEditorServices/Services/PowerShell/Console/PSReadLineProxy.cs
```diff
@@ -99,10 +98,10 @@ internal class PSReadLineProxy
         {
             _logger = loggerFactory.CreateLogger<PSReadLineProxy>();
 
-            ReadLine = (Func<Runspace, EngineIntrinsics, CancellationToken, string>)psConsoleReadLine.GetMethod(
+            ReadLine = (Func<Runspace, EngineIntrinsics, CancellationToken, bool?, string>)psConsoleReadLine.GetMethod(
                 ReadLineMethodName,
-                new[] { typeof(Runspace), typeof(EngineIntrinsics), typeof(CancellationToken) })
-                ?.CreateDelegate(typeof(Func<Runspace, EngineIntrinsics, CancellationToken, string>));
+                new[] { typeof(Runspace), typeof(EngineIntrinsics), typeof(CancellationToken), typeof(bool?) })
+                ?.CreateDelegate(typeof(Func<Runspace, EngineIntrinsics, CancellationToken, bool?, string>));
```
rjmholt, line 104: Using the preferred PSRL overload now

src/PowerShellEditorServices/Services/PowerShell/Debugging/PowerShellDebugContext.cs
```diff
@@ -83,22 +79,21 @@ public void StepOver()
...
 
+        // This must be called AFTER the new PowerShell has been pushed
         public void EnterDebugLoop(CancellationToken loopCancellationToken)
         {
-            _debugLoopCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(loopCancellationToken);
             RaiseDebuggerStoppedEvent();
-
         }
 
+        // This must be called BEFORE the debug PowerShell has been popped
```
rjmholt, line 94: I thought about deleting this method, but it might turn out we need to put some state cleanup in here

src/PowerShellEditorServices/Services/PowerShell/Execution/BlockingConcurrentDeque.cs
```diff
@@ -0,0 +1,93 @@
...
+    ///   - The ability to block consumers, so that <see cref="Prepend(T)"/> can also guarantee the state of the consumer
+    /// </summary>
+    /// <typeparam name="T">The type of item held by this collection.</typeparam>
+    /// <remarks>
+    /// The prepend/append semantics of this class depend on the implementation semantics of <see cref="BlockingCollection{T}.TryTakeFromAny(BlockingCollection{T}[], out T)"/>
+    /// and its overloads checking the supplied array in order.
+    /// This behavior is unlikely to change and ensuring its correctness at our layer is likely to be costly.
+    /// See https://stackoverflow.com/q/26472251.
+    /// </remarks>
+    internal class BlockingConcurrentDeque<T>
```
rjmholt, line 20: Hopefully this class is fairly self-explanatory. It's quite useful for our purposes. It doesn't work too hard to reimplement a new concurrent data structure, but instead takes advantage of the `BlockingCollection<T>.TryTakeFromAny()` method that .NET provides.

All the actual cleverness is still handled by [the `BlockingCollection<T>` implementation](https://github.com/dotnet/runtime/blob/27536d59df9b9e7263d48dfe910b48e3792c57a9/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/BlockingCollection.cs#L659-L765), where consumers and producers just hold on to semaphores that track the capacity of the underlying collection.

src/PowerShellEditorServices/Services/PowerShell/Execution/ExecutionOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+            Priority = ExecutionPriority.Normal,
+            MustRunInForeground = false,
+            InterruptCurrentForeground = false,
+        };
+
+        public ExecutionPriority Priority { get; init; }
+
+        public bool MustRunInForeground { get; init; }
+
+        public bool InterruptCurrentForeground { get; init; }
```
rjmholt, line 30: I tried to work out what the simplest and most orthogonal set of options here was, and I'm still not really happy with it, but these two specifically are slightly different. One just prepends the next item, whereas the other will cancel the currently running thing.

src/PowerShellEditorServices/Services/PowerShell/Execution/PipelineThreadExecutor.cs
```diff
@@ -1,284 +0,0 @@
-﻿using Microsoft.Extensions.Logging;
```
rjmholt, line 1: This used to handle the pipeline thread loop. Now moved to the internal host.

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
```diff
@@ -34,19 +32,23 @@ internal class SynchronousPowerShellTask<TResult> : SynchronousTask<IReadOnlyLis
...
         {
             _pwsh = _psesHost.CurrentPowerShell;
 
-            if (_executionOptions.WriteInputToHost)
+            if (PowerShellExecutionOptions.WriteInputToHost)
             {
                 _psesHost.UI.WriteLine(_psCommand.GetInvocationText());
             }
 
-            return !_executionOptions.NoDebuggerExecution && _pwsh.Runspace.Debugger.InBreakpoint
```
rjmholt, line 49: I have no idea why I had this configuration before and I couldn't find any reference to it, so I got rid of it

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
```diff
@@ -32,14 +41,39 @@ protected SynchronousTask(ILogger logger, CancellationToken cancellationToken)
...
+                }
+
+                if (_exceptionInfo is not null)
+                {
+                    _exceptionInfo.Throw();
+                }
+
+                return _result;
+            }
+        }
```
rjmholt, line 60: Some commands we need to evaluate synchronously and this allows us to do that

src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHost.cs
```diff
@@ -1,362 +1,50 @@
-﻿using Microsoft.Extensions.Logging;
```
rjmholt, line 1: The new public host is just a wrapper that allows us to make our host public without exposing anything dangerous to users

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+                await SetInitialWorkingDirectoryAsync(startOptions.InitialWorkingDirectory, CancellationToken.None).ConfigureAwait(false);
+            }
+        }
+
+        public void SetExit()
+        {
+            if (_psFrameStack.Count <= 1)
+            {
+                return;
+            }
```
rjmholt, line 203: You can't exit from the PSIC, since if you do all your LSP services die too

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+
+        public void CancelCurrentTask()
+        {
+            _cancellationContext.CancelCurrentTask();
+        }
+
+        public Task<TResult> ExecuteDelegateAsync<TResult>(
+            string representation,
+            ExecutionOptions executionOptions,
+            CancellationToken cancellationToken,
```
rjmholt, line 250: Should move the cancellation token to the last argument here, even though it's ugly to have the delegate in the middle of the parameters

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+            finally
+            {
+                frame.Dispose();
+            }
+        }
+
+        private void RunTopLevelExecutionLoop()
+        {
+            // Make sure we execute any startup tasks first
+            if (_psFrameStack.Count == 1)
```
rjmholt, line 443: We should take this test out, since the caller performs it

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+            }
+
+            RunExecutionLoop();
+        }
+
+        private void RunDebugExecutionLoop()
+        {
+            try
+            {
+                DebugContext.EnterDebugLoop(CancellationToken.None);
```
rjmholt, line 458: One of a few cancellation tokens that we might want to decide whether to use or get rid of. My philosophy has generally been to pass them up through callers until I can't any more

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+            }
+        }
+
+        private void DoOneRepl(CancellationToken cancellationToken)
+        {
+            if (_skipNextPrompt)
+            {
+                _skipNextPrompt = false;
+                return;
+            }
```
rjmholt, line 495: We need this so that when we run something that cancels the current task to run immediately, the REPL doesn't get in its way

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+        private string GetPrompt(CancellationToken cancellationToken)
+        {
+            var command = new PSCommand().AddCommand("prompt");
+            IReadOnlyList<string> results = InvokePSCommand<string>(command, PowerShellExecutionOptions.Default, cancellationToken);
+            string prompt = results.Count > 0 ? results[0] : DefaultPrompt;
+
+            if (CurrentRunspace.RunspaceOrigin != RunspaceOrigin.Local)
+            {
+                prompt = Runspace.GetRemotePrompt(prompt);
+            }
```
rjmholt, line 542: We steal this from PowerShell to get the nice remote prompt customisation. It's something that PowerShell should just expose publicly

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,778 @@
...
+        private void OnRunspaceStateChanged(object sender, RunspaceStateEventArgs runspaceStateEventArgs)
+        {
+            if (!_shouldExit && !runspaceStateEventArgs.RunspaceStateInfo.IsUsable())
+            {
+                //PopOrReinitializeRunspaceAsync();
+            }
+        }
+
+        /*
+        private void PopOrReinitializeRunspace()
```
rjmholt, line 737: I wrote this as a sort of last ditch attempt to stop the extension from falling over when everything goes wrong. The code currently isn't compatible with the new implementation, so we should review it and decide whether it's worth keeping in any way or should just be gotten rid of

src/PowerShellEditorServices/Services/PowerShell/Host/PowerShellFactory.cs
```diff
@@ -1,120 +0,0 @@
-﻿using System;
```
rjmholt, line 1: This now belongs to the host, since it needed enough internal information from it anyway

src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
```diff
@@ -91,16 +93,25 @@ internal sealed class ExtensionService
         /// </summary>
         /// <param name="editorOperations">An IEditorOperations implementation.</param>
         /// <returns>A Task that can be awaited for completion.</returns>
-        internal async Task InitializeAsync()
+        internal Task InitializeAsync()
         {
+            if (Interlocked.Exchange(ref _initialized, 1) != 0)
```
rjmholt, line 98: We use this pattern in a few places and it might be worth wrapping it in a struct for reuse

src/PowerShellEditorServices/Services/PowerShell/Execution/BlockingConcurrentDeque.cs
PaulHigin: I don't see where this is used anywhere in this PR.
Also, should this not have a Dispose() method to clean up collections, event objects?
It is not clear to me what this is for, it would be nice to have comments illustrating its use.

PaulHigin (changes requested):

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,775 @@
...
+
+        private readonly ILoggerFactory _loggerFactory;
+
+        private readonly ILogger _logger;
+
+        private readonly ILanguageServerFacade _languageServer;
+
+        private readonly HostStartupInfo _hostInfo;
+
+        private readonly BlockingConcurrentDeque<ISynchronousTask> _taskQueue;
```
rjmholt, line 44: @PaulHigin

PaulHigin: Thanks.  I think I need to clone this repo so I can better see how the changes are used.  Searching within this PR instance does not provide enough context.

rjmholt (commented):
Need to add copyright headers

src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility
     using System.IO;
     using System.Management.Automation;
     using System.Runtime.CompilerServices;
+    using UnixConsoleEcho;
```
SeeminglyScience, line 14: Is this used somewhere?

SeeminglyScience:
Looking great!

I did an initial run of feedback.  It's mostly just a few missing `ConfigureAwait`'s but it may be best to just ignore those here and hit it with an analyzer later.

Hopefully the "suggestions" work, haven't tried to add them from the VSCode PR extension yet.

I'll also need to spend some time with the build and see how it all feels, but I'm loving the changes and it all looks very well done. Great work Rob!

SeeminglyScience (commented):
Hmm, it didn't submit the review with the comment. 🤷

andschwa:
> I did an initial run of feedback. It's mostly just a few missing `ConfigureAwait`'s but it may be best to just ignore those here and hit it with an analyzer later.

FYI the analyzer is enabled so if this is somewhat up to date with `master` them missing should cause a compiler failure.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -103,26 +118,81 @@ private async Task LaunchScriptAsync(string scriptToLaunch)
...
+        private PSCommand BuildPSCommandFromArguments(string command, IReadOnlyList<string> arguments)
+        {
+            if (arguments is null
+                || arguments.Count == 0)
+            {
+                return new PSCommand().AddCommand(command);
+            }
+
+            // We are forced to use a hack here so that we can reuse PowerShell's parameter binding
+            var sb = new StringBuilder()
```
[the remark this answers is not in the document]
rjmholt: Ah in this case I prefer `var` for variables. I discussed it with some others on the team like @daxian-dbw and we all agreed that `new()` makes sense for fields but that we prefer `var` for variables. For different reasons, but in my case because the type of the rvalue is the ground truth and belongs on the RHS.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -19,6 +23,14 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 {
     internal class ConfigurationDoneHandler : IConfigurationDoneHandler
     {
+        private readonly PowerShellExecutionOptions s_debuggerExecutionOptions = new()
```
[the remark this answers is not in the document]
rjmholt:
```suggestion
        private static readonly PowerShellExecutionOptions s_debuggerExecutionOptions = new()
```

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -103,26 +118,81 @@ private async Task LaunchScriptAsync(string scriptToLaunch)
...
+                    .ConfigureAwait(false);
             }
 
             _debugAdapterServer.SendNotification(EventNames.Terminated);
         }
+
+        private PSCommand BuildPSCommandFromArguments(string command, IReadOnlyList<string> arguments)
+        {
+            if (arguments is null
+                || arguments.Count == 0)
```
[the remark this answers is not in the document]
rjmholt: I'd like this, but I'm not sure it works with our language version configuration

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
```diff
@@ -0,0 +1,775 @@
...
+    using Microsoft.PowerShell.EditorServices.Utility;
+    using System.IO;
+    using System.Management.Automation;
+    using System.Management.Automation.Runspaces;
+    using System.Reflection;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    internal class InternalHost : PSHost, IHostSupportsInteractiveSession, IRunspaceContext
```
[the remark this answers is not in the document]
rjmholt: Yeah I based it off that one. I'm not attached to the name, it was just a useful distinction.

src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
rjmholt:
```suggestion
```

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
```diff
@@ -68,7 +70,17 @@ private IReadOnlyList<TResult> ExecuteNormally(CancellationToken cancellationTok
...
             {
-                result = _pwsh.InvokeCommand<TResult>(_psCommand);
+                var invocationSettings = new PSInvocationSettings
+                {
+                    AddToHistory = PowerShellExecutionOptions.AddToHistory,
+                };
+
+                if (!PowerShellExecutionOptions.WriteErrorsToHost)
+                {
+                    invocationSettings.ErrorActionPreference = ActionPreference.Stop;
```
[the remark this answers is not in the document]
rjmholt: I need to look into this

src/PowerShellEditorServices/Services/PowerShell/Host/InternalHost.cs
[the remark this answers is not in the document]
rjmholt: Renamed

rjmholt:
I'm going to merge this in now so we can continue work on the main branch, breaking out smaller work items into new PRs.

You can see the changes I've made to address comments in individual commits, which will be preserved in the work branch by a merge commit.

PR #1575, opened 2021-09-29
Fix incorrect ErrorAction values

alexbuzzbee:
I noticed in the logs of my editor's Language Server Protocol client that PowerShellEditorServices was throwing errors from commands with `-ErrorAction Ignore` in them. I checked to be sure, and `Ignore` isn't the correct value; `SilentlyContinue` does that.

I searched the codebase and fixed the three instances of the `Ignore` value I found.


alexbuzzbee:
I'm reasonably confident these changes aren't copyrightable, but I'll get permission from whoever needs to approve this on my company's side and sign it later.

PR #1466, opened 2021-04-29
Remove "All Rights Reserved" from copyright notices

andschwa:
Per Microsoft policy at:
https://docs.opensource.microsoft.com/content/releasing/copyright-headers.html

All copyright notices no longer include "All Rights Reserved" and the comment style has been updated to be consistent with policy.

Resolves #1465 

Also double-checked that all `*.cs` and `*.ps1` files have a copyright notice.

PR #1580, opened 2021-10-12
Fix usings and copyright headers

rjmholt:


rjmholt:
Not for the copyright headers, but I was for the usings

andschwa:
> Not for the copyright headers

Hm, I must be misremembering. It looks like I did it via find-replace (so that's kinda tooling, eh?) in https://github.com/PowerShell/PowerShellEditorServices/pull/1466.

rjmholt:
I remember writing a PS script that did it for some repo...

Wait, just remembered. It's [here](https://github.com/PowerShell/PSArm/blob/master/tools/addCopyrightHeaders.ps1).

PR #1585, opened 2021-10-13
Setting to Disable Pester Code Lens

JustinGrote:
Allows the Pester Codelens to be disabled so as not to confuse users who want to use an alternate Pester Test Runner

Correlated vscode-powershell PR: https://github.com/PowerShell/vscode-powershell/pull/3613

### DIFF
--- a/src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
+++ b/src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
@@ -404,11 +404,13 @@ namespace Microsoft.PowerShell.EditorServices.Services.Configuration
             if (settings is null) {
                 return;
             }
+
             if (this.CodeLens != settings.CodeLens)
             {
                 this.CodeLens = settings.CodeLens;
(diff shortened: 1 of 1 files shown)


JustinGrote:
@andschwa not sure why this is failing the test, something in the way you mock the settings maybe?

andschwa:
> I agree I thought this was super bizarre but with lack of comments as to why it was this way I assumed there was a good reason.

I checked the Git blame and saw no good reason. Looked like the code was added in a community PR and so was probably just checked as thoroughly as we're doing now.

JustinGrote:
@andschwa banged on this for a while, everything looks fine after making some additional adjustments, but the Pester Code lens test fails for me **even on master** with the same error, can you confirm it fails on master for you too?

image

JustinGrote:
@andschwa should pass but it's in "awaiting approval" status for the MacOS build.

JustinGrote:
@rjmholt @andschwa thank you! And thanks for putting up with my low level of C# experience.

PR #1604, opened 2021-10-29
Fix profile loading and `$PROFILE` variable

andschwa:
Fixes issues identified in: https://github.com/PowerShell/vscode-powershell/issues/3650

- [x] Profiles are loaded
- [x] `$PROFILE` is a string with value of path to `CurrentUserCurrentHost`
- [x] All profile paths are available on `$PROFILE` as `PSNote` properties
- [x] `$PROFILE` is initialized _before_ profiles are loaded


andschwa:
Replaces #1602.

andschwa:
Tests will come with `PowerShellContextServiceTests` are brought back online.

andschwa:
TODO: Don't know if this will fix whitespace in path issues, and it definitely doesn't fix the `PrivateData` issue (which is separate).

PR #1605, opened 2021-10-29
Reinstate debug service tests

andschwa:
Needs a rebase. Replaces #1599. Also needs to be finished.


andschwa:
Replaced by #1635

PR #1606, opened 2021-10-29
Add `LengthInBufferCells` back to `EditorServicesConsolePSHostRawUserInterface`

### LINKED ISSUE #1603: Changes from #1571 are missing from preview
I'll submit a PR later, just wanted to get it logged.

andschwa:
Fixes #1603.

These changes were accidentally forgotten in the rewrite and needed to be added back.

@SeeminglyScience Do you know how we could add a regression test for this?

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostRawUserInterface.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostRawUserInterface.cs
@@ -263,6 +263,53 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
             _internalRawUI.SetBufferContents(origin, contents);
         }
 
+        /// <summary>
+        /// Determines the number of BufferCells a character occupies.
+        /// </summary>
+        /// <param name="source">
+        /// The character whose length we want to know.
+        /// </param>
+        /// <returns>
+        /// The length in buffer cells according to the original host
+        /// implementation for the process.
+        /// </returns>
+        public override int LengthInBufferCells(char source)
+        {
+            return _internalRawUI.LengthInBufferCells(source);
+        }
+        /// <summary>
+        /// Determines the number of BufferCells a string occupies.
+        /// </summary>
+        /// <param name="source">
+        /// The string whose length we want to know.
+        /// </param>
+        /// <returns>
+        /// The length in buffer cells according to the original host
+        /// implementation for the process.
+        /// </returns>
+        public override int LengthInBufferCells(string source)
+        {
+            return _internalRawUI.LengthInBufferCells(source);
+        }
+
+        /// <summary>
+        /// Determines the number of BufferCells a substring of a string occupies.
+        /// </summary>
+        /// <param name="source">
+        /// The string whose substring length we want to know.
+        /// </param>
+        /// <param name="offset">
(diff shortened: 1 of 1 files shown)


andschwa:
Fixes https://github.com/PowerShell/PowerShellEditorServices/issues/840 for the preview.

SeeminglyScience:
> @SeeminglyScience Do you know how we could add a regression test for this?

Do we have any tests where we have a `PSHost` that wraps SMA's `ConsoleHost`?

If yes, then running something like this:

```powershell
[pscustomobject]@{ a = "`e[30mtest`e[0m" } |
	Format-Table @{ Width = 4; Expression = 'a' } |
    Out-Default
```

and then reading what is sent to `PSHost.UI.WriteLine`, ensuring it does not include ellipses, would work.  A bit awkward but doable.

andschwa:
> and then reading what is sent to `PSHost.UI.WriteLine`, ensuring it does not include ellipses, would work. A bit awkward but doable.

Any idea how to do this part? I've got a test that starts up everything and runs the command as if via `F8` (so same as typing into the integrated console), but I can only seem to get the results. Not sure if that's going to be different than what PSHost.UI has. Also, I don't see ellipses in the Preview extension when I run this manually, but I do see 'test' being dropped (and it's there in a normal `pwsh` session).

SeeminglyScience:
> Any idea how to do this part? I've got a test that starts up everything and runs the command as if via `F8` (so same as typing into the integrated console), but I can only seem to get the results. Not sure if that's going to be different than what PSHost.UI has.

Yeah so when it hits the UI it'll have already gone through the formatter.  The results are just the raw unformatted objects. For certain things you can also use `Out-String` as that'll also go through the formatter, but it eats ansi escape sequences and doesn't ask the host for buffer cell length.

andschwa:
Hm, well this is as far as I got with the test and I think I'm going to postpone getting a test for this so we can just get the fix in:

```csharp
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

using System;
using System.Management.Automation;
using System.Threading;
using Microsoft.Extensions.Logging.Abstractions;
using Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution;
using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
using Xunit;

namespace Microsoft.PowerShell.EditorServices.Test.Console
{
    public class PSHostTests : IDisposable
    {
        private readonly PsesInternalHost _psesHost;
        public PSHostTests()
        {
            _psesHost = PsesHostFactory.Create(NullLoggerFactory.Instance);
        }

        public void Dispose()
        {
            _psesHost.StopAsync().GetAwaiter().GetResult();
        }

        [Trait("Category", "LengthInBufferCells")]
        [Fact]
        public void HasLengthInBufferCells()
        {
            var script = @"[pscustomobject]@{ a = ""`e[30mtest`e[0m"" } | Format-Table @{ Width = 4; Expression = 'a' } | Out-Default";

            IReadOnlyList<string> results = _psesHost.InvokePSCommand<string>(
                new PSCommand().AddScript(script),
                new PowerShellExecutionOptions { WriteInputToHost = true, WriteOutputToHost = true, ThrowOnError = false, InterruptCurrentForeground = true },
                CancellationToken.None);
            Assert.NotEmpty(results);
        }
    }
}

```

andschwa:
Part of #1677

PR #1609, opened 2021-10-29
Fix: Add special EscapePowershellArgument handling

### LINKED ISSUE #1608: PowershellArgumentNeedsEscaping does not account for a single quoted string
https://github.com/PowerShell/PowerShellEditorServices/blob/8c568bceecfc8f188e6cb53f23b1ed5673bd5af2/src/PowerShellEditorServices/Utility/StringEscaping.cs#L17-L33

Will match true on `'test string'`, it should match false. The \' case needs an additional check to make sure the argument isn't already enclosed in quotes.

JustinGrote:
Resolves #1608

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
@@ -163,16 +163,9 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
             foreach (string arg in arguments)
             {
-                sb.Append(' ');
-
-                if (StringEscaping.PowerShellArgumentNeedsEscaping(arg))
-                {
-                    sb.Append(StringEscaping.SingleQuoteAndEscape(arg));
-                }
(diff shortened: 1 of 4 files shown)


JustinGrote:
@andschwa Well, I didn't quite fix the issue but the tests I added surface the issue...
```
  Error Message:
   Assert.Equal() Failure
           ↓ (pos 1)
Expected: 'C:\&nimals\утка\qu`*ck`?.ps1'
Actual:   '''C:\&nimals\утка\qu`*ck`?.ps1'''
           ↑ (pos 1)
```

JustinGrote:
@andschwa @rjmholt OK, this works for me, but I question the wisdom of being "overly helpful" on the argument escaping, I think it will just set up editors for failure in the future by submitting commands that are syntactically incorrect but we "fix for them" and then when they try to run those commands in some other context, they fail. 

Should we not just return some sort of invalid syntax error instead to the client so that they are able to submit it right the first time?

JustinGrote:
I have a weird uppercase lowercase branch thing happening that is messing with this PR, will resubmit as new PR.

PR #1611, opened 2021-11-01
Improve PowerShell command and argument escaping

### LINKED ISSUE #1608: PowershellArgumentNeedsEscaping does not account for a single quoted string

JustinGrote:
Fixes #1608

@rjmholt please ensure that changing '' to backtick-quote is what was intended, it looked like a bug to me that was causing the triple-quotes.

@SeeminglyScience @andschwa Per my last PR, maybe we are being "overly helpful" with the escaping of the arguments and should instead throw errors to encourage editors to fix their problems at the source? 

For instance in a launch.json, specifying a parameter like this is a clear mistake, but this tool will accept it, escape it, and maybe produce a difficult-to-troubleshoot result:

`args: "-Path","I'putaquoteinthewrongplaceaccidentally"`


andschwa:
Really appreciating the tests

JustinGrote:
I'm gonna rewrite this to be less aggressive on the argument escaping. 

For instance, if someone has a launch.json that is "<command>","arg1","arg2", we should allow expressions to be used.

 e.g. `"Path/to/my/script.ps1" "-Path" "$ENV:MyCustomEnvPath"` will not work with the current escaping I've set up 

I think very minimal escaping should be done for these parameters, it doesn't provide any defense really because I can just code inject into the "command" .ps1 script
If you need the character explicitly you can still backtick it out e.g. `"-Path" "I need a `' literal quote here"`

JustinGrote:
@andschwa the first parameter in the launch config is now always quoted, which will work for both path and command arguments. It passes all tests and all vscode extension tests currently (I removed the pathing escape tests which are no longer relevant).

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -156,23 +156,19 @@ private static PSCommand BuildPSCommandFromArguments(string command, IReadOnlyLi
...
-
-                if (StringEscaping.PowerShellArgumentNeedsEscaping(arg))
-                {
-                    sb.Append(StringEscaping.SingleQuoteAndEscape(arg));
-                }
-                else
-                {
-                    sb.Append(arg);
-                }
+                sb
```
SeeminglyScience, line 169: Nit: consider,

```csharp
sb.Append(' ').Append(ArgumentEscaping.Escape(arg));
```

Personally I like using `text` as the variable for a string builder to enable this:

```csharp
text.Append(' ')
    .Append(ArgumentEscaping.Escape(arg));
```

Very minor.

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (Arg.Contains(" "))
+            {
+                if (Arg.StartsWith("\"") && Arg.EndsWith("\"") || Arg.StartsWith("'") && Arg.EndsWith("'"))
+                {
+                    return Arg;
+                }
+                else
+                {
+                    return "\"" + Arg + "\"";
+                }
```
SeeminglyScience, line 33: This section can be:

```csharp
                if (Arg.StartsWith("\"") && Arg.EndsWith("\"") || Arg.StartsWith("'") && Arg.EndsWith("'"))
                {
                    return Arg;
                }
                
                return "\"" + Arg + "\"";
```

src/PowerShellEditorServices/Utility/PathUtils.cs
```diff
@@ -38,48 +38,22 @@ public static string NormalizePathSeparators(string path)
...
-                {
-                    // Escape '[', ']', '?' and '*' with '`'
-                    case '[':
-                    case ']':
-                    case '*':
-                    case '?':
-                    case '`':
-                        sb.Append('`').Append(curr);
-                        break;
+            var wildcardEscapedPath = WildcardPattern.Escape(path);
```
SeeminglyScience, line 50: May need to manually handle <kbd>`</kbd> still due to PowerShell/PowerShell#16306.  There is a good chance that this seemingly over engineered logic was here due to issue with that API. Unsure when it was added though.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -156,23 +156,19 @@ private static PSCommand BuildPSCommandFromArguments(string command, IReadOnlyLi
                 return new PSCommand().AddCommand(command);
             }
 
-            // We are forced to use a hack here so that we can reuse PowerShell's parameter binding
+            // HACK: We use AddScript instead of AddArgument/AddParameter to reuse Powershell parameter binding logic.
+            // We quote the command parameter so that expressions can still be used in the arguments.
             var sb = new StringBuilder()
                 .Append("& ")
-                .Append(StringEscaping.SingleQuoteAndEscape(command));
+                .Append('"')
```
SeeminglyScience, line 163: Should these be <kbd>'</kbd> or are we purposefully trying to enable variable expansion?

SeeminglyScience (commented):
Lookin' good :) couple of questions and suggestions

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
JustinGrote: Purposefully allowing expressions to close another issue. Since arbitrary code can be specified anyways single quoting doesnt buy us anything.

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
JustinGrote: Yep, that was an auto fill, thanks!

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
JustinGrote: I was matching style of other things in PSES, we can certainly do this if @andschwa doesn't mind.

src/PowerShellEditorServices/Utility/PathUtils.cs
JustinGrote: Probably do need to escape it to handle a situation where it exists in a file name at least at the command level, will probably leave it alone at an argument level

JustinGrote: https://github.com/PowerShell/PowerShellEditorServices/issues/123 is the relevant reasoning, I don't think we need to escape *every* path that comes in, this should be more handled on the output side if required as part of a cmdlet argument, if it's thought the cmdlet will interpret handling characters, because in some cases that may be exactly what the caller wants.

JustinGrote: I'll try to add a test for SetLineBreakpoint which is the only other place this is referenced.

SeeminglyScience: > Probably do need to escape it to handle a situation where it exists in a file name at least at the command level, will probably leave it alone at an argument level

Yeah it gets tricky with no real way to please everyone here.  On one hand you want to enable VSCode variables like `${workspaceFolder}` in which case you can't really escape manually, and on the other hand some folks may want to use expressions.

Like you said though for the command path always escape is a clear winner for sure.

JustinGrote: @SeeminglyScience the compromise for my implementation is that:
1. Command is always wrapped in double quotes because it is used by the call operator, so ${} expressions will still work if you want to pass a file and not a command, the only file that won't work in this case is if the filename contains $(), I think that's reasonable enough to just say "dont name your files this way", should probably have an explicit exception check for this.
2. Argument is only wrapped in quotes if you specify a space. If you really need to be literal or use workspacefolder in an arg and have it be safe, you should wrap it in quotes, a scriptblock, or $(). I think that's a good compromise

JustinGrote: @SeeminglyScience looks like all the debugserver tests were commented out by @rjmholt after his commit and probably need to be rewritten. @andschwa can you come up with a way to get a debugservice instantiated to a point I can run tests against it? Been trying to figure it out and it's too much for my non-C# brain at the moment.

JustinGrote:
I built a .vsix of this and it is working for pester test extension, native pester codelens, and all my debug launch.json configurations. It is good as is unless we want to adjust for @SeeminglyScience's nit, and cover the debugservice tests in a separate PR since looks like they all need to be redone anyways.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
andschwa: Totally fine. I'm not super aggressive about style, I think anytime a style question is in play it should be addressed by tooling rather than back and forth on PRs. Do what you think is readable 😄

src/PowerShellEditorServices/Utility/PathUtils.cs
andschwa: > Yeah it gets tricky with no real way to please everyone here. On one hand you want to enable VSCode variables like ${workspaceFolder} in which case you can't really escape manually, and on the other hand some folks may want to use expressions.

Wouldn't this be interpolated _before_ it gets to the LSP server? I could be wrong, I just assumed that the Code client would interpolate their own variable expressions, and then send the full string over.

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,37 @@
+using System.Text;
```
andschwa, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

using System.Text;
```

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Management.Automation.Language;
+
+namespace Microsoft.PowerShell.EditorServices.Utility
+{
+    internal static class ArgumentEscaping
+    {
+        /// <summary>
+        /// Escape a powershell argument while still making it evaluatable in AddScript.
```
andschwa, line 9:
```suggestion
        /// Escape a PowerShell argument while still making it able to be evaluated in AddScript.
```

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+        ///
+        /// NOTE: This does not "santize" parameters, e.g. a pipe in one argument might affect another argument.
```
andschwa, line 11:
```suggestion
        /// NOTE: This does not "sanitize" parameters, e.g., a pipe in one argument might affect another argument.
```

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// This is intentional to give flexibility to specifying arguments.
+        /// It also does not try to fix invalid PowerShell syntax, e.g. a single quote in a string literal.
```
andschwa, line 13:
```suggestion
        /// It also does not try to fix invalid PowerShell syntax, e.g., a single quote in a string literal.
```

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public static string Escape(string Arg)
+        {
+            // if argument is a scriptblock return as-is
+            if (Arg.StartsWith("{") && Arg.EndsWith("}"))
```
andschwa, line 18: @JustinGrote Should we strip whitespace first? In case it's ` { ... } ` or some other variation with (pointless) spaces before/after.

test/PowerShellEditorServices.Test/Utility/ArgumentEscapingTests.cs
```diff
@@ -0,0 +1,66 @@
+using Xunit;
```
andschwa, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

using Xunit;
```

andschwa (approved):

src/PowerShellEditorServices/Utility/PathUtils.cs
JustinGrote: > > Yeah it gets tricky with no real way to please everyone here. On one hand you want to enable VSCode variables like ${workspaceFolder} in which case you can't really escape manually, and on the other hand some folks may want to use expressions.
> 
> Wouldn't this be interpolated _before_ it gets to the LSP server? I could be wrong, I just assumed that the Code client would interpolate their own variable expressions, and then send the full string over.

They do, the problem is that you don't know that they are a path or an interpolation once they arrive to the LSP. I think this implemention is a good compromise.

src/PowerShellEditorServices/Utility/ArgumentUtils.cs
JustinGrote: > @JustinGrote Should we strip whitespace first? In case it's `{ ... }` or some other variation with (pointless) spaces before/after.

I think this is another case of "overly helpful", if you specified your debug config and left a space at the beginning or the end before your scriptblock, I have to think you may have had an intention there.

andschwa:
Coming back to to this @JustinGrote and deciding we need to not quote escape args with spaces in them.

PR #1616, opened 2021-11-02
Require .NET 6 SDK and remove other unnecessary SDKs

JustinGrote:
This will make bootstrapping more efficient and avoid potential development environment differences.


andschwa:
We need 3.1 and 5 to produce the cross-platform builds necessary for supporting PowerShell 7.1, 7.0, and 5.1. It sucks, I know, but it is what it is 😩

andschwa:
Hm I see in the Discord you want to include just the various _runtimes_ without including their associated SDKs. Not a bad idea...

JustinGrote:
> Hm I see in the Discord you want to include just the various _runtimes_ without including their associated SDKs. Not a bad idea...

Correct, the SDK 6 can produce all the different assemblies via multitargeting just fine, the only part where the runtimes are actually required is testing.

andschwa:
I think this is cool. I'll make it happen, just a minute.

andschwa:
Ok I made this happen but had to tweak it quite a bit!

PR #1628, opened 2021-11-10
Fix build for Apple M1 when running PowerShell 7.2 (arm64)

andschwa:
The previous fix assumed PowerShell was always running in an emulated mode (i.e. running PowerShell (x64) but on an Apple M1). Now that PowerShell 7.2 has a native build for Apple M1 (arm64) the check needs to handle both cases.

We must also skip asking `dotnet-install` to install the 3.1 or 5.0 .NET Runtimes on an Apple M1 because they don't exist, but their installer tries to get them anyway. This was a non-breaking bug previously where the macOS x64 runtimes would be installed, but not used.

### DIFF
--- a/PowerShellEditorServices.build.ps1
+++ b/PowerShellEditorServices.build.ps1
@@ -29,7 +29,9 @@ $script:dotnetTestArgs = @(
 )
 
 $script:IsNix = $IsLinux -or $IsMacOS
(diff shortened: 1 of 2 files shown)


andschwa:
Replaces #1625.

StevenBucher98 (approved):

PR #1629, opened 2021-11-15
Fix null pathways when a naked throw is issued at the PSIC console

JustinGrote:
Fixes https://github.com/PowerShell/vscode-powershell/issues/3677

### DIFF
--- a/src/PowerShellEditorServices/Utility/PSCommandExtensions.cs
+++ b/src/PowerShellEditorServices/Utility/PSCommandExtensions.cs
@@ -55,10 +55,13 @@ namespace Microsoft.PowerShell.EditorServices.Utility
 
         public static PSCommand MergePipelineResults(this PSCommand psCommand)
         {
-            // We need to do merge errors and output before rendering with an Out- cmdlet
-            Command lastCommand = psCommand.Commands[psCommand.Commands.Count - 1];
-            lastCommand.MergeMyResults(PipelineResultTypes.Error, PipelineResultTypes.Output);
-            lastCommand.MergeMyResults(PipelineResultTypes.Information, PipelineResultTypes.Output);
+            if (psCommand.Commands.Count > 0)
+            {
+                // We need to do merge errors and output before rendering with an Out- cmdlet
+                Command lastCommand = psCommand.Commands[psCommand.Commands.Count - 1];
+                lastCommand.MergeMyResults(PipelineResultTypes.Error, PipelineResultTypes.Output);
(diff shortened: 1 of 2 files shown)


JustinGrote:
Works as expected now, the color styling is another issue
image

src/PowerShellEditorServices/Utility/PSCommandExtensions.cs
```diff
@@ -55,10 +55,13 @@ public static PSCommand AddDebugOutputCommand(this PSCommand psCommand)
...
-            Command lastCommand = psCommand.Commands[psCommand.Commands.Count - 1];
-            lastCommand.MergeMyResults(PipelineResultTypes.Error, PipelineResultTypes.Output);
-            lastCommand.MergeMyResults(PipelineResultTypes.Information, PipelineResultTypes.Output);
+            if (psCommand.Commands.Count > 0)
+            {
+                // We need to do merge errors and output before rendering with an Out- cmdlet
+                Command lastCommand = psCommand.Commands[psCommand.Commands.Count - 1];
+                lastCommand.MergeMyResults(PipelineResultTypes.Error, PipelineResultTypes.Output);
+                lastCommand.MergeMyResults(PipelineResultTypes.Information, PipelineResultTypes.Output);
+            }
```
SeeminglyScience, line 64: I'm kinda curious what circumstance we're calling this method without any commands added though 🤔

JustinGrote: With `throw`, there was no "previous command output" and that's where the null was coming from.

JustinGrote: https://github.com/JustinGrote/PowerShellEditorServices/blob/ceb20e8f47754f5c829c6f4ae9c102241f7e59da/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs#L106-L109

andschwa:
@JustinGrote Did you mean to close this?

JustinGrote:
@andschwa I probably just deleted the branch during a cleanup accidentally, should still be valid.

andschwa (approved):
Heh, I thought there would be more too this for some reason. Great and simple fix!

andschwa:
@JustinGrote I'm going to see if I can get the old PowerShell context service tests online and add a regression test for this.

andschwa:
Ok, I enabled those tests! After they're merged I'll try to write a regression test for this (or you are free to take a stab at it) and then merge it.

andschwa:
Closing since I got this change in!

PR #1634, opened 2021-11-23
Display `IEnumerables` and `IDictionaries` in debugger prettily (with "Raw View" available)

### LINKED ISSUE #1633: IEnumerable/IDictionary Display should be simplified
We can simplify the display of standard array and dictionary objects by hiding a lot of the boilerplate, either directly, or into a separate "Detailed Properties" expandable. 99% of the time the user does not need this information and it just serves as noise.

If additional properties are added via inheritance, then those should be shown.

image

JustinGrote:
Resolves #1633

Adds a "Raw View" expandable to the variable pane much like the C# variable provider to hide "noisy" properties of IEnumerable/IDictionary.
image


src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -356,10 +357,23 @@ private VariableDetails[] GetChildren(object obj, ILogger logger)
...
             Type objectType = obj.GetType();
+
+            // For certain array or dictionary types, we want to hide additional properties under a "raw view" header
+            // to reduce noise. This is inspired by the C# vscode extension.
+            if (!noRawView &&
+                (
+                    obj is IEnumerable
+                    || obj is IDictionary
+                )
+            )
```
SeeminglyScience, line 371: This can just be `if (!noRawView && obj is IEnumerable)`

`IDictionary` implements `IEnumerable` so it'll never hit that branch and be true. Actually the compiler just doesn't emit the check, but still

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -414,4 +428,19 @@ public override string ToString()
             }
         }
     }
+
+    /// <summary>
+    /// A VariableDetails that only returns the raw view properties of the object, rather than its values.
+    /// </summary>
+    internal class VariableDetailsRawView : VariableDetails
+    {
+        private const string RawViewName = "Raw View";
```
SeeminglyScience, line 437:
```suggestion
        private const string RawViewName = "Raw View";

```

(new line)

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -414,4 +428,19 @@ public override string ToString()
...
+        public VariableDetailsRawView(object value) : base(RawViewName, value) { }
```
SeeminglyScience, line 438:
```suggestion
        public VariableDetailsRawView(object value) : base(RawViewName, value)
        {
        }

```

I personally don't mind how you have it, but once we set up a formatter they tend to complain. Also new line to separate

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -27,7 +28,7 @@ internal class VariableDetails : VariableDetailsBase
         /// </summary>
         public const string DollarPrefix = "$";
 
-        private object valueObject;
+        protected object valueObject;
```
SeeminglyScience, line 31: If it's going to be used outside of the declaring class it's best to make this a property instead of a field. Preferably a `{ get; }` if you can init it in the ctor, otherwise `{ get; private set; }`

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Linq.Expressions;
```
SeeminglyScience, line 9:
```suggestion
```

Accidental add I think. Ignore if not

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -414,4 +428,19 @@ public override string ToString()
             }
         }
     }
+
+    /// <summary>
+    /// A VariableDetails that only returns the raw view properties of the object, rather than its values.
+    /// </summary>
+    internal class VariableDetailsRawView : VariableDetails
```
SeeminglyScience, line 435:
```suggestion
    internal sealed class VariableDetailsRawView : VariableDetails
```

Good to seal internals when you don't need to subclass it. Can always change it later if needed, but sealing leads to better code gen.

SeeminglyScience (commented):
Awesome :) thanks for this! In your screen shot it shows a `$true` without color. Is that new with this change? or just something separate we need an issue for?

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
JustinGrote: Good catch

JustinGrote:
> Awesome :) thanks for this! In your screen shot it shows a `$true` without color. Is that new with this change? or just something separate we need an issue for?

Pretty sure that's separate, let me check without

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
JustinGrote: Implemented 8179caa03cedbbf4600397808f87202ca23779af

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
JustinGrote: Implemented 95542b32d6d7439df98c604fe7673caf0852814f

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
SeeminglyScience: Hmm looks like it's still a field. You'd want this:

```csharp
protected object ValueObject { get; }
```

Fields are usually kept to private to the class implementation detail only. Properties make it easier to change how it's accessed later. Not a huge deal since it's all internal anyway but it can still make refactoring a pain later on.

JustinGrote:
> > Awesome :) thanks for this! In your screen shot it shows a `$true` without color. Is that new with this change? or just something separate we need an issue for?
> 
> Pretty sure that's separate, let me check without

Yep, it is, opened issue https://github.com/PowerShell/vscode-powershell/issues/3703

SeeminglyScience (approved):
LGTM, thanks again!

andschwa:
Ditto https://github.com/PowerShell/PowerShellEditorServices/pull/1636#issuecomment-983048305

andschwa:
What tests can we add to cover this @JustinGrote?

JustinGrote:
I would say test both an ienumerable and an idictionary and verify that:
1. The object display contains all necessary properties
2. The "raw view" variable display contains what you would expect to see from an idictionary/ienumerable
3. Test a derived idictionary with additional properties/methods and make sure they show up similarly under "Raw view" instead of the direct collection

JustinGrote:
Other test length updates were because their "view" has changed to summarize under the "Raw View" header

JustinGrote:
@SeeminglyScience please re-review, code has changed a little to clean up the "Raw View" header, but this is mostly net-new tests. Thanks :)

andschwa:
@JustinGrote I just added a commit with a couple cleanups. Namely I took out the new `AssertDebuggerStoppedCommand` and just made the existing `AssertDebuggerStopped` optionally accept a `CommandBreakpointDetails`, and I used the `GetVariables(scope)` helper I'd previously introduced to shorten your (lovely!) new tests.

andschwa (approved):
Awesome work Justin!

JustinGrote:
@andschwa thanks, I actually originally wrote it with the optional parameters but I wasn't sure if that was muddling up the original AssertDebuggerStopped with too many code paths, hence the separate function for clarity. All good though!

andschwa:
It made sense to me cause they're all doing the same thing: asserting the debugger stopped, and then optionally asserting it stopped on the right line (or command).

PR #1635, opened 2021-11-23
Re-enable `DebugServiceTests` suite

andschwa:
Please see commits for work done.

### DIFF
--- a/PowerShellEditorServices.build.ps1
+++ b/PowerShellEditorServices.build.ps1
@@ -238,6 +238,10 @@ task TestServer TestServerWinPS,TestServerPS7,TestServerPS72
 
 task TestServerWinPS -If (-not $script:IsNix) {
     Set-Location .\test\PowerShellEditorServices.Test\
+    # TODO: See https://github.com/dotnet/sdk/issues/18353 for x64 test host
+    # that is debuggable! If architecture is added, the assembly path gets an
+    # additional folder, necesstiating fixes to find the commands definition
+    # file and test files.
     exec { & $script:dotnetExe $script:dotnetTestArgs $script:NetRuntime.Desktop }
 }
 
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using Microsoft.Extensions.Logging;
@@ -70,33 +70,30 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
             if (_debugStateService.OwnsEditorSession)
             {
-                // If this is a debug-only session, we need to start
-                // the command loop manually
+                // TODO: If this is a debug-only session, we need to start the command loop manually
+                //
                 //_powerShellContextService.ConsoleReader.StartCommandLoop();
             }
 
             if (!string.IsNullOrEmpty(_debugStateService.ScriptToLaunch))
             {
-                LaunchScriptAsync(_debugStateService.ScriptToLaunch)
-                    .HandleErrorsAsync(_logger);
+                LaunchScriptAsync(_debugStateService.ScriptToLaunch).HandleErrorsAsync(_logger);
             }
 
-            if (_debugStateService.IsInteractiveDebugSession)
+            if (_debugStateService.IsInteractiveDebugSession && _debugService.IsDebuggerStopped)
             {
-                if (_debugService.IsDebuggerStopped)
+                if (_debugService.CurrentDebuggerStoppedEventArgs is not null)
                 {
-                    if (_debugService.CurrentDebuggerStoppedEventArgs != null)
-                    {
-                        // If this is an interactive session and there's a pending breakpoint,
-                        // send that information along to the debugger client
-                        _debugEventHandlerService.TriggerDebuggerStopped(_debugService.CurrentDebuggerStoppedEventArgs);
-                    }
-                    else
-                    {
-                        // If this is an interactive session and there's a pending breakpoint that has not been propagated through
-                        // the debug service, fire the debug service's OnDebuggerStop event.
-                        _debugService.OnDebuggerStopAsync(null, _debugContext.LastStopEventArgs);
-                    }
+                    // If this is an interactive session and there's a pending breakpoint, send that
+                    // information along to the debugger client.
+                    _debugEventHandlerService.TriggerDebuggerStopped(_debugService.CurrentDebuggerStoppedEventArgs);
+                }
+                else
+                {
+                    // If this is an interactive session and there's a pending breakpoint that has
+                    // not been propagated through the debug service, fire the debug service's
+                    // OnDebuggerStop event.
+                    _debugService.OnDebuggerStopAsync(null, _debugContext.LastStopEventArgs);
                 }
             }
 
@@ -119,6 +116,8 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
                     ScriptBlockAst ast = Parser.ParseInput(untitledScript.Contents, untitledScript.DocumentUri.ToString(), out Token[] tokens, out ParseError[] errors);
 
                     // This seems to be the simplest way to invoke a script block (which contains breakpoint information) via the PowerShell API.
+                    //
+                    // TODO: Fix this so the added script doesn't show up.
                     var cmd = new PSCommand().AddScript(". $args[0]").AddArgument(ast.GetScriptBlock());
                     await _executionService
                         .ExecutePSCommandAsync<object>(cmd, CancellationToken.None, s_debuggerExecutionOptions)
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using Microsoft.Extensions.Logging;
@@ -13,10 +13,8 @@ using Microsoft.PowerShell.EditorServices.Utility;
 using OmniSharp.Extensions.DebugAdapter.Protocol.Events;
 using OmniSharp.Extensions.DebugAdapter.Protocol.Requests;
 using OmniSharp.Extensions.DebugAdapter.Protocol.Server;
-using System.Collections.Generic;
 using System.Management.Automation;
 using System.Management.Automation.Language;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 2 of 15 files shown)


JustinGrote:
Pushed a commit to the wrong branch, just reverted it, not a good look for my first day on the job :P

andschwa:
@JustinGrote Progress! The first two tests are passing.

JustinGrote:
@andschwa great! Once this gets merged I'll rebase my existing PRs and try to add tests accordingly.

andschwa:
@JustinGrote Hm, everything but the four that I've skipped (and they're unfortunately indicative of real broken things) pass on my Mac...but not in Windows CI 😢

andschwa:
The common failure is around the event not getting added to the debugger stopped queue, which is what breaks the first skipped test and I've been debugging. I believe there's a race condition in the debugger or its handlers. Working on finding it. @SeeminglyScience if you have any ideas, let me know!

andschwa:
@SeeminglyScience @JustinGrote I found the deadlock! In the tests, any code that calls into `debugService` can deadlock. On PS Core this only appeared with `debugService.Break()`: calling that API essentially requires the PSES pipeline thread to produce some events which eventually push to our debugger stopped queue, and calling it _puts us_ into the pipeline thread. Wrapping it in a `Task.Run()` is a viable solution because it lets the test code return to a non pipeline thread. _On Windows PowerShell_ the problem becomes far more evident because the PS Core breakpoint APIs are not available, and the "legacy" fallback code then runs into the same problem, so `SetLine/CommandBreakpointAsync` tosses us onto the pipeline thread. These are async tasks and are fixed by using `ConfigureAwait(true)`, which says "return us to the captured thread context." This actually makes sense given that the tests are essentially UI code, I may just replace all `ConfigureAwait(false)` in _test_ code with `true`.

Unfortunately there's still a large hill (but not a mountain any more) of work here: I need to clean all of this up for one, and for two, the set variable logic is broken and the associated tests currently skipped. I might punt the latter.

JustinGrote:
@andschwa look at you with your green checkmarks :). Thanks for what I'm sure was a very perplexing effort.

PowerShellEditorServices.build.ps1
```diff
@@ -238,6 +238,10 @@ task TestServer TestServerWinPS,TestServerPS7,TestServerPS72
 
 task TestServerWinPS -If (-not $script:IsNix) {
     Set-Location .\test\PowerShellEditorServices.Test\
+    # TODO: See https://github.com/dotnet/sdk/issues/18353 for x64 test host
+    # that is debuggable! If architecture is added, the assembly path gets an
+    # additional folder, necesstiating fixes to find the commands definition
+    # file and test files.
```
andschwa, line 244: Fix Windows PowerShell test host so that it's debuggable.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -121,6 +116,8 @@ private async Task LaunchScriptAsync(string scriptToLaunch)
                     ScriptBlockAst ast = Parser.ParseInput(untitledScript.Contents, untitledScript.DocumentUri.ToString(), out Token[] tokens, out ParseError[] errors);
 
                     // This seems to be the simplest way to invoke a script block (which contains breakpoint information) via the PowerShell API.
+                    //
+                    // TODO: Fix this so the added script doesn't show up.
                     var cmd = new PSCommand().AddScript(". $args[0]").AddArgument(ast.GetScriptBlock());
                     await _executionService
                         .ExecutePSCommandAsync<object>(cmd, CancellationToken.None, s_debuggerExecutionOptions)
```
andschwa, line 123: This is the cause of https://github.com/PowerShell/vscode-powershell/issues/3666

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -245,6 +245,11 @@ public void TriggerShutdown()
             if (Interlocked.Exchange(ref _shuttingDown, 1) == 0)
             {
                 _cancellationContext.CancelCurrentTaskStack();
+                // NOTE: This is mostly for sanity's sake, as during debugging of tests I became
+                // concerned that the repeated creation and disposal of the host was not also
+                // joining and disposing this thread, leaving the tests in a weird state. Because
+                // the tasks have been canceled, we should be able to join this thread.
+                _pipelineThread.Join();
```
andschwa, line 252: I have questions around this!

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -532,14 +539,7 @@ private void RunTopLevelExecutionLoop()
...
                 _started.TrySetResult(true);
 
-                if (_hostInfo.ConsoleReplEnabled)
-                {
-                    RunExecutionLoop();
-                }
-                else
-                {
-                    RunNoPromptExecutionLoop();
-                }
```
andschwa, line 542: And also around this!

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -2,101 +2,141 @@
...
+            // Load the test debug files
+            workspace = new WorkspaceService(NullLoggerFactory.Instance);
+            debugScriptFile = GetDebugScript("DebugTest.ps1");
+            variableScriptFile = GetDebugScript("VariableTest.ps1");
+        }
+
+        public void Dispose()
+        {
+            debugService.Abort();
+            psesHost.StopAsync().GetAwaiter().GetResult();
```
andschwa, line 64: `.Wait()`

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -2,101 +2,141 @@
...
+            GC.SuppressFinalize(this);
```
andschwa, line 65: Try `sealed` class to see if warning goes away.

daxian-dbw (approved):
LGTM. @andschwa walked me through the changes, thanks Andy!

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
SeeminglyScience: Note that `Wait()` will wrap the exception in an `AggregateException` and `GetAwaiter().GetResult()` does not.

SeeminglyScience (commented):
Awesome work! A couple of mostly small things

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
andschwa: We're in a `Dispose` so that's fine!

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
andschwa: @daxian-dbw Ah ha, xUnit will not allow the unit tests' classes to be `sealed`, they _must_ be `public` and you can't do `public sealed` (or `sealed public`). Oh well.

src/PowerShellEditorServices/Utility/PSCommandExtensions.cs
```diff
@@ -119,5 +126,31 @@ private static StringBuilder AddCommandText(this StringBuilder sb, Command comma
...
+        {
+            if (arguments is null or { Count: 0 })
+            {
+                return new PSCommand().AddCommand(command);
+            }
+
+            // HACK: We use AddScript instead of AddArgument/AddParameter to reuse Powershell parameter binding logic.
+            // We quote the command parameter so that expressions can still be used in the arguments.
+            var sb = new StringBuilder()
+                .Append('&')
```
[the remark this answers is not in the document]
andschwa: Yup! Doing it in a separate PR.

src/PowerShellEditorServices/Utility/PSCommandExtensions.cs
```diff
@@ -11,11 +12,11 @@
 
 namespace Microsoft.PowerShell.EditorServices.Utility
 {
-    internal static class PSCommandExtensions
+    internal static class PSCommandHelpers
```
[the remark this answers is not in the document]
andschwa: 🤷

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
andschwa: @SeeminglyScience Also saw this.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
andschwa: @dkattan I'm removing this `RunNoPromptExecutionLoop()` FYI, but like `RunExecutionLoop` works without an integrated console just fine based on a) its logic that handles no console being available and b) these unit tests that work without a console using it.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
SeeminglyScience: Yeah that'll do it for sure. I believe there's supposed to be some option on `ExecutePSCommandAsync` or it's options class that allows you to specify a "display/history" string for the command.

andschwa:
@SeeminglyScience For some reason I can't reply to this:

https://github.com/PowerShell/PowerShellEditorServices/pull/1635#discussion_r771960126

I get that's what it's trying to do, but I'm having trouble following the logic. The existing logic runs into a null dereference issue, because first a frame is popped, and then we do a peek on the frame stack. If this logic is _correct_ it needs to be adjusted for the case where the stack is empty after that pop. The current change in this PR assumes that the existing logic was an oversight and that it was intending to compare to the frame being popped. What do you think? Dongbo and I had decided just to test one, see what happens, and change it if needed 😅

andschwa:
@daxian-dbw Upon examination, yes your suspicion is correct: xUnit adds its own synchronization context even if parallelism is disabled, since we're using `async` test methods:

https://github.com/xunit/xunit/blob/7f5286f6587fa0b600c9bdebef949dff5301c83e/src/xunit.v3.core/Sdk/AsyncTestSyncContext.cs

SeeminglyScience:
> If this logic is correct it needs to be adjusted for the case where the stack is empty after that pop. The current change in this PR assumes that the existing logic was an oversight and that it was intending to compare to the frame being popped. What do you think?

Is it possible that `PopPowerShell` is being called too many times? Either way I think this change is going to make the runspace changed determination lag behind by one frame if that makes sense.

Unless I'm missing something, right now it's checking if the current frame is on the current runspace, but that'll always be the case. I think the current fix is maybe hiding a different bug.

andschwa:
> Is it possible that `PopPowerShell` is being called too many times? Either way I think this change is going to make the runspace changed determination lag behind by one frame if that makes sense.

I don't think so. The case where it's an empty stack is an easy one: when the _last_ PowerShell frame is being popped, the stack is now empty. `CurrentPowerShell.Runspace` then does `_psFrameStack.Peek()` on said empty stack.

> Unless I'm missing something, right now it's checking if the current frame is on the current runspace, but that'll always be the case. I think the current fix is maybe hiding a different bug.

This I agree with, it doesn't seem right to me now. Potential fix:

```csharp
if (_psFrameStack.Count != 0 && previousRunspaceFrame.Runspace != _psFrameStack.Peek().PowerShell.Runspace)
```

SeeminglyScience:
> This I agree with, it doesn't seem right to me now. Potential fix:

Something still seems a little off. I think there should always be something on the stack unless I'm remembering wrong. Is it possible this only happens at tear down? And just a lot more noticeable in tests since it's happening more often?

If that's the case then yeah that seems like a good fix. If it's not only during tear down then I'm a little confused how it can function after `PopPowerShell` exits with an empty stack.

andschwa:
@SeeminglyScience I stepped through it a few more times and I'm convinced the null-dereference is happening during cleanup, and so was an overlooked edge case. Here's what I'm about to rebase into this PR:

```csharp
        private void PopPowerShell(RunspaceChangeAction runspaceChangeAction = RunspaceChangeAction.Exit)
        {
            _shouldExit = false;
            PowerShellContextFrame frame = _psFrameStack.Pop();
            try
            {
                // If we're changing runspace, make sure we move the handlers over. If we just
                // popped the last frame, then we're exiting and should pop the runspace too.
                if (_psFrameStack.Count == 0
                    || _runspaceStack.Peek().Runspace != _psFrameStack.Peek().PowerShell.Runspace)
                {
                    RunspaceFrame previousRunspaceFrame = _runspaceStack.Pop();
                    RemoveRunspaceEventHandlers(previousRunspaceFrame.Runspace);

                    // If there is still a runspace on the stack, then we need to re-register the
                    // handlers. Otherwise we're exiting and so don't need to run 'RunspaceChanged'.
                    if (_runspaceStack.Count > 0)
                    {
                        RunspaceFrame newRunspaceFrame = _runspaceStack.Peek();
                        AddRunspaceEventHandlers(newRunspaceFrame.Runspace);
                        RunspaceChanged?.Invoke(
                            this,
                            new RunspaceChangedEventArgs(
                                runspaceChangeAction,
                                previousRunspaceFrame.RunspaceInfo,
                                newRunspaceFrame.RunspaceInfo));
                    }
                }
            }
            finally
            {
                frame.Dispose();
            }
        }
```

PR #1636, opened 2021-11-24
Add magic value formatting for variable booleans

JustinGrote:
This PR fixes https://github.com/PowerShell/vscode-powershell/issues/3703

$true and $false now highlight correctly regardless of vscode bug, and number/string type is still preserved.
image

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
@@ -168,12 +168,22 @@ namespace Microsoft.PowerShell.EditorServices.Services.DebugAdapter
             }
 
             Type objType = value.GetType();
-            typeName = $"[{objType.FullName}]";
+
+            // This is the type format PowerShell users expect and will appear when you hover a variable name
+            typeName = '[' + objType.FullName + ']';
 
             if (value is bool)
             {
                 // Set to identifier recognized by PowerShell to make setVariable from the debug UI more natural.
                 valueString = (bool) value ? "$true" : "$false";
+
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
```diff
@@ -168,12 +168,22 @@ private static string GetValueStringAndType(object value, bool isExpandable, out
             }
 
             Type objType = value.GetType();
-            typeName = $"[{objType.FullName}]";
+
+            // This is the type format PowerShell users expect and will appear when you hover a variable name
+            typeName = '[' + value.GetType().FullName + ']';
```
SeeminglyScience, line 173:
```suggestion
            typeName = '[' + objType.FullName + ']';
```

We may actually even want to use `ToStringCodeMethods(Type)` here. Handles type accelerators and not assembly qualifying generic arguments. Probably out of scope though

SeeminglyScience (approved):
LGTM!

src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs
JustinGrote: Thanks, I originally removed objType but then realized why it needed to be there but didn't go back and fix it here.

andschwa:
I really like this, but I want to get the debug service tests re-enabled first. At that point, this may break (but the break will be "update tests' expectations." Sound good? I'm making progress on the tests!

JustinGrote:
@andschwa thanks for fixing the debugger, I've added related tests.

andschwa (approved):

PR #1638, opened 2021-11-30
Fix `BundledModulePath` and PSReadLine loading (redux)

andschwa:
This redoes prior work that was lost during the rewrite. Specifically
this actually respects the user configuration of `BundledModulePath`
(also used by unit tests to provide compatibililty with xUnit), and
forces the use of only our bundled PSReadLine dependency.

Essentially this redoes #1514 and #1522.


andschwa:
@SeeminglyScience and @JustinGrote I discovered this while working on #1605. Specifically when debugging the first test I noticed that `$Error` had a real error in it, and that it was a failure to load the base commands module. Upon inspection, all this hard work had been dropped during the rewrite.

daxian-dbw (approved):
LGTM

andschwa:
Part of #1677.

PR #1641, opened 2021-12-05
Added check to see if PSModulePath already contained BundledModulePath

dkattan:
This prevents PSModulePath from having BundledModulePath added each time PSES is started and shutdown in the same process.


andschwa (approved):
Great fix, thanks!

PR #1656, opened 2022-01-06
Resurrect support to resolve aliased references

andschwa:
This work was spawned by a skipped unit test that "indicated a bug in the product code." The bug was that the support was removed, but could be brought back fairly simply now that the pipeline threading work is finished.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
@@ -177,15 +177,9 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility
         /// Gets all aliases found in the runspace
         /// </summary>
         /// <param name="executionService"></param>
-        public static async Task<(ConcurrentDictionary<string, List<string>>, ConcurrentDictionary<string, string>)> GetAliasesAsync(IInternalPowerShellExecutionService executionService)
+        public static async Task<(Dictionary<string, List<string>>, Dictionary<string, string>)> GetAliasesAsync(IInternalPowerShellExecutionService executionService)
         {
             Validate.IsNotNull(nameof(executionService), executionService);
-
-            // TODO: Should we return the caches if they're not empty, or always update?
-            // if (!s_cmdletToAliasCache.IsEmpty || !s_aliasToCmdletCache.IsEmpty)
-            // {
-            //     return (s_cmdletToAliasCache, s_aliasToCmdletCache);
-            // }
 
             IEnumerable<CommandInfo> aliases = await executionService.ExecuteDelegateAsync<IEnumerable<CommandInfo>>(
                 nameof(GetAliasesAsync),
@@ -209,7 +203,8 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility
                 s_aliasToCmdletCache.TryAdd(aliasInfo.Name, aliasInfo.Definition);
             }
 
-            return (s_cmdletToAliasCache, s_aliasToCmdletCache);
+            return (new Dictionary<string, List<string>>(s_cmdletToAliasCache),
+                new Dictionary<string, string>(s_aliasToCmdletCache));
         }
     }
 }
--- a/src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System.Collections.Concurrent;
@@ -50,8 +50,9 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility
             };
 
         private static readonly ConcurrentDictionary<string, CommandInfo> s_commandInfoCache = new();
-
         private static readonly ConcurrentDictionary<string, string> s_synopsisCache = new();
+        private static readonly ConcurrentDictionary<string, List<string>> s_cmdletToAliasCache = new(System.StringComparer.OrdinalIgnoreCase);
+        private static readonly ConcurrentDictionary<string, string> s_aliasToCmdletCache = new(System.StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Gets the CommandInfo instance for a command with a particular name.
@@ -170,6 +171,45 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility
             }
 
             return synopsisString;
+        }
+
+        /// <summary>
+        /// Gets all aliases found in the runspace
+        /// </summary>
+        /// <param name="executionService"></param>
+        public static async Task<(ConcurrentDictionary<string, List<string>>, ConcurrentDictionary<string, string>)> GetAliasesAsync(IInternalPowerShellExecutionService executionService)
+        {
+            Validate.IsNotNull(nameof(executionService), executionService);
+
+            // TODO: Should we return the caches if they're not empty, or always update?
+            // if (!s_cmdletToAliasCache.IsEmpty || !s_aliasToCmdletCache.IsEmpty)
+            // {
+            //     return (s_cmdletToAliasCache, s_aliasToCmdletCache);
+            // }
+
+            IEnumerable<CommandInfo> aliases = await executionService.ExecuteDelegateAsync<IEnumerable<CommandInfo>>(
+                nameof(GetAliasesAsync),
+                Execution.ExecutionOptions.Default,
+                (pwsh, _) =>
+                {
+                    CommandInvocationIntrinsics invokeCommand = pwsh.Runspace.SessionStateProxy.InvokeCommand;
+                    return invokeCommand.GetCommands("*", CommandTypes.Alias, nameIsPattern: true);
+                },
+                CancellationToken.None).ConfigureAwait(false);
+
+            foreach (AliasInfo aliasInfo in aliases)
+            {
+                // TODO: When we move to netstandard2.1, we can use another overload which generates
+                // static delegates and thus reduces allocations.
+                s_cmdletToAliasCache.AddOrUpdate(
+                    aliasInfo.Definition,
+                    (_) => new List<string> { aliasInfo.Name },
+                    (_, v) => { v.Add(aliasInfo.Name); return v; });
+
+                s_aliasToCmdletCache.TryAdd(aliasInfo.Name, aliasInfo.Definition);
+            }
+
+            return (s_cmdletToAliasCache, s_aliasToCmdletCache);
         }
     }
 }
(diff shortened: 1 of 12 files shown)


src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
```diff
@@ -168,5 +172,44 @@ internal static class CommandHelpers
...
+                CancellationToken.None).ConfigureAwait(false);
+
+            foreach (AliasInfo aliasInfo in aliases)
+            {
+                List<string> aliasList = new() { aliasInfo.Name };
+
+                s_cmdletToAliasCache.AddOrUpdate(
+                    aliasInfo.Definition,
+                    aliasList,
+                    (_, v) => v.Concat(aliasList).ToList());
```
daxian-dbw, line 207: Maybe it's better to use the other overload for `AddOrUpdate` to avoid creating extra `List` instances when multiple alias mapping to the same commands.
```
s_cmdletToAliasCache.AddOrUpdate(
    aliasInfo.Definition,
    (_, arg) => new List<string>(arg),
    (_, v, arg) => { v.Add(arg); return v },
    aliasInfo.Name);
```
Those 2 lambdas will be static delegates.

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
```diff
@@ -153,42 +154,21 @@ public static SymbolReference FindCommandAtPosition(Ast scriptAst, int lineNumbe
...
+        /// <param name="cmdletToAliasDictionary">Dictionary maping cmdlets to aliases for finding alias references</param>
+        /// <param name="aliasToCmdletDictionary">Dictionary maping aliases to cmdlets for finding alias references</param>
         /// <returns></returns>
         public static IEnumerable<SymbolReference> FindReferencesOfSymbol(
             Ast scriptAst,
             SymbolReference symbolReference,
-            Dictionary<String, List<String>> CmdletToAliasDictionary,
-            Dictionary<String, String> AliasToCmdletDictionary)
+            ConcurrentDictionary<string, List<string>> cmdletToAliasDictionary = default,
+            ConcurrentDictionary<string, string> aliasToCmdletDictionary = default)
```
daxian-dbw, line 164: I'm curious about the change from `Dictionary` to `ConccurrentDictionary`. Can you please briefly describe the scenarios where concurrent updates to the dictionary is needed?

src/PowerShellEditorServices/Services/Symbols/Vistors/FindReferencesVisitor.cs
```diff
@@ -12,48 +13,47 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
...
+                return;
+            }
+
+            _needsAliases = true;
+            _cmdletToAliasDictionary = cmdletToAliasDictionary;
+            _aliasToCmdletDictionary = aliasToCmdletDictionary;
 
             // Try to get the symbolReference's command name of an alias,
             // if a command name does not exists (if the symbol isn't an alias to a command)
             // set symbolRefCommandName to and empty string value
```
daxian-dbw, line 50:
```suggestion
            // set symbolRefCommandName to an empty string value
```

src/PowerShellEditorServices/Services/Symbols/Vistors/FindReferencesVisitor.cs
```diff
@@ -68,50 +68,44 @@ public override AstVisitAction VisitCommand(CommandAst commandAst)
...
                     {
                         // Check if the found symbol's name is the same as the commandAst's name OR
                         // if the symbol's name is an alias for this commandAst's name (commandAst is a cmdlet) OR
                         // if the symbol's name is the same as the commandAst's cmdlet name (commandAst is a alias)
-                        if (commandName.Equals(symbolRef.SymbolName, StringComparison.CurrentCultureIgnoreCase) ||
-                        alaises.Contains(symbolRef.ScriptRegion.Text.ToLower()) ||
-                        command.Equals(symbolRef.ScriptRegion.Text, StringComparison.CurrentCultureIgnoreCase) ||
-                        (!string.IsNullOrEmpty(command) && command.Equals(symbolRefCommandName, StringComparison.CurrentCultureIgnoreCase)))
+                        if (commandName.Equals(_symbolRef.SymbolName, StringComparison.CurrentCultureIgnoreCase)
+                            || aliases.Contains(_symbolRef.ScriptRegion.Text.ToLower())
```
daxian-dbw, line 91: Are the elements in `aliases` always lower-case strings? If not, this may return `false` when you're expecting `true`. e.g. `new List<string>() { "Abc" }.Contains("abc")`.

src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
andschwa: Oh I like that! Thank you.

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
andschwa: Really just that there is no guarantee that `GetAliasesAsync` will be called with a lock around it, so concurrent requests to resolve references may concurrently ask for aliases to be resolved, and so concurrently access the original `ConcurrentDictionary` caches in `CommandHelpers.cs`. Now, `AstOperations.FindReferences` is called with a newly constructed instance each time, so _it_ should be thread safe since it's passed copies of the cache dictionaries, but it was easiest to just make the types match. I suppose I should instead receive an `IDictionary<K,V>` here.

andschwa: Fixed!

src/PowerShellEditorServices/Services/Symbols/Vistors/FindReferencesVisitor.cs
andschwa: Good catch, it also needs the ignore case comparer.

src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
```diff
@@ -168,5 +172,37 @@ internal static class CommandHelpers
...
+                    return invokeCommand.GetCommands("*", CommandTypes.Alias, nameIsPattern: true);
+                },
+                CancellationToken.None).ConfigureAwait(false);
+
+            foreach (AliasInfo aliasInfo in aliases)
+            {
+                s_cmdletToAliasCache.AddOrUpdate(
+                    aliasInfo.Definition,
+                    (_) => new List<string> { aliasInfo.Name},
+                    (_, v) => { v.Add(aliasInfo.Name); return v; });
```
daxian-dbw, line 199: This would beat the purpose, as these 2 lambdas are closures and thus a new instance will be created for every `aliasInfo` object, so it will end up with more allocations.

The overload I suggested to use will generate static delegates can are reused for all the calls to `AddOrUpdate` here, and thus could reduce allocations.

daxian-dbw (approved):
LGTM except the one last comment I left above.

src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
andschwa: The suggested code didn't compile...what's the fix for it?

andschwa: I see [this](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-6.0#system-collections-concurrent-concurrentdictionary-2-addorupdate-1(-0-system-func((-0-0-1))-system-func((-0-1-0-1))-0)) let me see if I can get it right.

daxian-dbw: My suggestion is pseudo code :) I didn't try compiling it. I guess the following would work
```
s_cmdletToAliasCache.AddOrUpdate(
    aliasInfo.Definition,
    (_, arg) => new List<string> { arg },
    (_, v, arg) => { v.Add(arg); return v; },
    aliasInfo.Name);
```

andschwa: I'm getting this:

> No overload for method 'AddOrUpdate' takes 4 arguments

Which is why I made the change I did instead. But I agree the documentation says this overload exists and takes four arguments, and we're using .NET 6 here.

andschwa: If I write it like this:

```c#
                s_cmdletToAliasCache.AddOrUpdate<string>(
                    aliasInfo.Definition,
                    (_, arg) => new List<string> { arg },
                    (_, v, arg) => { v.Add(arg); return v; },
                    aliasInfo.Name);
```

Which I think is right, I get:

> The non-generic method 'ConcurrentDictionary<string, List<string>>.AddOrUpdate(string, List<string>, Func<string, List<string>, List<string>>)' cannot be used with type arguments

Do I have a typo somewhere or a misunderstanding?

daxian-dbw: That's weird. The following works fine in a `net6.0` project:
```c#
    class Program
    {
        static void Main(string[] args)
        {
            ConcurrentDictionary<string, List<string>> cmdletToAliasCache = new(StringComparer.CurrentCultureIgnoreCase);
            string key = "key";
            string value = "value";

            cmdletToAliasCache.AddOrUpdate(
                key,
                (_, arg) => new List<string> { arg },
                (_, v, arg) => { v.Add(arg); return v; },
                value);

            cmdletToAliasCache.TryGetValue(key, out var myValue);
            Console.WriteLine(myValue[0]);
        }
    }
```
image

daxian-dbw: This overload only exists in .NET Framework.

andschwa: I wonder if it's because our target framework is `netstandard2.0`.

andschwa: Screen Shot 2022-01-07 at 5 13 13 PM

daxian-dbw: I see, so PSES is targeting netstandard2.0?
In that case, please keep your original code unchanged.

daxian-dbw: > we're using .NET 6 here

Just so I'm clear :) we are using .NET 6 SDK but the build targets netstandard 2.0, is that right?

andschwa: @daxian-dbw Since we're using `netstandard2.0` and `net461` in order to continue to support Windows PowerShell 5.1, I can't use this overload 😢

daxian-dbw: That's OK. It's an optimization change anyway 😄

andschwa: I've added a TODO to improve this with the better delegate if/when we move to .NET 4.7.2 and/or .NET Standards 2.1 (so really, the day we drop PowerShell 5.1 support).

andschwa:
Thanks for your time reviewing @daxian-dbw! This resolves the last disabled test from the original set of Language Server tests.

PR #1657, opened 2022-01-07
Re-enable `ExtensionCommandTests.cs`

### LINKED ISSUE #1444: Fix tests in `Extensions/ExtensionCommandTests.cs`

andschwa:
These tests actually did not require the language server nor the service provider, we just had to manually initialize the extension service itself. Resolves #1444.

### DIFF
--- a/test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
+++ b/test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
@@ -1,41 +1,48 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.Logging.Abstractions;
-using Microsoft.PowerShell.EditorServices.Extensions;
-using Microsoft.PowerShell.EditorServices.Extensions.Services;
-using Microsoft.PowerShell.EditorServices.Services;
-using Microsoft.PowerShell.EditorServices.Services.TextDocument;
-using Microsoft.PowerShell.EditorServices.Test.Shared;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Management.Automation;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.PowerShell.EditorServices.Extensions;
+using Microsoft.PowerShell.EditorServices.Extensions.Services;
+using Microsoft.PowerShell.EditorServices.Services.Extension;
+using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
+using Microsoft.PowerShell.EditorServices.Services.TextDocument;
+using Microsoft.PowerShell.EditorServices.Test.Shared;
 using Xunit;
 
 namespace Microsoft.PowerShell.EditorServices.Test.Extensions
 {
-    // TODO:
-    // These tests require being able to instantiate a language server and use the service provider.
-    // Re-enable them when we have mocked out more infrastructure for testing.
-
-    /*
+    [Trait("Category", "Extensions")]
     public class ExtensionCommandTests : IDisposable
     {
-        private readonly PowerShellContextService _powershellContextService;
+        private readonly PsesInternalHost psesHost;
 
-        private readonly IExtensionCommandService _extensionCommandService;
-
-        private readonly ExtensionService _extensionService;
+        private readonly ExtensionCommandService extensionCommandService;
 
         public ExtensionCommandTests()
         {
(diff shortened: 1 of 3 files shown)


andschwa:
Yay, an easy one!

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
```diff
@@ -48,34 +55,30 @@ public async Task CanRegisterAndInvokeCommandWithCmdletName()
...
 
-            await _powershellContextService.ExecuteScriptStringAsync(
-                TestUtilities.NormalizeNewlines($@"
-function Invoke-Extension {{ $global:testValue = 5 }}
-Register-EditorCommand -Name {commandName} -DisplayName ""{commandDisplayName}"" -Function Invoke-Extension"));
+            await psesHost.ExecutePSCommandAsync(
+                new PSCommand().AddScript(
+                    "function Invoke-Extension { $global:extensionValue = 5 }; " +
+                    $"Register-EditorCommand -Name {commandName} -DisplayName \"{commandDisplayName}\" -Function Invoke-Extension"),
+                CancellationToken.None).ConfigureAwait(true);
```
daxian-dbw, line 67: `ConfigureAwait(true)` is the default behavior, so why adding this call explicitly?
It looks you are emphasizing that current synchronization context needs to be captured, so where is the synchronization context come from?

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
```diff
@@ -154,41 +151,33 @@ public async Task CanUnregisterCommand()
...
 
             // Ensure that the command has been unregistered
-            await Assert.ThrowsAsync<KeyNotFoundException>(() =>
-                _extensionCommandService.InvokeCommandAsync("test.scriptblock", editorContext));
-        }
-
-        public void Dispose()
-        {
-            _powershellContextService.Dispose();
+            _ = await Assert.ThrowsAsync<KeyNotFoundException>(
```
daxian-dbw, line 179: Maybe `_ =` is not needed?
```suggestion
            await Assert.ThrowsAsync<KeyNotFoundException>(
```

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
andschwa: The code analyzers are setup to require `ConfigureAwait` to be set explicitly, because not setting `ConfigureAwait(false)` in the rest of the codebase (where it acts like a library) leads to a recurring bug when users load stuff like `Windows.Forms` into the extension's console.

The xUnit synchronization context is what's being captured, and in the tests it needs to be captured so the test code always returns to a thread that isn't the pipeline.

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
andschwa: It is needed, there is a Roslyn code analyzer warning otherwise.

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
andschwa: > Consider calling ConfigureAwait on the awaited task [PowerShellEditorServices.Test] csharp(CA2007)

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
andschwa: Nvm, that's just the first suggested fix, after adding `ConfigureAwait(true)` the actual error goes away. Strange.

test/PowerShellEditorServices.Test/Extensions/ExtensionCommandTests.cs
andschwa: https://github.com/PowerShell/PowerShellEditorServices/blob/70f31d9e5a0268a8e1512c735da675f342a08d94/.editorconfig#L31-L32

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007 enabled in ec4bc2f44

SeeminglyScience (approved):
LGTM!

PR #1662, opened 2022-01-12
Enable resolution of an alias to its function definition

andschwa:
Two caveats:

1. The alias must be set in the runspace, which means the script needs
   to have been executed.
2. The definition is the function itself, not the `Set-Alias` command.

Resolves https://github.com/PowerShell/vscode-powershell/issues/2800


andschwa:
FYI @PrzemyslawKlys

PrzemyslawKlys:
@andschwa Thanks :-) I am closely monitoring your progress ;-)

daxian-dbw (approved):
LGTM

PR #1664, opened 2022-01-13
Fix `DebuggerSetsVariablesWithConversion` test

### LINKED ISSUE #1661: Fix `DebuggerSetsVariablesWithConversion` test which indicates bug
I confirmed this is a real bug when reproducing manually. Setting variables with the preview's debugger can fail if they need to be converted. Left a note actually where it throws the exception.

https://github.com/PowerShell/PowerShellEditorServices/blob/20a35163c39cceaee2023b49783bef6a0c4dfaec/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs#L616-L672

andschwa:
With these two bugs fixed, we can now set variables which are strongly typed (and so need a conversion) through the debugger. This was a great unit test that indicated a bug, and it was easily reproducible.

While testing manually, the scope bug revealed itself when I attempted to set a variable in the "automatic variables" container (the test is using the local variables container). I think assuming a default scope of "0" is sane because if it's an automatic variable, it must be in the local scope, and every other container is mapped to a named scope. As we learned previously, we cannot "guess" at a scope because they do not map one-to-one with stack frames.

The conversion bug was due to `SessionStateProxy.GetVariable("ExecutionContext")` throwing [`NoSessionStateProxyWhenPipelineInProgress`](https://github.com/PowerShell/PowerShell/blob/7dc4587014bfa22919c933607bf564f0ba53db2e/src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs#L1187). Initially I thought it could be solved by running it forcibly in the foreground (as the stack-trace showed it was being run via `OnIdle`); however, the same exception was thrown even then. I am not certain, but I think it is because the debugger is currently in a stopped state, which has left the pipeline running, and so this concurrency check fails. I don't know why it can't work concurrently, because this alternative means to get the `EngineIntrinsics` value works, and indeed, was the way we did it previously:

https://github.com/PowerShell/PowerShellEditorServices/blob/99b5b7e1a50f334bb0654e801484802c345a9cdf/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs#L499-L521 

Fixes #1661

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -400,8 +400,14 @@ namespace Microsoft.PowerShell.EditorServices.Services
             }
 
             VariableDetailsBase variable = variableContainer.Children[name];
-            // Determine scope in which the variable lives so we can pass it to `Get-Variable -Scope`.
-            string scope = null; // TODO: Can this use a fancy pattern matcher?
+
+            // Determine scope in which the variable lives so we can pass it to `Get-Variable
+            // -Scope`. The default is scope 0 which is safe because if a user is able to see a
+            // variable in the debugger and so change it through this interface, it's either in the
+            // top-most scope or in one of the following named scopes. The default scope is most
+            // likely in the case of changing from the "auto variables" container.
+            string scope = "0";
+            // NOTE: This can't use a switch because the IDs aren't constant.
             if (variableContainerReferenceId == localScopeVariables.Id)
             {
                 scope = VariableContainerDetails.LocalScopeName;
@@ -413,11 +419,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
             else if (variableContainerReferenceId == globalScopeVariables.Id)
             {
                 scope = VariableContainerDetails.GlobalScopeName;
-            }
-            else
-            {
-                // Hmm, this would be unexpected. No scope means do not pass GO, do not collect $200.
-                throw new Exception("Could not find the scope for this variable.");
             }
 
             // Now that we have the scope, get the associated PSVariable object for the variable to be set.
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -457,22 +457,25 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
             if (argTypeConverterAttr is not null)
             {
+                // PSVariable *is* strongly typed, so we have to convert it.
                 _logger.LogTrace($"Setting variable '{name}' using conversion to value: {expressionResult ?? "<null>"}");
 
-                // TODO: This is throwing a 'PSInvalidOperationException' thus causing
-                // 'DebuggerSetsVariablesWithConversion' to fail.
-                psVariable.Value = await _executionService.ExecuteDelegateAsync(
-                    "PS debugger argument converter",
-                    ExecutionOptions.Default,
-                    (pwsh, _) =>
-                    {
-                        var engineIntrinsics = (EngineIntrinsics)pwsh.Runspace.SessionStateProxy.GetVariable("ExecutionContext");
-
-                        // TODO: This is almost (but not quite) the same as LanguagePrimitives.Convert(), which does not require the pipeline thread.
-                        //       We should investigate changing it.
-                        return argTypeConverterAttr.Transform(engineIntrinsics, expressionResult);
-                    },
+                // NOTE: We use 'Get-Variable' here instead of 'SessionStateProxy.GetVariable()'
+                // because we already have a pipeline running (the debugger) and the latter cannot
+                // run concurrently (threw 'NoSessionStateProxyWhenPipelineInProgress').
+                IReadOnlyList<EngineIntrinsics> results = await _executionService.ExecutePSCommandAsync<EngineIntrinsics>(
+                    new PSCommand()
+                        .AddCommand(@"Microsoft.PowerShell.Utility\Get-Variable")
+                        .AddParameter("Name", "ExecutionContext")
+                        .AddParameter("ValueOnly"),
                     CancellationToken.None).ConfigureAwait(false);
+                EngineIntrinsics engineIntrinsics = results.Count > 0
+                    ? results[0]
+                    : throw new Exception("Couldn't get EngineIntrinsics!");
+
+                // TODO: This is almost (but not quite) the same as 'LanguagePrimitives.Convert()',
+                // which does not require the pipeline thread. We should investigate changing it.
+                psVariable.Value = argTypeConverterAttr.Transform(engineIntrinsics, expressionResult);
             }
             else
             {
@@ -642,7 +645,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
         private async Task<VariableContainerDetails> FetchVariableContainerAsync(string scope, bool autoVarsOnly)
         {
-            PSCommand psCommand = new PSCommand().AddCommand("Get-Variable").AddParameter("Scope", scope);
+            PSCommand psCommand = new PSCommand().AddCommand(@"Microsoft.PowerShell.Utility\Get-Variable").AddParameter("Scope", scope);
 
             var scopeVariableContainer = new VariableContainerDetails(nextVariableId++, "Scope: " + scope);
             variables.Add(scopeVariableContainer);
--- a/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
+++ b/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
@@ -613,7 +613,7 @@ namespace Microsoft.PowerShell.EditorServices.Test.Debugging
             Assert.Equal(newGlobalIntValue, intGlobalVar.ValueString);
         }
 
-        [Fact(Skip = "Variable conversion is broken")]
+        [Fact]
         public async Task DebuggerSetsVariablesWithConversion()
         {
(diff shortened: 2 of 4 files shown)


src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -400,8 +400,14 @@ public async Task<string> SetVariableAsync(int variableContainerReferenceId, str
...
             VariableDetailsBase variable = variableContainer.Children[name];
-            // Determine scope in which the variable lives so we can pass it to `Get-Variable -Scope`.
-            string scope = null; // TODO: Can this use a fancy pattern matcher?
+
+            // Determine scope in which the variable lives so we can pass it to `Get-Variable
+            // -Scope`. The default is scope 0 which is safe because if a user is able to see a
+            // variable in the debugger and so change it through this interface, it's either in the
+            // top-most scope or in one of the following named scopes. The default scope is most
+            // likely in the case of changing from the "auto variables" container.
+            string scope = "0";
```
andschwa, line 409: @SeeminglyScience This seem sane to you?

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -456,22 +457,25 @@ public async Task<string> SetVariableAsync(int variableContainerReferenceId, str
...
-                        // TODO: This is almost (but not quite) the same as LanguagePrimitives.Convert(), which does not require the pipeline thread.
-                        //       We should investigate changing it.
-                        return argTypeConverterAttr.Transform(engineIntrinsics, expressionResult);
-                    },
+                // NOTE: We use 'Get-Variable' here instead of 'SessionStateProxy.GetVariable()'
+                // because we already have a pipeline running (the debugger) and the latter cannot
+                // run concurrently (threw 'NoSessionStateProxyWhenPipelineInProgress').
+                IReadOnlyList<EngineIntrinsics> results = await _executionService.ExecutePSCommandAsync<EngineIntrinsics>(
+                    new PSCommand()
+                        .AddCommand(@"Microsoft.PowerShell.Utility\Get-Variable")
```
andschwa, line 468: @rjmholt Do you have a better idea or explanation as to why that exception was being thrown?

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -456,22 +457,25 @@ public async Task<string> SetVariableAsync(int variableContainerReferenceId, str
...
+                        .AddParameter("Name", "ExecutionContext")
+                        .AddParameter("ValueOnly"),
                     CancellationToken.None).ConfigureAwait(false);
+                EngineIntrinsics engineIntrinsics = results.Count > 0
+                    ? results[0]
+                    : throw new Exception("Couldn't get EngineIntrinsics!");
+
+                // TODO: This is almost (but not quite) the same as 'LanguagePrimitives.Convert()',
+                // which does not require the pipeline thread. We should investigate changing it.
```
andschwa, line 477: I looked into this, but I don't think it will quite as "well" as the transformation, and I'm unsure actually how to go about it because [`LanguagePrimitives.ConvertTo()`](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.languageprimitives.convertto?view=powershellsdk-7.0.0#System_Management_Automation_LanguagePrimitives_ConvertTo_System_Object_System_Type_System_IFormatProvider_) doesn't accept an `ArgumentTypeConverterAttribute`.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -641,7 +645,7 @@ private Task<VariableContainerDetails> FetchVariableContainerAsync(string scope)
 
         private async Task<VariableContainerDetails> FetchVariableContainerAsync(string scope, bool autoVarsOnly)
         {
-            PSCommand psCommand = new PSCommand().AddCommand("Get-Variable").AddParameter("Scope", scope);
+            PSCommand psCommand = new PSCommand().AddCommand(@"Microsoft.PowerShell.Utility\Get-Variable").AddParameter("Scope", scope);
```
andschwa, line 648: I'm not sure _why_ we consistently use the fully qualified name, but since we do, fixed this one too.

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -628,7 +628,7 @@ public async Task DebuggerSetsVariablesWithConversion()
             VariableDetailsBase[] variables = GetVariables(VariableContainerDetails.LocalScopeName);
 
             // Test set of a local string variable (not strongly typed but force conversion)
-            const string newStrValue = "False";
+            const string newStrValue = "\"False\"";
```
andschwa, line 631: This was erroneously changed by me previously, and is now back to [its original value](https://github.com/PowerShell/PowerShellEditorServices/blob/99b5b7e1a50f334bb0654e801484802c345a9cdf/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs#L760).

test/PowerShellEditorServices.Test/PsesHostFactory.cs
```diff
@@ -53,24 +45,23 @@ public static PsesInternalHost Create(ILoggerFactory loggerFactory)
...
+                logLevel: (int)LogLevel.None,
                 consoleReplEnabled: false,
                 usesLegacyReadLine: false,
                 bundledModulePath: BundledModulePath);
 
             var psesHost = new PsesInternalHost(loggerFactory, null, testHostDetails);
 
             // NOTE: Because this is used by constructors it can't use await.
-            // TODO: Should we actually load profiles here?
             if (psesHost.TryStartAsync(new HostStartOptions { LoadProfiles = true }, CancellationToken.None).GetAwaiter().GetResult())
```
andschwa, line 65: Hm, now I'm actually wondering if this should be false. I thought I had my answer, and that's that we [test we can load the profiles](https://github.com/PowerShell/PowerShellEditorServices/blob/f3d4bf2c6c2701d1cf60290b4abf72be09a4192a/test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs#L110-L144) but we actually do the profile loading call explicitly. @rjmholt Do you know why this is explicitly set to true?

test/PowerShellEditorServices.Test/PsesHostFactory.cs
```diff
@@ -53,25 +45,24 @@ public static PsesInternalHost Create(ILoggerFactory loggerFactory)
...
                 consoleReplEnabled: false,
                 usesLegacyReadLine: false,
                 bundledModulePath: BundledModulePath);
 
             var psesHost = new PsesInternalHost(loggerFactory, null, testHostDetails);
 
             // NOTE: Because this is used by constructors it can't use await.
-            // TODO: Should we actually load profiles here?
-            if (psesHost.TryStartAsync(new HostStartOptions { LoadProfiles = true }, CancellationToken.None).GetAwaiter().GetResult())
+            if (psesHost.TryStartAsync(new HostStartOptions { LoadProfiles = false }, CancellationToken.None).GetAwaiter().GetResult())
```
andschwa, line 65: I've now set this to false because the tests which assert profile loading works do so by explicitly running the same code as would otherwise run in startup:

https://github.com/PowerShell/PowerShellEditorServices/blob/20a35163c39cceaee2023b49783bef6a0c4dfaec/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L217-L221

I don't think we _really_ need to be calling this delegate for every other test.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
PaulHigin: This looks correct to me.  The debugger knows it is running in an active pipeline and runs commands as 'nested' on the target (locally or remotely), whereas SessionStateProxy does not.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
PaulHigin: This will only work for local debugging sessions.  Does this code make local/remote distinctions?

PaulHigin (approved):
Added a couple of comments/questions.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
andschwa: It does not make that distinction. Also, do you mean that `LanguagePrimitives` would only work for local, or the currently proposed change?

PaulHigin: The ExecutionContext variable (engineIntrinsics) passes through the serialization system for the remoting case, and is just a PSObject property bag and not a System.Management.Automation.EngineIntrinsics type.  The serialization system does not rehydrate returned object into their original types, except for a small set of special cases.

So if you have script code that determines variable scope, that script should run entirely on the remote target, so that it can run in the correct context and 'live' type instances.

andschwa: I'm not quite sure I'm following. Can I ask: will the proposed change behave _differently_ than the existing code in the remoting scenario? If not, then let's kick that down the road a bit.

andschwa: (I.e. are both implementations broken in the same way?)

PaulHigin: Yes, I think so (remoting may be broken).  But I need to look more closely at the code (and not just the changes).

PR #1665, opened 2022-01-13
Add missing overrides in `EditorServicesConsolePSHostRawUserInterface`

andschwa:
This is an attempt to fix https://github.com/PowerShell/vscode-powershell/issues/3651


andschwa:
@SeeminglyScience My quick testing where I used `_consoleHostUI` instead of `_underlyingHostUI` also did not work. I debugged it and the test code in the given issue definitely just calls `public override void WriteLine(string value)` (not `WriteInformation`) and it's doing with just the raw string values `Test 1` etc. So something is dropping the `-NoNewLine` and the colors _before_ it gets here.

andschwa:
Closing as we cherry-picked this commit into #1668

PR #1676, opened 2022-01-24
Use EditorConfig for dotnet build and suppress existing issues

JustinGrote:
This PR will configure the various .NET projects to use the editorconfig suppression settings, and adds several to demote existing (but working) warnings to suggestions, including rosylnator.

This is just meant as a stop-gap to ensure that dotnet build in CI looks "clean", so that finding/inspecting errors takes less time. Next step is to go back and selectively remove the warnings and either fix the code or drop a selective exception in for that particular warning. "Net-New" warning categories not already present in the current code have not been demoted and will continue to be reported as such.

With this, the VSCode is no longer a sea of yellow warnings, so when working on a PR it's much easier to see where specifically *your* code has issues. All issues are still reported as suggestions however (currently 1273 total...) but that can be easily toggled as a filter in VSCode Problems pane.
image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -32,16 +32,56 @@ dotnet_diagnostic.CA1823.severity = error
 dotnet_diagnostic.CA2007.severity = error
 # CA2016: Forward the CancellationToken parameter to methods that take one
 dotnet_diagnostic.CA2016.severity = error
-# All maintainability issues (dead code etc.)
+
+# TODO: Enable all maintainability issues (dead code etc.) and enforce
 # See: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/maintainability-warnings
 dotnet_analyzer_diagnostic.category-Maintainability.severity = error
-# VSTHRD002: Synchronously waiting on tasks or awaiters may cause deadlocks
-# TODO: Fix all of these issues and explicitly ignore the intentional ones.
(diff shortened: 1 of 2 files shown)


JustinGrote:
> Woo! Thanks for this @JustinGrote. I added documentation to all the suggestions added, and used the `common.props` file to set the enforcement (instead of each file individually). I double checked that all the files I reverted did indeed include `common.props` and resolved the TODO in that file.

Sounds good! Forgot about common.props

andschwa (approved):
🥳

.editorconfig
```diff
@@ -32,16 +32,56 @@ dotnet_diagnostic.CA1823.severity = error
...
+dotnet_diagnostic.VSTHRD114.severity = suggestion
 # VSTHRD200: Use "Async" suffix for awaitable methods
-dotnet_diagnostic.VSTHRD200.severity = silent
+dotnet_diagnostic.VSTHRD200.severity = suggestion
+
+# xUnit2013: Do not use equality check to check for collection size
+dotnet_diagnostic.xUnit2013.severity = suggestion
 # IDE0003: this and Me preferences
-dotnet_diagnostic.IDE0003.severity = silent
+dotnet_diagnostic.IDE0003.severity = suggestion
```
qmqmpzpz, line 84: "when": "connectionProvider == MSSQL && nodeType == Folder && nodeLabel == Tables && config.workbench.enablePreviewFeatures",

Issue #1677, opened 2022-01-24
Recover PRs lost in pipeline rewrite rebase 


SydneyhSmith:
When rebasing the pipeline rewrite work we werent pulling in the changes we made, just overwriting them with rewritten code that wasn't reconciled with updates...as a result of this a number of PRs need to be rewritten/re-added to the code base. these are the PRs we know of that need to be recovererd: 

- [x] https://github.com/PowerShell/PowerShellEditorServices/pull/1514
- [x] https://github.com/PowerShell/PowerShellEditorServices/pull/1522
- [x] https://github.com/PowerShell/PowerShellEditorServices/pull/1570
- [x] https://github.com/PowerShell/PowerShellEditorServices/pull/1571

andschwa:
Thanks for capturing these @SydneyhSmith. Good news: we've managed to already get 3/4 of these.

PR #1685, opened 2022-01-26
Send `stopDebugger` during REPL if no longer debugging

andschwa:
This finishes #1677 by resurrecting the bug-fix originally implemented in #1570. This allows the user to start a debug session within the integrated console and then have the VS Code debugger stop when the PowerShell debug session stops.


PaulHigin (approved):
This looks right to me.  `Debugger.InBreakpoint` property is intended to inform when the runspace associated script debugger has stopped script execution (in breakpoint) and is awaiting debug commands to process, for both local and remote cases.  When normal script execution resumes, `InBreakpoint` becomes false.

andschwa:
The newest commit should also fix https://github.com/PowerShell/vscode-powershell/issues/3700.

Issue #1686, opened 2022-01-27
Variable Expansion: ScriptProperty ToStrings are slow to perform


JustinGrote:
![image](https://user-images.githubusercontent.com/15258962/151281274-b443b5f7-c8f3-47e4-abf4-298a17384a98.png)

Per @SeeminglyScience it's probably because the ToString is trying to marshal to the pipeline thread and deadlocking.

JustinGrote:
Looks like it doesn't happen unless $this is involved (this probably gets translated into something that doesn't require the pipeline to execute)
![image](https://user-images.githubusercontent.com/15258962/151477226-ba3fdf3d-d9af-47ac-8bb5-45e012999d76.png)

JustinGrote:
Narrowed it down to this loop for enumerables that is "slow", I'll dig deeper into `VariableDetails` later.
https://github.com/PowerShell/PowerShellEditorServices/blob/624fe30eda7b7dc918978e238a59b26b81a40dc7/src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs#L338-L344

SeeminglyScience:
> @SeeminglyScience looks like it's not hard-hanging, just a performance issue. Could this have to do with the marshaling still?

Yep!

In certain circumstances (like this one) it will timeout after 250ms and then just invoke it anyway from a different thread (while still targeting the same runspace) potentially causing state corruption. So I'd say that still perfectly explains it and potentially makes it a little bit more scary.

JustinGrote:
Here's the culprit: 
https://github.com/PowerShell/PowerShellEditorServices/blob/624fe30eda7b7dc918978e238a59b26b81a40dc7/src/PowerShellEditorServices/Services/DebugAdapter/Debugging/VariableDetails.cs#L198

If you set it to null instead of `SafeToString()`, everything is fast again.

JustinGrote:
More specifically:
https://github.com/PowerShell/PowerShellEditorServices/blob/3274c230cc6675859d5c89d6c176219c2c9430f7/src/PowerShellEditorServices/Utility/Extensions.cs#L25

@SeeminglyScience how would I get this obj.ToString() method on the correct pipe? It does feel like about a 250ms delay on each invocation.

JustinGrote:
Adding @SeeminglyScience as I've hit my limits on figuring this out for now :)

andschwa:
And fixed, with your test too!

PR #1688, opened 2022-01-30
WIP: Marshal Script-based ToString to PS Execution Pipeline for Performance

JustinGrote:
#1686

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Management.Automation;
 using System.Management.Automation.Language;
+using System.Management.Automation.Runspaces;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -263,7 +264,30 @@ public VariableDetailsBase[] GetVariables(int variableReferenceId)
...
+                    // to run on the pipeline thread to avoid retry timeouts
+                    // TODO: Cancellation Token Maybe?
+                    // TODO: Maybe pass
+                    childVariables = _psesHost.InvokeDelegate<VariableDetailsBase[]>
+                    (
+                        representation: "GetChildren",
+                        ExecutionOptions.Default,
+                        (_) =>
+                        {
+                            var existingRunspace = Runspace.DefaultRunspace;
```
andschwa, line 277: I'm just curious why you need to save, change, and then reset the runspace?

JustinGrote: @SeeminglyScience can comment but I think it's just a safety/idempotency thing, in all likelihood it's the same runspace so this basically doesn't do anything.

andschwa: Hm, if that's something we need to do perhaps we should do it in a wrapper? Like, where/when else might we have to do something like that?

And if we're not sure that we _have_ to do it, I'd rather leave a TODO for now so if we encounter a bug we fix it later.

SeeminglyScience: Yeah that was a troubleshooting step that didn't do anything right? If so that can be removed for sure.

andschwa:
@JustinGrote was this something you're still working on?

JustinGrote:
@andschwa I think I may abandon this PR in favor of the lazy variable inititialization if it is a scriptproperty like Typescript does it, seems like the better way to go.

andschwa:
Hey, sounds good, I think that's the way to go too.

andschwa:
Going to close this out as I did practically the same thing in https://github.com/PowerShell/PowerShellEditorServices/pull/1975.

PR #1690, opened 2022-02-01
Display prompt after `F8` finishes

andschwa:
This fixes https://github.com/PowerShell/vscode-powershell/issues/3685 by awaiting the task spawned by the evaluate handler (triggered by F8) and skipping the "next" prompt that actually should have been interrupted by that foreground task.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebugEvaluateHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebugEvaluateHandler.cs
@@ -47,10 +47,10 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
             if (isFromRepl)
             {
-                _executionService.ExecutePSCommandAsync(
+                await _executionService.ExecutePSCommandAsync(
                     new PSCommand().AddScript(request.Expression),
                     CancellationToken.None,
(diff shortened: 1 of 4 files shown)


andschwa:
@SeeminglyScience Could you test this out? I'm trying to think of / find any unintended consequences, but it sure _seems_ to have solved the problem.

andschwa:
Ditto @JustinGrote

andschwa:
@SeeminglyScience Have you had a chance to look at this?

SeeminglyScience:
I haven't gotten a chance to test myself yet. I'm inclined to say this will be fine though and was just an oversight. I believe previously we were purposefully not waiting for completion but I don't think the architecture requires it anymore.

One thing I'd check is that <kbd>F8</kbd>ing `$psEditor.GetEditorContext()` doesn't dead lock. That would be my main concern

andschwa:
> I haven't gotten a chance to test myself yet. I'm inclined to say this will be fine though and was just an oversight. I believe previously we were purposefully not waiting for completion but I don't think the architecture requires it anymore.
> 
> One thing I'd check is that F8ing `$psEditor.GetEditorContext()` doesn't dead lock. That would be my main concern

I couldn't get it to deadlock 👌

SeeminglyScience (approved):
LGTM!

PR #1691, opened 2022-02-01
Print prompt and command when `WriteInputToHost` is true

andschwa:
This fixes https://github.com/PowerShell/vscode-powershell/issues/3786 by simply printing the prompt before the input of the command run by F8. It's almost cosmetic-only, except it does invoke the user's `prompt` function again which could have side-effects. What's more, we'll still have an empty prompt before the invocation; however, I believe that to be an accurate representation of what's happening because that prompt is canceled per my explanation here: https://github.com/PowerShell/vscode-powershell/issues/3786#issuecomment-1026354061

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
@@ -1,16 +1,16 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.Logging;
-using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
-using Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility;
-using Microsoft.PowerShell.EditorServices.Utility;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Management.Automation;
 using System.Management.Automation.Remoting;
 using System.Threading;
+using Microsoft.Extensions.Logging;
+using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
+using Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility;
+using Microsoft.PowerShell.EditorServices.Utility;
 using SMA = System.Management.Automation;
 
 namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
@@ -49,7 +49,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
 
             if (PowerShellExecutionOptions.WriteInputToHost)
             {
-                _psesHost.UI.WriteLine(_psCommand.GetInvocationText());
+                _psesHost.WriteWithPrompt(_psCommand, cancellationToken);
             }
 
             return _pwsh.Runspace.Debugger.InBreakpoint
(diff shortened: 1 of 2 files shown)


andschwa:
@PrzemyslawKlys, @JustinGrote, and @SeeminglyScience what do you think of this? I'm not sure I like it to be quite honest.

PrzemyslawKlys:
Current behavior of prod version is:
- if the prompt is empty - just use it
- if the prompt has anything in it - insert everything above it, and move the text user typed after it

image

- i have first typed chars, then pressed F5 and that's the result

This seems most obvious.

SeeminglyScience:
LGTM to be honest. The side effects of running `prompt` again are fine imo. We are essentially pretending the user cancelled out of their current REPL, ran the command they <kbd>F8</kbd>'d and then restored the REPL.

JustinGrote:
The only other way it could work is to capture the user's prompt every time and then replicate the characters as-is to avoid side-effects. I think if you have a prompt that isn't idempotent youre going to have other problems and this would just reveal them, so I don't think it's that big of a deal. LGTM, this is how I would have solved it too :)

SeeminglyScience:
We are cancelling PSReadLine for this right? When we do, it saves the current REPL input for us to restore later. I believe that's done automatically by PSRL.

All we should need to do is:

1. Trigger the cancellation token source
2. Wait for PSRL to cancel and do not restart it
3. Write a new prompt
4. Write invocation text
5. Invoke selected lines
6. Restart ReadLine

That should result in the desired behavior.

andschwa:
@SeeminglyScience There is something wrong with it. I'm debugging. I managed to accidentally reproduce https://github.com/PowerShell/vscode-powershell/issues/3751 while just trying to manually save and then restore the user's input, not knowing what you just told me about PSReadLine magically doing it for us.

andschwa:
> * Wait for PSRL to cancel and do not restart it

@SeeminglyScience As far as I can see we do not _wait_ for PSRL to cancel. Only entry-point is:

https://github.com/PowerShell/PowerShellEditorServices/blob/f890a758729279658d0f6a31ec0d42baf735399d/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L675-L678

And it's just being synchronously called:

https://github.com/PowerShell/PowerShellEditorServices/blob/f890a758729279658d0f6a31ec0d42baf735399d/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L623-L646

SeeminglyScience:
> @SeeminglyScience As far as I can see we do not _wait_ for PSRL to cancel. Only entry-point is:

That very well could be fine as well. The only reason we had to do that before was because `ReadLine` was tied to the an execution status event. So as long as the pipeline execution thread is moving to our new evaluate task after `ReadLine` cancellation completes we're still good to go. We should just need the evaluate task to write `prompt` and invocation text.

andschwa:
> We should just need the evaluate task to write `prompt` and invocation text.

This part is fine and easy. It's saving the existing input (if there is any) and then restoring it that's proving tricky.

SeeminglyScience:
> This part is fine and easy. It's saving the existing input (if there is any) and then restoring it that's proving tricky.

Ah right that code is already where I was suggesting. For some reason I thought it was in the `EvaluateHandler` itself instead.

I'll have to dive into the code tonight to see where it's being cancelled, how it's being restored, and see if I can figure out what's going wrong there.

andschwa:
Umm nevermind...it's working perfectly once I combined it with my fix in #1690

andschwa:
WTF now it's not working. What did I do!?

SeeminglyScience:
Consistently? Or we got one of them super fun race conditions goin' on?

andschwa:
Super fun race condition, it's half working. Mostly works, can sometimes cause it to lose the input.

andschwa:
It definitely works much better than it was, and the occasional loss of input is not due to this change but is an extant bug, so marking as ready for review.

SeeminglyScience (approved):
LGTM!

PR #1693, opened 2022-02-01
Replace superfluous `Default` member with default values

andschwa:
I think this was a case of premature optimization. We're using C#
records, which should be as cheap as an object as they come, especially
when we can rely on the compiler for default values. Now we're passing
fewer objects around too.

### DIFF
--- a/src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
+++ b/src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
@@ -107,7 +107,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Extension
             // Register the editor object in the runspace
             return ExecutionService.ExecuteDelegateAsync(
                 $"Create ${PSEditorVariableName} object",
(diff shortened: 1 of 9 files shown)


andschwa:
No clue why that CI failed, everything passes locally. Rerunning.

SeeminglyScience:
> We're using C# records, which should be as cheap as an object as they come

You just mean that there isn't likely to be a lot of initialization involved right? Just clarifying because a record is just sugar for a class with some extra methods and an interface defined. It's still extra heap allocations.

That said, it's just some extra heap allocations that are unlikely to even be measurable so if you prefer it I don't feel strongly either way.

SeeminglyScience (approved):
LGTM (assuming the CI failures is sorted)

andschwa:
Damn, I want to know what's up with the CI...it doesn't fail locally 😅

andschwa:
@SeeminglyScience Tracked down the source of the failure and it was because we actually needed to do the null coalescing in the `SynchronousTask` constructors.

PR #1700, opened 2022-02-08
Clean up LSP message tests

andschwa:
Just applying automatic cleanups.

### DIFF
--- a/test/PowerShellEditorServices.Test.E2E/DebugAdapterClientExtensions.cs
+++ b/test/PowerShellEditorServices.Test.E2E/DebugAdapterClientExtensions.cs
@@ -14,23 +14,24 @@ namespace PowerShellEditorServices.Test.E2E
     {
         public static async Task LaunchScript(this DebugAdapterClient debugAdapterClient, string filePath, TaskCompletionSource<object> started)
         {
-            LaunchResponse launchResponse = await debugAdapterClient.Launch(new PsesLaunchRequestArguments
-            {
-                NoDebug = false,
-                Script = filePath,
-                Cwd = "",
-                CreateTemporaryIntegratedConsole = false,
-            }).ConfigureAwait(false);
+            LaunchResponse launchResponse = await debugAdapterClient.Launch(
+                new PsesLaunchRequestArguments
+                {
+                    NoDebug = false,
+                    Script = filePath,
+                    Cwd = "",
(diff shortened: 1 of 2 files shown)


test/PowerShellEditorServices.Test.E2E/DebugAdapterClientExtensions.cs
```diff
@@ -14,23 +14,24 @@ public static class DebugAdapterClientExtensions
...
+            if (launchResponse is null)
             {
                 throw new Exception("Launch response was null.");
             }
 
             // This will check to see if we received the Initialized event from the server.
             await Task.Run(
-                async () => await started.Task.ConfigureAwait(false),
-                new CancellationTokenSource(2000).Token).ConfigureAwait(false);
+                async () => await started.Task.ConfigureAwait(true),
```
JamesWTruher, line 33: since you're changing the behavior of the ConfigureAwait consider a comment to explain why the advice of https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.configureawait?view=net-6.0#remarks isn't applicable.

test/PowerShellEditorServices.Test.E2E/LanguageServerProtocolMessageTests.cs
```diff
@@ -51,15 +51,14 @@ public LanguageServerProtocolMessageTests(ITestOutputHelper output, LSPTestsFixt
...
-            _fixture = data;
-
             PwshExe = PsesStdioProcess.PwshExe;
         }
 
         public void Dispose()
         {
             Diagnostics.Clear();
             TelemetryEvents.Clear();
+            GC.SuppressFinalize(this);
```
PaulHigin, line 61: Should this ever run on a finalizer thread?  Is there a class destructor to finalize?  Just wondering if this is needed.

PaulHigin (approved):

test/PowerShellEditorServices.Test.E2E/LanguageServerProtocolMessageTests.cs
andschwa: I had a long conversation with Dongbo to learn all about this and while this doesn't cause a bug, the compiler barfs a warning about it, and this is the "correct" fix. If this were product code it would be more important, but I'm working to just get us to a zero-warning codebase. Slowly, but surely.

PR #1703, opened 2022-02-11
Use `static readonly` for default `ExecutionOptions`

andschwa:
But at the `SynchronousTask` level. Now we get the best of both worlds:
the records have the correct default values on initialization, and we
only heap allocate a single static instance of `ExecutionOptions` for
the default case.


andschwa:
@PaulHigin Originally I was fixing up `DebugService.cs` to set `AddToHistory = false`, thinking that was the issue; however, that was the default and instead I discovered that those `InvocationSettings` aren't passed at all in `ExecuteInDebugger`. Anyway, while working on that I cleaned this up a little bit.

andschwa:
Reviewed this over the phone with Paul, merging!

andschwa:
Err, GitHub still has that bug we're I can't squash and merge if I use override as admin privileges.

PaulHigin (approved):

PR #1704, opened 2022-02-11
Avoid recording debugger commands in the history

andschwa:
This changes how we decide to run a given command through
`ExecuteInDebugger` or `ExecuteNormally`. It now restricts the former
such that it is only used for PowerShell's intrinsic debugger commands
(like `s`, `c`, etc.) and the latter is used for any other user command
or, more importantly, our own implementation's debugger commands (like
when we call `Get-Variable`).

### DIFF
--- a/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
+++ b/test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
@@ -586,10 +586,9 @@ namespace Microsoft.PowerShell.EditorServices.Test.Debugging
             VariableDetailsBase[] variables = GetVariables(VariableContainerDetails.LocalScopeName);
 
             // Test set of a local string variable (not strongly typed)
-            const string newStrValue = "Goodbye";
+            const string newStrValue = "\"Goodbye\"";
             VariableScope localScope = Array.Find(scopes, s => s.Name == VariableContainerDetails.LocalScopeName);
-            // TODO: Fix this so it has the second quotes again?
-            string setStrValue = await debugService.SetVariableAsync(localScope.Id, "$strVar", '"' + newStrValue + '"').ConfigureAwait(true);
+            string setStrValue = await debugService.SetVariableAsync(localScope.Id, "$strVar", newStrValue).ConfigureAwait(true);
             Assert.Equal(newStrValue, setStrValue);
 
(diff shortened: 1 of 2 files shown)


andschwa:
The [debugger commands](https://github.com/PowerShell/PowerShell/blob/7dc4587014bfa22919c933607bf564f0ba53db2e/src/System.Management.Automation/engine/debugger/debugger.cs#L5032-L5049) haven't changed in eons (at least not since Sergei and I imported Source Depot to GitHub. I'd _love_ to reference them directly but this super helpful class is marked `internal` (which is also why I can't use the [`ProcessCommand`](https://github.com/PowerShell/PowerShell/blob/7dc4587014bfa22919c933607bf564f0ba53db2e/src/System.Management.Automation/engine/debugger/debugger.cs#L5103-L5106) which accepts an `InvocationSettings`).

andschwa:
Uh oh things are broken.

andschwa:
Interesting...

```
[xUnit.net 00:00:16.40]     DebuggerSetsVariablesNoConversion [FAIL]
  Failed DebuggerSetsVariablesNoConversion [910 ms]
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: Goodbye
Actual:   "Goodbye"
          ↑ (pos 0)
```

I actually wonder if this fixes a regression in the tests when we did the pipeline rewrite.

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -581,10 +581,9 @@ public async Task DebuggerSetsVariablesNoConversion()
             VariableDetailsBase[] variables = GetVariables(VariableContainerDetails.LocalScopeName);
 
             // Test set of a local string variable (not strongly typed)
-            const string newStrValue = "Goodbye";
+            const string newStrValue = "\"Goodbye\"";
             VariableScope localScope = Array.Find(scopes, s => s.Name == VariableContainerDetails.LocalScopeName);
-            // TODO: Fix this so it has the second quotes again?
```
andschwa, line 586: The quotes are back!

andschwa:
Need to test with PowerShell 5.1.

PaulHigin (approved):

PR #1713, opened 2022-02-16
Allow multiple code actions if there is more than one SuggestedCorrection of a diagnostic record returned by PSSA

bergmeister:
Currently, PSSA's `SuggestCorrections` property of the `DiagnosticRecord` object already is an array but no built-in rules return more than one suggested correction, which lead to `PowerShellEditorServices` not correctly translating this array before returning it as an code action but it only ever displayed one. This fixes it makes it generic again so it returns a code action for each suggest correction. It's basically just performing a foreach loop and calling `ToTextEdit` just once, a diff that ignored whitespace due to increased indentation will make this much clearer
This is a pre-requisite for an implementation of this, where a built-inrule will return multiple suggest corrections: https://github.com/PowerShell/PSScriptAnalyzer/issues/1767

I've manually tested this with a modified version of PSSA where I return 2 suggested corrections and when I build the PS extension with this change, the UI shows me the two different suggested code actions and also applies them correctly (as in they are different as they should be by design)

### DIFF
--- a/src/PowerShellEditorServices/Services/TextDocument/Handlers/CodeActionHandler.cs
+++ b/src/PowerShellEditorServices/Services/TextDocument/Handlers/CodeActionHandler.cs
@@ -81,7 +81,7 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
                 string diagnosticId = AnalysisService.GetUniqueIdFromDiagnostic(diagnostic);
                 if (corrections.TryGetValue(diagnosticId, out MarkerCorrection correction))
                 {
-                    foreach(ScriptRegion edit in correction.Edits)
+                    foreach (ScriptRegion edit in correction.Edits)
                     {
                         codeActions.Add(new CodeAction
                         {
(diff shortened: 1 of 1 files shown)


andschwa:
Thanks @bergmeister! Just to double check, this is backwards compatible right? Like it won't break things with the current version of PSSA?

bergmeister:
Yes. Nothing is changing in PSSA's interface, which is the DiagnosticRecord object. This PR fixes PSES to actually treat the SuggestedCorrections property as an array instead and therefore also returning a code action for each one instead of just only one for the first one. Even with the current version of PSSA one can write a custom rule that returns multiple suggested corrections. Just to be sure I also tested that the code action for the most common case of diagnostic records having just one suggested correction still works.

andschwa:
Uh, weird, I tried to cherry pick a13202cb487285766d04479edf8a3cf21086a4c9 and got `fatal: bad object a13202cb487285766d04479edf8a3cf21086a4c9`

bergmeister:
Weird indeed, have you tried doing that in a new clone as well in case the git db of your everyday PSES got issues. Or maybe there was a commit hash conflict since my fork is a different repo and although chances are very low, thery are not zero...
Well the change is small, so manually copying it over to another branch is fine anyway

andschwa:
@bergmeister Not a clue! I did cherry-pick over to here if you want to double check https://github.com/PowerShell/PowerShellEditorServices/pull/1718

andschwa:
Fixed in https://github.com/PowerShell/PowerShellEditorServices/pull/1718! Closing. Thanks @bergmeister!

andschwa:
This is out in [PowerShell Preview v2022.2.2](https://marketplace.visualstudio.com/items?itemName=ms-vscode.powershell-preview)!

PR #1716, opened 2022-02-17
Use global `dotnet` and update build dependencies

andschwa:
This removes our build script's dependency on the `install-dotnet` script (which is only meant for use in CI, but we now use an ADO task for that, and so that we can rely on pre-installed versions of .NET if available). Developers will need to follow .NET Core's instructions to install .NET SDK 6.0, and the 3.1 runtime for the end-to-end tests. Fortunately, this actually eases the developer workflow by no longer being sensitive to a local installation of `dotnet`, so the C# extension in Code now "just works."

This updates the release pipeline to use the `PMMSWindows2019-Secure` for build, sign, and release.

This also updates our required and installed `InvokeBuild` and `platyPS` modules for the build, which is probably the most annoying part of this PR as devs will need to run `Update-Module` for these (and possibly install `playPS` if they had never run the dependent task before).

### DIFF
--- a/.vsts-ci/azure-pipelines-release.yml
+++ b/.vsts-ci/azure-pipelines-release.yml
@@ -29,7 +29,8 @@ stages:
   jobs:
   - job: Build
     pool:
-      vmImage: windows-2019
+      name: 1ES
+      demands: ImageOverride -equals PSMMS2019-Secure
     steps:
     - template: templates/ci-general.yml
 
@@ -39,7 +40,7 @@ stages:
   - job: Sign
     pool:
       name: 1ES
-      demands: ImageOverride -equals MMSWindows2019-Secure
+      demands: ImageOverride -equals PSMMS2019-Secure
     variables:
     - group: ESRP
     steps:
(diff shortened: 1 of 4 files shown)


.vsts-ci/azure-pipelines-release.yml
```diff
@@ -29,7 +29,8 @@ stages:
   jobs:
   - job: Build
     pool:
-      vmImage: windows-2019
+      name: 1ES
+      demands: ImageOverride -equals PMMSWindows2019-Secure
```
TravisEz13, line 33:
```suggestion
      demands: ImageOverride -equals PMMS2019-Secure
```

.vsts-ci/azure-pipelines-release.yml
```diff
@@ -39,7 +40,7 @@ stages:
   - job: Sign
     pool:
       name: 1ES
-      demands: ImageOverride -equals MMSWindows2019-Secure
+      demands: ImageOverride -equals PMMSWindows2019-Secure
```
TravisEz13, line 43:
```suggestion
      demands: ImageOverride -equals PMMS2019-Secure
```

PR #1718, opened 2022-02-19
Return a code action for each diagnostic record

andschwa:
Currently, PSSA's `SuggestCorrections` property of the
`DiagnosticRecord` object already is an array, but no built-in rules
return more than one suggested correction, which led to
`PowerShellEditorServices` not correctly translating this array before
returning it as an code action as it only ever displayed one. This fixes
it by making it generic again so it returns a code action for each
suggest correction. It's basically just performing a `foreach` loop and
calling `ToTextEdit` just once.

This is a pre-requisite for an implementation of this, where a built-in
rule will return multiple suggest corrections:
https://github.com/PowerShell/PSScriptAnalyzer/issues/1767

I've manually tested this with a modified version of PSScriptAnalyzer
where I return two suggested corrections. The extension's UI now shows
me the two different suggested code actions and also applies them
correctly.

Replaces #1713.

### DIFF
--- a/src/PowerShellEditorServices/Services/TextDocument/Handlers/CodeActionHandler.cs
+++ b/src/PowerShellEditorServices/Services/TextDocument/Handlers/CodeActionHandler.cs
@@ -3,14 +3,12 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.PowerShell.EditorServices.Services;
 using Microsoft.PowerShell.EditorServices.Services.TextDocument;
 using Microsoft.PowerShell.EditorServices.Utility;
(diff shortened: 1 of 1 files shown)


SeeminglyScience (approved):
LGTM

bergmeister:
Sorry, forgot about this one. Since this change is slightly different, will re-test just to confirm

andschwa:
The actual changes should be no different, I cherry-picked the commit: https://github.com/PowerShell/PowerShellEditorServices/pull/1718/commits/a4c4fdf5cb19244c026dff9753becae361d284d6

bergmeister:
Ok, just seeing the other changes made me think there is a chance it could affect it. I tested with the latest preview extension since it's already released and can confirm it still works the way I tested it with my own build :-)

andschwa:
Yay! Thanks for testing!!

PR #1719, opened 2022-02-19
Reset progress messages at end of REPL

andschwa:
This seems to work. We needed to hold onto processed progress records,
and then at the end of the REPL mark each as completed and re-write it
(which uses the underlying host and effectively clears it).

Fixes https://github.com/PowerShell/vscode-powershell/issues/3807

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostUserInterface.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostUserInterface.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Management.Automation;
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostUserInterface.cs
```diff
@@ -22,6 +22,8 @@ internal class EditorServicesConsolePSHostUserInterface : PSHostUserInterface
 
         private readonly PSHostUserInterface _consoleHostUI;
 
+        private readonly HashSet<(long, ProgressRecord)> _currentProgressRecords = new();
```
SeeminglyScience, line 25: `ProgressRecord.GetHashCode` is just the default object impl, so in most cases we'd be storing a `ProgressRecord` for every update sent. If the goal is to avoid the previous work around regarding the dummy string, a `Dictionary<long, ProgressRecord>` is probably needed.

SeeminglyScience (changes requested):

src/PowerShellEditorServices/Services/PowerShell/Host/EditorServicesConsolePSHostUserInterface.cs
andschwa: I'd actually just stored the tuple cause we need both items later, the `long` as the source ID and the record to mark as completed. Should be fine as-is in a hash set right? I'm not sure if the source IDs are unique so they may not work well as keys in a dictionary.

andschwa: Fixed up!

andschwa:
This seems to be working now, and it's very close to how it was 😂

SeeminglyScience (approved):
LGTM

PR #1724, opened 2022-02-23
Re-enable line breakpoints for untitled scripts

andschwa:
We managed to make the previous hack work while continuing to support
passing the users' arguments. As there was demand for this feature to
continue working, despite being a hack, we're keeping it.


SeeminglyScience (approved):
LGTM!

Ty for fully detailing the jank in the comments so it's easier to tell why it is the way it is ❤️

JustinGrote (approved):
LHTM (looks hacky to me), thanks for taking the time to doc as already mentioned.

PR #1725, opened 2022-02-23
Fix editor commands to interrupt current prompt

andschwa:
Also refactored it slightly to accept the existing `cancellationToken` send to the handler, and removed an unused logger service. This fixes https://github.com/PowerShell/vscode-powershell/issues/3691.

### DIFF
--- a/src/PowerShellEditorServices/Services/Extension/Handlers/InvokeExtensionCommandHandler.cs
+++ b/src/PowerShellEditorServices/Services/Extension/Handlers/InvokeExtensionCommandHandler.cs
@@ -4,40 +4,28 @@
 using System.Threading;
 using System.Threading.Tasks;
 using MediatR;
-using Microsoft.Extensions.Logging;
 using Microsoft.PowerShell.EditorServices.Extensions;
 
 namespace Microsoft.PowerShell.EditorServices.Services.Extension
 {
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices/Services/Extension/Handlers/InvokeExtensionCommandHandler.cs
```diff
@@ -4,40 +4,28 @@
...
-            // We can now await here because we handle asynchronous message handling.
-            EditorContext editorContext =
-                _editorOperationsService.ConvertClientEditorContext(
-                    request.Context);
-
-            await _extensionService.InvokeCommandAsync(
-                request.Name,
-                editorContext).ConfigureAwait(false);
-
-            return await Unit.Task.ConfigureAwait(false);
```
andschwa, line 40: No idea why this was the old return type. Almost everything else using the same API returns `Unit.Value` (which seems the right thing to return).

andschwa: Ah, clarification on that "almost" it's that `async Task<Unit>` with an `await` needs to return `Unit.Value`, and if it's not `async` then it's a `Unit.Task` that needs to be returned.

SeeminglyScience (approved):
LGTM

PR #1726, opened 2022-02-24
Avoid error when `exclude` entry is a clause

andschwa:
At this point it just ignores the entries that are clauses instead of a straight boolean, but hey now it doesn't crash 😄 

Fixes https://github.com/PowerShell/vscode-powershell/issues/2658


SeeminglyScience (approved):
LGTM!

PR #1728, opened 2022-02-24
Delete unnecessary `ConsoleColor` files

andschwa:
No longer necessary with the new host implementation.


SeeminglyScience (approved):
Nice catch, LGTM!

PR #1729, opened 2022-02-25
WIP: Wrap `ReadLine` with lock and fallback to legacy when re-entering

andyleejordan:
Just me trying things. So, weird thing, but in the repro for https://github.com/PowerShell/vscode-powershell/issues/3751, while we're in the debugger, things can't print. Like if you run `Get-Process` at the debug prompt it shows nothing, and if you're debugging the pwsh process you see a million exceptions. This gets resolved when I set the host explicitly in the new `PSInvocationSettings`. When we either don't create the linked cancellation token source, or when we move its scope outside the function its currently created in, the repro issue goes away. The monitor logic I guess is useless 🥲

Play around with it @SeeminglyScience?

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.Logging;
 using System;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
 {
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
```diff
@@ -79,16 +80,16 @@ public void ExecuteSynchronously(CancellationToken executorCancellationToken)
...
+            if (executorCancellationToken.IsCancellationRequested)
             {
                 SetCanceled();
                 return;
             }
 
             try
             {
-                TResult result = Run(cancellationSource.Token);
+                TResult result = Run(executorCancellationToken);
```
SeeminglyScience, line 92: Alright so think of a linked cancellation token source kinda like this:

```csharp
static void MyMethod(CancellationToken callersToken)
{
    CancellationTokenSource myToken = new(millisecondsDelay: 5000);

    // Real method name is "Register", but bare with me
    callersToken.OnCancel(() => myToken.Cancel());
    OtherMethod(myToken.Token);
}
```

That's all linking does. The issue with the WIP here is instead of passing `myToken.Token`, you're passing `callersToken`. So if we call `myToken.Cancel()`, that's not going to do anything. That's also why it looks like it works: it's never cancelled.

I haven't played with it yet, but I would guess we can no longer cancel PSRL like this (e.g. F8/editor commands/continue). If we *can* still cancel PSRL, we should probably look at references for `_taskRequesterCancellationToken` to see if we ever even cancel it.

PR #1732, opened 2022-02-28
Redirect `PSHost.Notify*Application` methods

### LINKED ISSUE #1639: Cursor position is set with wrong values after finish running a git command in integrated console
### Summary

Cursor position contains the wrong values after the prompt comes back from running a `git` command, such as `git push`, `git log`, and `git branch`. But it's fine after running `git status`. See the screen GIF below.

Please be noted that, the color of `*regex` in the prompt is different after running `git push`, `git log` and `git branch`, but the color is correct after running `git status`. When the color of `*regex` becomes different, you will observe the misplacing of cursor when typing.

Animation

However, things are working fine within the default `pwsh` in the same VSCode terminal (not the integrated console). See the GIF below.
The default `pwsh` is the same version as the integrated console `7.2.0`, and uses the same version of `PSReadLine`.

Animation2

### Investigation

I tried to find what's happening within PSReadLine under a debugger, and I found that, when `PSConsoleReadLine.ReadLine` gets called after finish running `git log -1` or `git push`, `cosnole.CursorLeft == 19`, even though the cursor-left looks to be at `7` based on what's shown on the terminal. It looks the terminal state somehow gets corrupted after running `git push`, `git log`, or `git branch`.

#### PSES Extension Version

```
PS:5> $host  

Name             : Visual Studio Code Host
Version          : 2021.11.1
InstanceId       : 3c0bc9b5-f6b2-45be-96c8-44ed2d1c767f
UI               : System.Management.Automation.Internal.Host.InternalHostUserInterface
CurrentCulture   : en-US
CurrentUICulture : en-US
PrivateData      : 
DebuggerEnabled  : True
IsRunspacePushed : False
Runspace         : System.Management.Automation.Runspaces.LocalRunspace
```

#### PSReadLine version

```
PS:15> gmo PSReadLine

ModuleType Version    PreRelease Name
---------- -------    ---------- ----
Script     2.2.0      beta4      PSReadLine
```

SeeminglyScience:
Fixes #1639

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -168,12 +168,12 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
 
         public override void NotifyBeginApplication()
         {
-            // TODO: Work out what to do here
+            _hostInfo.PSHost.NotifyBeginApplication();
         }
 
         public override void NotifyEndApplication()
         {
-            // TODO: Work out what to do here
+            _hostInfo.PSHost.NotifyEndApplication();
         }
 
         public void PopRunspace()


andschwa (approved):
😂

PR #1734, opened 2022-02-28
Finish redirection of `$Host.PrivateData`

andschwa:
We missed the middle layer: public host redirected to internal host, but
then internal host needed to redirect to `_hostInfo.PSHost`. Now that it
does, `$Host.PrivateData` actually shows the correct console color
values, and so `Write-Error 'something'` shows up in the correct color.

Also applied Roslynator suggestions to clean some things up.


SeeminglyScience (approved):
LGTM

PR #1738, opened 2022-03-08
Improve completion logic

andschwa:
This makes the deduced `CompletionItemKind` enum more accurate, which
results in better icons for the user when viewing completions in an
editor. Also cleaned up the file as there were remnants of dead code.

Fixes https://github.com/PowerShell/vscode-powershell/issues/3364


andschwa:
@MartinGC94 I think you'll like this!

From your demo (ignore `ProviderItem`, Code does some magic to get filetype specific icons later):
Screen Shot 2022-03-07 at 4 42 07 PM

Files:
Screen Shot 2022-03-07 at 4 44 43 PM

Operators:
Screen Shot 2022-03-07 at 4 46 06 PM

Namespace:
Screen Shot 2022-03-07 at 4 43 38 PM

And even classes:
Screen Shot 2022-03-07 at 4 43 09 PM

MartinGC94:
You are right, this looks great and great choice of icons as well. The only icon that looks a little out of place is the ParameterName. I know it shares its icon with the operators but parameters are used much more frequently so maybe the chosen icon should look more like a parameter?

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -110,13 +104,9 @@ public async Task<CompletionItem> Handle(CompletionItem request, CancellationTok
             }
 
             // Get the documentation for the function
-            CommandInfo commandInfo =
-                await CommandHelpers.GetCommandInfoAsync(
-                    request.Label,
-                    _runspaceContext.CurrentRunspace,
-                    _executionService).ConfigureAwait(false);
+            CommandInfo commandInfo = await CommandHelpers.GetCommandInfoAsync( request.Label, _runspaceContext.CurrentRunspace, _executionService).ConfigureAwait(false);
```
SeeminglyScience, line 107:
```suggestion
            CommandInfo commandInfo = await CommandHelpers.GetCommandInfoAsync(request.Label, _runspaceContext.CurrentRunspace, _executionService).ConfigureAwait(false);
```

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -323,43 +304,33 @@ public void SetCapability(CompletionCapability capability, ClientCapabilities cl
...
+                CompletionType.Property => CompletionItemKind.Property,
+                CompletionType.Variable => CompletionItemKind.Variable,
+                CompletionType.Namespace => CompletionItemKind.Module,
+                CompletionType.Type => CompletionItemKind.TypeParameter,
+                CompletionType.Keyword => CompletionItemKind.Keyword,
+                CompletionType.File => CompletionItemKind.File,
+                CompletionType.Folder => CompletionItemKind.Folder,
+                CompletionType.History => CompletionItemKind.Reference,
+                CompletionType.Text => CompletionItemKind.Text,
+                _ => throw new NotImplementedException(),
```
SeeminglyScience, line 327:
```suggestion
                _ => throw new ArgumentOutOfRangeException(nameof(completionType)),
```

`NotImplementedException` is usually for interface implementations that are explicitly not implemented.

SeeminglyScience (approved):
Looks great! Just a couple of nits

andschwa:
> You are right, this looks great and great choice of icons as well. The only icon that looks a little out of place is the ParameterName. I know it shares its icon with the operators but parameters are used much more frequently so maybe the chosen icon should look more like a parameter?

@MartinGC94 That is an artifact of the way that test displays all of them at once and the logic that figures out if it's a parameter. They're actually currently mapped to `CompletionItemKind.Variable` (which was their previous mapping) and look like this:
Screen Shot 2022-03-08 at 9 47 05 AM

However, I'm trying to decide if instead they should map to `Field` or `Property` (which are maybe slightly more accurate? The parameters declared on a PowerShell function are kind of like fields/properties).

Field (which is so far unused):
Screen Shot 2022-03-08 at 9 50 11 AM

Property (which already has the PowerShell `CompletionResultType.Property` mapping to it):
Screen Shot 2022-03-08 at 9 51 57 AM

MartinGC94:
I see. All 3 of those are fine but if I had to pick I would go with Field because I like the icon and the fact that it's unique.

JustinGrote:
Using other languages as prior art, parameters are generally considered variables (Typescript as example):

image
image

C# as another:
image


So I think it should stay as variable for a consistent experience.

andschwa:
Thanks for pointing that out @JustinGrote, I agree. What tool are you using to inspect like that?

JustinGrote:
@andschwa it's built into vscode, very useful for troubleshooting syntax highlighting
image

andschwa:
> @andschwa it's built into vscode, very useful for troubleshooting syntax highlighting
> image

I'm so sad I've learned about this a year later...what else should I know?!

JustinGrote:
@andschwa uuuh, you can access devtools for the running vscode instance maybe?
image

Basically check out everything prefixed with `Developer:`

andschwa:
@JustinGrote @SeeminglyScience @MartinGC94 Give this new commit a whirl would you? I removed the middle abstraction layer as it seemed superfluous.

andschwa:
Oops, probably tests expecting old code. Fixing up.

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -97,34 +80,32 @@ public static bool CanResolve(CompletionItem value)
...
-            {
-                return request;
-            }
-
-            // No details means the module hasn't been imported yet and Intellisense shouldn't import the module to get this info.
-            if (request.Detail is null)
+            // We currently only support this request for anything that returns a CommandInfo:
+            // functions, cmdlets, aliases. No detail means the module hasn't been imported yet and
+            // IntelliSense shouldn't import the module to get this info.
+            if (request.Kind != CompletionItemKind.Function || request.Detail is null)
```
SeeminglyScience, line 86:
```suggestion
            if (request.Kind is not CompletionItemKind.Function || request.Detail is null)
```

nit: `is` can be used whenever the RHS is a constant

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -149,217 +130,170 @@ public void SetCapability(CompletionCapability capability, ClientCapabilities cl
...
-        private static CompletionItemKind MapCompletionKind(CompletionType completionType)
+        /// <summary>
+        /// Look for type encoded in the tooltip for parameters and variables. Display PowerShell
+        /// type names in [] to be consistent with PowerShell syntax and how the debugger displays
+        /// type names.
+        /// </summary>
+        /// <param name="toolTipText"></param>
+        /// <param name="type"></param>
+        /// <returns>Whether or not the type was found.</returns>
+        private static bool ExtractTypeFromToolTip(string toolTipText, out string type)
```
SeeminglyScience, line 256: nit: For this pattern usually the convention in C# is to name it `TryEtc`

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -149,217 +130,170 @@ public void SetCapability(CompletionCapability capability, ClientCapabilities cl
...
         {
-            switch (completionType)
+            Regex _typeRegex = new(@"^(\[.+\])");
```
SeeminglyScience, line 258:
```suggestion
            Regex typeRegex = new(@"^(\[.+\])");
```

For locals skip `_` (I know it's to mirror the real field but still, better to just reference the new local)

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -149,217 +130,170 @@ public void SetCapability(CompletionCapability capability, ClientCapabilities cl
...
+            MatchCollection matches = _typeRegex.Matches(toolTipText);
```
SeeminglyScience, line 259:
```suggestion
            MatchCollection matches = typeRegex.Matches(toolTipText);
```

SeeminglyScience (approved):
LGTM, couple of little things

PR #1746, opened 2022-03-22
Replace `_consoleHostUI` with `_underlyingHostUI`

andschwa:
And stop using reflection to find `_externalUI`. This is not necessary in order to reuse PowerShell's built-in prompt logic (which was its original goal). Unfortunately this doesn't fix the issues it was written in an attempt to fix, but it does remove unnecessary reflection logic.


dkattan:
While you're at it, can we have the ability to bring our own ReadLine and ReadKey and not forcefully use the given implementations? The current implementations ultimately hardcode Console.ReadKey()

andschwa:
@dkattan Definitely something I'm looking at!

SeeminglyScience (approved):
LGTM

PR #1749, opened 2022-03-30
Correctly map `SuggestedCorrection` to `MarkerCorrection`

bergmeister:
With the recent fix in PR #1718, PSES processes now all Correction objects from PSSA but the message specifically was just taken from the last correction [here](https://github.com/PowerShell/PowerShellEditorServices/blob/e555dc050d3ef36272ba6f84925086cb7abc048a/src/PowerShellEditorServices/Services/TextDocument/ScriptFileMarker.cs#L150). I did not notice this at first because I thought I had to tweak my rule first to emit two different messages until I realized it was another bug in PSES.
Related: https://github.com/PowerShell/PSScriptAnalyzer/pull/1782
Proof here that the message of each correction is associated to it as well (the actual code fix already was)

BEFORE
image

AFTER
image

### DIFF
--- a/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
@@ -402,7 +402,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
                 Diagnostic diagnostic = GetDiagnosticFromMarker(marker);
 
-                if (marker.Corrections != null)
+                if (marker.Corrections is not null)
                 {
                     string diagnosticId = GetUniqueIdFromDiagnostic(diagnostic);
                     fileCorrections.Corrections[diagnosticId] = marker.Corrections;
--- a/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
@@ -208,7 +208,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
             ScriptFileMarker[] analysisResults = await AnalysisEngine.AnalyzeScriptAsync(functionText, commentHelpSettings).ConfigureAwait(false);
 
-            if (analysisResults.Length == 0 || analysisResults[0].Corrections.Any())
+            if (analysisResults.Length == 0 || !analysisResults[0].Corrections.Any())
             {
                 return null;
             }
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
```diff
@@ -404,10 +402,10 @@ private void PublishScriptDiagnostics(ScriptFile scriptFile, IReadOnlyList<Scrip
 
                 Diagnostic diagnostic = GetDiagnosticFromMarker(marker);
 
-                if (marker.Correction != null)
+                if (marker.Corrections != null)
```
andschwa, line 405:
```suggestion
                if (marker.Corrections is not null)
```

src/PowerShellEditorServices/Services/TextDocument/ScriptFileMarker.cs
```diff
@@ -124,32 +123,25 @@ internal static ScriptFileMarker FromDiagnosticRecord(PSObject psObject)
...
+                        Name = suggestedCorrection.Description ?? diagnosticRecord.Message,
+                        Edit = new ScriptRegion(
+                                diagnosticRecord.ScriptPath,
+                                suggestedCorrection.Text,
+                                startLineNumber: suggestedCorrection.StartLineNumber,
+                                startColumnNumber: suggestedCorrection.StartColumnNumber,
+                                endLineNumber: suggestedCorrection.EndLineNumber,
+                                endColumnNumber: suggestedCorrection.EndColumnNumber,
+                                startOffset: -1,
+                                endOffset: -1),
```
andschwa, line 142:
```suggestion
                        Edit = new ScriptRegion(
                                diagnosticRecord.ScriptPath,
                                suggestedCorrection.Text,
                                startLineNumber: suggestedCorrection.StartLineNumber,
                                startColumnNumber: suggestedCorrection.StartColumnNumber,
                                startOffset: -1,
                                endLineNumber: suggestedCorrection.EndLineNumber,
                                endColumnNumber: suggestedCorrection.EndColumnNumber,
                                endOffset: -1),
```

andschwa (changes requested):
Everything looks good, just one recurring test value. Some collection is coming through `null` and something isn't happy about it 😅 Thank you for working on this, excited to get it in!

bergmeister:
Yes, saw the test failure only after I opened the PR. Somehow, I cannot run the test with VS2022 with test explorer, can you give me some tips how I can run it locally to figure out why the value ends up being null by attaching the debugger and stepping through. Also happy to use VS Code instead alternatives

andschwa:
I exclusively use the C# extension in VS Code, it adds a "Run Test" and "Debug Test" code lens to each unit test, and it works great!

andschwa:
Good catch!

andschwa (approved):

andschwa:
Hey @bergmeister, I've added a PSSA test suite, what do you think a regression test for this would look like? See: https://github.com/PowerShell/PowerShellEditorServices/blob/main/test/PowerShellEditorServices.Test/Services/Symbols/PSScriptAnalyzerTests.cs

PR #1757, opened 2022-04-14
Enable code analysis and formatting as errors on build

andschwa:
By upgrading these from `warning` to `error` we get the equivalent of `dotnet format --verify-no-changes` without having to call it twice (which slowed down the build, and was unnecessary since the compiler is already running all the rules). Also enabled _yet more_ rules 😅


SeeminglyScience (approved):
LGTM!

Issue #1754, opened 2022-04-13
Resolve cancellation race with PSReadLine


andschwa:
When canceling the `ReadKey` method sent to PSReadLine, we observed that there's a race where PSReadLine is still consuming (and using) the dummy character we're returning when the method is canceled:

https://github.com/PowerShell/PowerShellEditorServices/blob/312051bbdaab5953f419c4f115ab4f322c512856/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L878-L908

While the chosen dummy character _should_ be a no-op, as observed in https://github.com/PowerShell/vscode-powershell/issues/3225#issuecomment-1097590406 it can come through as an `à` and in https://github.com/PowerShell/vscode-powershell/issues/3881#issuecomment-1097616974 it comes through as an `"`.

@daxian-dbw, @SeeminglyScience and I discovered this when trying to just use `default` for the returned value, and we think that around [here](https://github.com/PowerShell/PSReadLine/blob/f46f15d2d634e2060bc0eabe4c81fc13a5a64a3a/PSReadLine/ReadLine.cs#L97) "PSRL is probably checking if key returned, *then* checking if cancelled. Meaning if we have a key it doesn't check if cancelled."

PR #1758, opened 2022-04-14
With a fix in PSReadLine, we don't have to return a "null" key press

andschwa:
Which wasn't reliable anyway, because it could be translated to different things depending on the system and encodings in use. PSReadLine now just discards any key input after cancellation.

Requires https://github.com/PowerShell/PSReadLine/pull/3274 and fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1754.

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -6,6 +6,6 @@
         "Version": "1.1.3"
     },
     "PSReadLine": {
-        "Version": "2.2.2"
+        "Version": "2.2.3"
     }
 }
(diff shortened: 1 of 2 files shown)


andschwa:
This requires that the PR in PSReadLine is merged an a new release pushed out, staying in draft until then.

andschwa:
Ah ha! @SeeminglyScience now that `ReadLine` and `ReadKey` are well and truly canceled properly, I can revert the workaround in #1691 and it's getting much closer!
Screen Shot 2022-04-18 at 10 50 37 AM

andschwa:
The second commit is more of an aesthetic cleanup, it's not necessary to get that working perfectly to merge this, but we do need a patch release of PSReadLine, and I'd like to get this into the next RC.

daxian-dbw (approved):

Issue #1479, opened 2021-05-13
Auto-cleanup unused `using` statements


andschwa:
While reading through the codebase in #1474 I noticed a lot of warnings from OmniSharp about unnecessary using directives, but instead of fixing these one by one I would like to just use an automated tool to do is one swoop. I think OmniSharp can do this: https://github.com/OmniSharp/omnisharp-vscode/issues/1179

PR #1759, opened 2022-04-18
Enable IDE0005 (unneccessary using statements) as error

andschwa:
Due to a quirk in how `msbuild` runs rules, documentation generation
must be enabled for this rule (and presumably others) to be respected.

See https://github.com/dotnet/roslyn/issues/22579.

Fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1479.


andschwa:
Interesting...passed locally, looks like in other configs though I need to clean up some using statements! And disable a bunch of XML warnings I don't care about. Converted to draft.

SeeminglyScience (approved):
Awesome :) LGTM!

PR #1762, opened 2022-04-19
Fix prompt spam and general debugger reliability improvements

SeeminglyScience:
This should make default prompt spam significantly less likely or impossible. Also fixes a scenario here if the REPL loop is evaluating and in an attached process, closing the attachee would dead lock.

Also fixed a logic bug that probably caused some not yet seen issues.


andschwa (approved):
LGTM.

PR #1763, opened 2022-04-19
Temporarily disable `IsDebuggerCommand` check

andschwa:
And instead run any commands while debugging through `ExecuteInDebugger`. This means the history will be polluted, but since this fix isn't working 100%, we're not shipping it yet.

Should be the other side of the fix for https://github.com/PowerShell/vscode-powershell/issues/3929.


andschwa:
Ugh @SeeminglyScience regression in a test...

Also https://github.com/PowerShell/vscode-powershell/issues/873 will need this for the final fix.

SeeminglyScience:
Think I jumped to conclusions here, I can't for the life of me get what I thought it was to repro. It's gotta be something else.

PR #1764, opened 2022-04-19
Fix formatting handlers and PSScriptAnalyzer loading

andschwa:
Now we return `null` per the LSP specification if the formatting request was canceled (such as will happen if the document is modified since the request was sent), and we only load our bundled PSScriptAnalzyer module. Similar to PSReadLine, we should not support loading an arbitrary version of the module, as that leads to difficult to fix bugs and complicates the code.

Should fix https://github.com/PowerShell/vscode-powershell/issues/3928 and https://github.com/PowerShell/vscode-powershell/issues/3926.


SeeminglyScience (approved):
LGTM!

PR #1769, opened 2022-04-20
Don't print extra prompt on cancellation of `ReadLine`

andschwa:
We print an extra prompt as we execute with F5, and seemingly afterwards too. This is annoying, but fortunately cosmetic.


andschwa:
Closing in favor of #1778.

PR #1774, opened 2022-04-26
Implement temporary workaround for race condition in REPL

andschwa:
This fixes https://github.com/PowerShell/vscode-powershell/issues/3937. It is not a good fix, but it is viable for now. We've identified where the race condition is, and can effectively workaround it with a small sleep in the REPL. Our top priority is to fix up the REPL so this race doesn't occur in the first place, but that will require more invasive changes.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System;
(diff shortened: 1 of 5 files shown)


daxian-dbw (approved):
The temp workaround looks fine, but why are we doing this temp fix? To unblock someone in the short turn?

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -739,6 +739,11 @@ private void DoOneRepl(CancellationToken cancellationToken)
                 StopDebugContext();
             }
 
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return;
```
andschwa, line 744: LMAO it did need this?!

PR #1776, opened 2022-04-27
Don't cancel on disposal of `CancellationScope`

andschwa:
This eliminates an extra cancellation on tokens that are getting disposed.


SeeminglyScience (approved):
LGTM!

PR #1778, opened 2022-04-27
Fix extra prompting and manual debugger commands

andschwa:
This requires that we set the REPL to exit correctly. Note that we don't
have a notion of "disconnect" so we removed that logic. We are assuming
the debugger is "active" a little earlier now, that is, also when we've
received an LSP notification to start it (via a launch or attach
handler). Because we receive a notification on "disconnect" even for a
script just finishing, we need to know that the server is not active and
so not cancel the current prompt.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -757,18 +757,22 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
 
                 // If the user input was empty it's because:
                 //  - the user provided no input
-                //  - the readline task was canceled
-                //  - CtrlC was sent to readline (which does not propagate a cancellation)
+                //  - the ReadLine task was canceled
+                //  - CtrlC was sent to ReadLine (which does not propagate a cancellation)
                 //
-                // In any event there's nothing to run in PowerShell, so we just loop back to the prompt again.
-                // However, we must distinguish the last two scenarios, since PSRL will not print a new line in those cases.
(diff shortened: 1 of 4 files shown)


andschwa:
@SeeminglyScience Just added the commit that removes the `WriteWithPrompt` workaround from the `andschwa/improvements` branch (and originally explored in #1769). It's just cleaner now, the first commit is obviously the actual fix.

andschwa:
Kind of funny cosmetic bug: if you run a script, and then type input at the next prompt, and then run the script again, then PSReadLine doesn't save that input. But if you've invoked anything else in between (or type that text before running the script) it works fine. Strange, we'll fix it later.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -825,25 +831,16 @@ private string GetPrompt(CancellationToken cancellationToken)
...
-        }
-
         private string InvokeReadLine(CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
             try
             {
+                // TODO: If we can pass the cancellation token to ReadKey directly in PSReadLine, we
+                // can remove this logic.
+                _logger.LogDebug("Saving ReadKey cancellation token and invoking ReadLine");
```
SeeminglyScience, line 841:
```suggestion
```

Optional removal but seems like it was left in by mistake.

SeeminglyScience (changes requested):
LGTM but one thing. HMU if you want to keep it, doing request changes to not trigger auto merge

SeeminglyScience (approved):

PR #1780, opened 2022-04-28
Revert "Consolidate `InterruptCurrentForeground` and `MustRunInForegr…

SeeminglyScience:
Fixes this error:

image

This reverts (most of) commit 39c9ed4.


andschwa:
I'm curious, does this problem go away if we write a line after invoking readline if the task has been canceled?

See this part of the change (which was what allowed us to "re-use" the existing prompt for F5 and F8 input):
Screen Shot 2022-04-28 at 9 50 30 AM

PR #1781, opened 2022-04-28
Bring back `WriteWithPrompt()`

SeeminglyScience:
We were running into a situation where we'd print the prompt on the same line that we had previously printed it. This was especially noticable when stepping through the debugger.

This change ensures that we always write a new line if that situation occurs.

Examples:

image

image

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
@@ -62,7 +62,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
 
                 if (PowerShellExecutionOptions.WriteInputToHost)
                 {
-                    _psesHost.UI.WriteLine(_psCommand.GetInvocationText());
+                    _psesHost.WriteWithPrompt(_psCommand, cancellationToken);
                 }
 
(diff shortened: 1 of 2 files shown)


andschwa:
I'm curious, does this problem go away if we write a line after invoking readline if the task has been canceled?

See this part of the change (which was what allowed us to "re-use" the existing prompt for F5 and F8 input):
Screen Shot 2022-04-28 at 9 50 30 AM

andschwa:
Do we just want to bring `WriteWithPrompt` back? It's not the _worst_ thing in the world, it's just annoying to have a (probably empty) prompt, and then hit F5 and have a new prompt be printed with the script it's about to run.

SeeminglyScience:
> I'm curious, does this problem go away if we write a line after invoking readline if the task has been canceled?
> 
> See this part of the change (which was what allowed us to "re-use" the existing prompt for F5 and F8 input):

Yeah probably. I wonder why that works for you if you're not setting cursor left to zero somewhere...

That said I don't know if we want to reuse the prompt even if we have the option. That might need to be opt in otherwise it may confuse some folks. But we'd need to skip writing the prompt as well for that to work I'd think

andschwa:
That work as "expected" now?

SeeminglyScience:
> That work as "expected" now?

Yep it works. Would be interesting to see if we can revert this and find a way to not write the prompt when we don't need to as an option though. I like the idea

andschwa (approved):

PR #1782, opened 2022-04-28
Fix ordering of startup tasks so `psEditor` is defined before profiles are loaded

andschwa:
The thing is, `RequiresForeground` necessarily implies an execution
priority of `Next`. So instead we set that priority explicitly for
`psEditor` and mark none of the startup tasks as requiring foreground,
because they're all started before we enter the REPL where that matters.


andschwa:
Replaces https://github.com/PowerShell/PowerShellEditorServices/pull/1780.

SeeminglyScience (approved):
LGTM!

PR #1791, opened 2022-05-05
Send `sendKeyPress` event across DAP for temporary integrated consoles

andschwa:
This is kind of messy, but it fixes https://github.com/PowerShell/vscode-powershell/issues/3950.


SeeminglyScience (approved):
LGTM with a small naming suggestion

PR #1797, opened 2022-05-11
Increase stack size for PowerShell 5

nohwnd:
Sets the Thread max stack size to use the same value that is normally used in console host. This gives PowerShell code more room for execution, so recursive calls, or calls that are wrapped in Pester (which itself is stack-hungry), won't fail with: 

```
System.Management.Automation.ScriptCallDepthException: The script failed due to call depth overflow.
```

Fix https://github.com/PowerShell/vscode-powershell/issues/3962

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -112,7 +112,16 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
             _runspaceStack = new Stack<RunspaceFrame>();
             _cancellationContext = new CancellationContext();
 
-            _pipelineThread = new Thread(Run)
+            // Default stack size on .NET Framework is 524288 (512KB) (as reported by GetProcessDefaultStackSize)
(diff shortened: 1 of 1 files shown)


nohwnd:
One of the builds seems to fail due to flakiness.

nohwnd:
The stack size is also echoed here in open code https://github.com/PowerShell/PowerShell/blob/44098e771a059c06f1619eac8f97deeb118d127e/src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs#L23-L29 even though it probably has no effect on .NET.

andschwa (approved):
Wow, nice fix. Thank you!

markdowd:
Any idea when the new version will appear in PSGallery? Currently showing as Version 5.3.3, which has the problem.

andschwa:
Ah, sorry, we thought we had two PRs that would get in sooner than they did. It'll be out this week! Though we don't update the PSGallery, you'll find the PSES versions at https://github.com/PowerShell/PowerShellEditorServices/releases and the extension at https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell with the preview available at https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell-Preview.

PR #1799, opened 2022-05-12
Fix a lot of IntelliSense issues

SeeminglyScience:
## Changes:

(each section is it's own commit for easier review)

### Mark completion request handler as `Serial`

This stops completion from being cancelled whenever a `DidChangeTextDocument` notification is sent. So previously, if you typed `Get-` very quickly, you'd get no results at all. It also lets completion cancel some other expensive requests like code lens.

**Note:** the one *not fantastic* down side of this is it makes "missing" code lenses more likely. By missing I mean:

image

I have not yet figured out a way to tell VSCode to just try again next time it's able. Also we were already getting this sometimes, but this change does make it at least a bit more likely.

### Fix tasks never completing if cancelled quickly

We were exiting `SynchronousTask.ExecuteSynchronously` early if the cancellation was already requested. This was not setting our tcs state correctly, causing the caller to just fall off without warning. This made some things like `finally` and `using` disposals never happen.

### Make alias discovery possible on idle

The use of `Runspace.SessionStateProxy` while the pipeline thread is busy (even if you are **on** the pipeline thread) results in an exception.

This also allows us to cancel `GetAliasesAsync`. This was really the main reason I was in there, but I also remembered a report from @ninmonkey about an exception here so two birds one stone.

### Add space to completion trigger characters

**Note:** We should be on the look out for feedback about this change and let it sit in preview slightly longer than we'd normally like to. It's possible this may need to be opt in, but I think the entire intellisense experience is now optimized enough to that the behavior is ideal.

Adding space to completion triggers lets us automatically give completion results for parameter values. This also required building in some support for marking completion results as "Incomplete". This means that the client will not cache results when standard identifier characters are typed.

Mainly, this solves the scenario where you type space, get file completion results, and then type `-`. You expect to then get parameter names, but without marking "Incomplete", it will instead filter the file results.

Also fixed a scenario where you type `$script:` and on typing the `:` it would not update to available script scope variables.

### Make completion requests cancellable

That way any completion results from custom argument completers can be cancelled if they take too long and the user keeps typing. `TabExpansion2` is a callback-like function similar to `prompt`, partially for customizing tab completion but also for providing a command to get completion results.

In non-remote runspaces, we can also use it as a completion source that we can cancel by way of issuing a pipeline stop request.

### DIFF
--- a/src/PowerShellEditorServices/Server/PsesLanguageServer.cs
+++ b/src/PowerShellEditorServices/Server/PsesLanguageServer.cs
@@ -12,6 +12,7 @@ using Microsoft.PowerShell.EditorServices.Services;
 using Microsoft.PowerShell.EditorServices.Services.Extension;
 using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
 using Microsoft.PowerShell.EditorServices.Services.Template;
+using OmniSharp.Extensions.JsonRpc;
 using OmniSharp.Extensions.LanguageServer.Protocol.Server;
 using OmniSharp.Extensions.LanguageServer.Server;
 using Serilog;
@@ -102,7 +103,14 @@ namespace Microsoft.PowerShell.EditorServices.Server
                     .WithHandler<PsesCodeLensHandlers>()
                     .WithHandler<PsesCodeActionHandler>()
                     .WithHandler<InvokeExtensionCommandHandler>()
-                    .WithHandler<PsesCompletionHandler>()
+                    // If PsesCompletionHandler is not marked as serial, then DidChangeTextDocument
+                    // notifications will end up cancelling completion. So quickly typing `Get-`
+                    // would result in no completions.
+                    //
+                    // This also lets completion requests interrupt time consuming background tasks
+                    // like the references code lens.
+                    .WithHandler<PsesCompletionHandler>(
+                        new JsonRpcHandlerOptions() { RequestProcessType = RequestProcessType.Serial })
                     .WithHandler<PsesHoverHandler>()
                     .WithHandler<PsesSignatureHelpHandler>()
                     .WithHandler<PsesDefinitionHandler>()
--- a/src/PowerShellEditorServices/Services/CodeLens/ICodeLensProvider.cs
+++ b/src/PowerShellEditorServices/Services/CodeLens/ICodeLensProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerShell.EditorServices.Services.TextDocument;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
@@ -25,8 +26,9 @@ namespace Microsoft.PowerShell.EditorServices.CodeLenses
         /// <param name="scriptFile">
         /// The document for which CodeLenses should be provided.
         /// </param>
+        /// <param name="cancellationToken"></param>
         /// <returns>An array of CodeLenses.</returns>
-        CodeLens[] ProvideCodeLenses(ScriptFile scriptFile);
+        CodeLens[] ProvideCodeLenses(ScriptFile scriptFile, CancellationToken cancellationToken);
 
         /// <summary>
         /// Resolves a CodeLens that was created without a Command.
@@ -37,11 +39,13 @@ namespace Microsoft.PowerShell.EditorServices.CodeLenses
         /// <param name="scriptFile">
         /// The ScriptFile to resolve it in (sometimes unused).
         /// </param>
+        /// <param name="cancellationToken"></param>
         /// <returns>
         /// A Task which returns the resolved CodeLens when completed.
         /// </returns>
         Task<CodeLens> ResolveCodeLens(
             CodeLens codeLens,
-            ScriptFile scriptFile);
+            ScriptFile scriptFile,
+            CancellationToken cancellationToken);
     }
 }
--- a/src/PowerShellEditorServices/Services/CodeLens/ReferencesCodeLensProvider.cs
+++ b/src/PowerShellEditorServices/Services/CodeLens/ReferencesCodeLensProvider.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerShell.EditorServices.Services;
 using Microsoft.PowerShell.EditorServices.Services.Symbols;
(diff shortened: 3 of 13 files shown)


src/PowerShellEditorServices/Server/PsesLanguageServer.cs
```diff
@@ -102,7 +103,14 @@ public async Task StartAsync()
...
                     .WithHandler<InvokeExtensionCommandHandler>()
-                    .WithHandler<PsesCompletionHandler>()
+                    // If PsesCompletionHandler is not marked as serial, then DidChangeTextDocument
+                    // notifications will end up cancelling completion. So quickly typing `Get-`
+                    // would result in no completions.
+                    //
+                    // This also lets completion requests interrupt time consuming background tasks
+                    // like the references code lens.
+                    .WithHandler<PsesCompletionHandler>(
+                        new JsonRpcHandlerOptions() { RequestProcessType = RequestProcessType.Serial })
```
andschwa, line 113: 🥳

src/PowerShellEditorServices/Services/CodeLens/ICodeLensProvider.cs
```diff
@@ -25,8 +26,9 @@ internal interface ICodeLensProvider
         /// <param name="scriptFile">
         /// The document for which CodeLenses should be provided.
         /// </param>
+        /// <param name="cancellationToken"></param>
         /// <returns>An array of CodeLenses.</returns>
-        CodeLens[] ProvideCodeLenses(ScriptFile scriptFile);
+        CodeLens[] ProvideCodeLenses(ScriptFile scriptFile, CancellationToken cancellationToken);
```
andschwa, line 31: 🥳

src/PowerShellEditorServices/Services/CodeLens/ReferencesCodeLensProvider.cs
```diff
@@ -107,6 +115,10 @@ public async Task<CodeLens> ResolveCodeLens(CodeLens codeLens, ScriptFile script
                 List<Location> acc = new();
                 foreach (SymbolReference foundReference in referencesResult)
                 {
+                    // This async method is pretty dense with synchronous code
+                    // so it's helpful to add some yields.
+                    await Task.Yield();
```
andschwa, line 120: 👍

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
```diff
@@ -35,7 +35,7 @@ internal abstract class SynchronousTask<TResult> : ISynchronousTask
             CancellationToken cancellationToken)
         {
             Logger = logger;
-            _taskCompletionSource = new TaskCompletionSource<TResult>();
+            _taskCompletionSource = new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
```
andschwa, line 38: Can we comment exactly why we're doing this? (Like what did you hit that we have to set it.)

src/PowerShellEditorServices/Services/PowerShell/Utility/CommandHelpers.cs
```diff
@@ -180,19 +184,21 @@ internal static class CommandHelpers
...
-                    return invokeCommand.GetCommands("*", CommandTypes.Alias, nameIsPattern: true);
-                },
-                CancellationToken.None).ConfigureAwait(false);
+            // Need to execute a PSCommand here as Runspace.SessionStateProxy cannot be used from
+            // our PSRL on idle handler.
+            IReadOnlyList<CommandInfo> aliases = await executionService.ExecutePSCommandAsync<CommandInfo>(
+                new PSCommand()
+                    .AddCommand("Microsoft.PowerShell.Core\\Get-Command")
+                    .AddParameter("ListImported", true)
+                    .AddParameter("CommandType", CommandTypes.Alias),
```
andschwa, line 200: Ah, makes sense.

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -160,18 +161,25 @@ public List<SymbolReference> FindSymbolsInFile(ScriptFile scriptFile)
...
         {
             if (foundSymbol == null)
             {
                 return null;
             }
 
-            (Dictionary<string, List<string>> cmdletToAliases, Dictionary<string, string> aliasToCmdlets) = await CommandHelpers.GetAliasesAsync(_executionService).ConfigureAwait(false);
+            CommandHelpers.AliasMap aliases = await CommandHelpers.GetAliasesAsync(
+                _executionService,
+                cancellationToken).ConfigureAwait(false);
```
andschwa, line 179: 🥳 yay for things being cancellable!

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
```diff
@@ -92,6 +93,34 @@ static AstOperations()
                 (pwsh, _) =>
                 {
                     stopwatch.Start();
+
+                    // If the current runspace is not out of process, then we call TabExpansion2 so
+                    // that we have the ability to issue pipeline stop requests on cancellation.
+                    if (executionService is PsesInternalHost psesInternalHost
```
andschwa, line 99: Oof, I would happily remove the `IExecutionService` interface as it makes us do things like this.

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
```diff
@@ -92,6 +93,34 @@ static AstOperations()
                 (pwsh, _) =>
                 {
                     stopwatch.Start();
+
+                    // If the current runspace is not out of process, then we call TabExpansion2 so
+                    // that we have the ability to issue pipeline stop requests on cancellation.
```
andschwa, line 98: 🥳

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -131,21 +136,32 @@ public override async Task<CompletionItem> Handle(CompletionItem request, Cancel
...
             for (int i = 0; i < result.CompletionMatches.Count; i++)
             {
+                CompletionResult completionMatch = result.CompletionMatches[i];
+
+                // If a completion result is a variable scope like `$script:` we want to
+                // mark as incomplete so on typing `:` completion changes.
+                if (completionMatch.ResultType is CompletionResultType.Variable
+                    && completionMatch.CompletionText.EndsWith(":"))
+                {
+                    isIncomplete = true;
```
andschwa, line 159: Nice work 🥳

andschwa (approved):

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
SeeminglyScience: You can kind of think of it like `.ConfigureAwait(false)` for completion sources. I didn't necessarily need to add it here, it was just a troubleshooting step, *but* we probably should be adding it everywhere. [Related blog post](https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/)

I can add a comment here but maybe I should just update all the other instances as well? If we go that route would you still prefer a comment over each one?

src/PowerShellEditorServices/Services/Symbols/Vistors/AstOperations.cs
SeeminglyScience: Yeah either the sync task itself needs to reference `IExecutionService` or we need to strip it out. Or maybe there should be another interface that has state or w/e.

Maybe we hold off on tackling that until we get a source generator that handles the interface definitions for us? 🤷

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousTask.cs
andschwa: We don't need to do it in this PR, but thanks for the blog post. I'll make an issue for further investigation.

Issue #1695, opened 2022-02-02
Stdio mode broken since v3.0.0


rashil2000:
Hi!

[Kate](https://kate-editor.org) is an awesome open source text editor made by the KDE folks, and it supports language servers. By default, they provide configurations for some popular languages, documented here: https://docs.kde.org/trunk5/en/kate/kate/kate-application-plugin-lspclient.html

I am trying to add PowerShell support in it, by using this config:

```json
{
    "servers": {
        "powershell": {
            "command": ["pwsh", "-NoLogo", "-NoProfile", "-Command", "%{ENV:USERPROFILE}/Desktop/PowerShellEditorServices/PowerShellEditorServices/Start-EditorServices.ps1", "-BundledModulesPath", "%{ENV:USERPROFILE}/Desktop/PowerShellEditorServices", "-LogPath", "%{ENV:TEMP}/kate/logs.log", "-SessionDetailsPath", "%{ENV:TEMP}/kate/session.json", "-HostName", "Kate", "-HostProfileId", "KDE.Kate", "-HostVersion", "2.0.0", "-Stdio", "-LogLevel", "Normal"],
            "highlightingModeRegex": "^PowerShell$"
        }
    }
}
```

The server starts successfully, and recognizes symbols, but that's it. Nothing else seems to work, definitions go-to, hover etc. This is the log:

```
2022-02-02 12:11:37.835 +05:30 [INF] highlight handler loaded
2022-02-02 12:11:38.942 +05:30 [INF] PSScriptAnalyzer settings file not found. Falling back to default rules
2022-02-02 12:11:43.975 +05:30 [FTL] Failed to handle request textDocument/hover 3
System.InvalidOperationException: Stack empty.
   at System.Collections.Generic.Stack`1.ThrowForEmptyStack()
   at System.Collections.Generic.Stack`1.Peek()
   at Microsoft.PowerShell.EditorServices.Services.PowerShell.Host.PsesInternalHost.get_CurrentFrame() in D:\a\1\s\src\PowerShellEditorServices\Services\PowerShell\Host\PsesInternalHost.cs:line 153
   at Microsoft.PowerShell.EditorServices.Services.PowerShell.Host.PsesInternalHost.get_CurrentRunspace() in D:\a\1\s\src\PowerShellEditorServices\Services\PowerShell\Host\PsesInternalHost.cs:line 137
[...]

PR #1801, opened 2022-05-17
Re-enable stdio clients by fixing initialization sequence

andschwa:
In addition to a few clean-ups, the primary fix here is to start the PowerShell host in the `initialize` request handler, instead of the `onDidChangeConfiguration` handler, which only worked as it relied on a quirk of the VS Code client. In order to do this, we now expect to receive the settings `EnableProfileLoading: bool` and `InitialWorkingDirectory: string` as initialization parameters.

Fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1695.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
+++ b/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
@@ -208,6 +208,7 @@ namespace Microsoft.PowerShell.EditorServices.Hosting
 #endif
 
             // Add the bundled modules to the PSModulePath
+            // TODO: Why do we do this in addition to passing the bundled module path to the host?
             UpdatePSModulePath();
 
             // Check to see if the configuration we have is valid
--- a/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
+++ b/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
@@ -12,7 +12,7 @@ using SMA = System.Management.Automation;
 using System.Management.Automation;
 using System.Management.Automation.Runspaces;
 
-#if DEBUG
+#if ASSEMBLY_LOAD_STACKTRACE
 using System.Diagnostics;
 #endif
 
@@ -98,7 +98,7 @@ namespace Microsoft.PowerShell.EditorServices.Hosting
 
             AssemblyLoadContext.Default.Resolving += (AssemblyLoadContext _, AssemblyName asmName) =>
             {
-#if DEBUG
+#if ASSEMBLY_LOAD_STACKTRACE
                 logger.Log(PsesLogLevel.Diagnostic, $"Assembly resolve event fired for {asmName}. Stacktrace:\n{new StackTrace()}");
 #else
                 logger.Log(PsesLogLevel.Diagnostic, $"Assembly resolve event fired for {asmName}");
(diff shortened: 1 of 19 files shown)


src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
```diff
@@ -98,7 +98,7 @@ public sealed class EditorServicesLoader : IDisposable
 
             AssemblyLoadContext.Default.Resolving += (AssemblyLoadContext _, AssemblyName asmName) =>
             {
-#if DEBUG
+#if ASSEMBLY_LOAD_STACKTRACE
```
SeeminglyScience, line 101: Yesssssssssss

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
```diff
@@ -184,7 +185,10 @@ public override async Task<CompletionItem> Handle(CompletionItem request, Cancel
                 // Retain PowerShell's sort order with the given index.
                 SortText = $"{sortIndex:D4}{result.ListItemText}",
                 FilterText = result.CompletionText,
-                TextEdit = textEdit // Used instead of InsertText.
+                // Used instead of Label when TextEdit is unsupported
+                InsertText = result.CompletionText,
```
SeeminglyScience, line 189: Ah so it just falls back then? That's handy!

How does `InsertText` work with things like type names where the completion text can be the whole type name? e.g. `[System.Management.Automation.<tab>`. Though even if that explodes it's better than not working at all so I'm cool with rollin' with it.

SeeminglyScience (approved):
LGTM! ❤️ 

One non-blocking question

src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
andschwa: Was _really_ bugging me, but didn't want to throw this away either.

src/PowerShellEditorServices/Services/TextDocument/Handlers/CompletionHandler.cs
SeeminglyScience: > but also as I was tabbing through SMA it erroneously at some point doubled up the SM part

That's what I was expecting. My PR won't fix that, I'm even not 100% sure we *can* fix it if the client doesn't support edits. Either way I'm cool with it just to get those clients at least working and we can take a pass at figuring something out later

dkattan:
YES, I can finally remove my hacky onDidChnageConfiguration workaround to get PSES to initialize!

andschwa:
@dkattan I saw that and just _cringed_. Don't ask me how it happened!

dkattan:
> @dkattan I saw that and just _cringed_. Don't ask me how it happened!

I'm pretty sure it's been like this since before the re-write. I first encountered the issue on 2021-06-03 and I stared at https://github.com/TypeFox/monaco-languageclient/issues/201 forever trying to figure out how to make it work.

This was what I had to do:
```javascript
const monacoLanguageClient = new MonacoLanguageClient({
            name: "Monaco PowerShell",
            clientOptions: {
              // use a language id as a document selector
              documentSelector: ["powershell"],
              middleware: {
                workspace: {
                  didChangeConfiguration: sections => {
                    console.debug(
                      this.$options.name,
                      "Language Server: Overriding workspace/didChangeConfiguration response; Sections: ",
                      sections
                    );
                    monacoLanguageClient.sendNotification(
                      "workspace/didChangeConfiguration",
                      sections
                    );
                  },
                },
              },
```

PR #1802, opened 2022-05-17
Set `EnableProfileLoading` default to `true`

andschwa:
So it can be turned off if needed, but otherwise has a sane default.


andschwa:
This seems to break in the end-to-end tests. Grr.

PR #1803, opened 2022-05-17
Fix execution of debug prompt commands

andschwa:
Thanks to a brilliant idea from Patrick, this simplifies how we choose which commands to execute through PowerShell's debugger API, and also fixes an issue where a user's custom prompt function wasn't aware that it was in the debugger.

This fixes https://github.com/PowerShell/vscode-powershell/issues/3980.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
@@ -48,10 +48,6 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
 
         public override ExecutionOptions ExecutionOptions => PowerShellExecutionOptions;
 
-        // These are PowerShell's intrinsic debugger commands that must be run via
-        // `ProcessDebugCommand`.
-        private static readonly string[] DebuggerCommands = { "continue", "c", "k", "h", "?", "list", "l", "stepInto", "s", "stepOut", "o", "stepOver", "v", "quit", "q", "detach", "d" };
-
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
```diff
@@ -65,8 +61,14 @@ public override IReadOnlyList<TResult> Run(CancellationToken cancellationToken)
...
 
+                // If we're in a breakpoint it means we're executing either interactive commands in
+                // a debug prompt, or our own special commands to query the PowerShell debugger for
+                // state that we sync with the LSP debugger. The former commands we want to send
+                // through PowerShell's `Debugger.ProcessCommand` so that they work as expected, but
+                // the latter we must not send through it else they pollute the history as this
+                // PowerShell API does not let us exclude them from it.
                 return _pwsh.Runspace.Debugger.InBreakpoint
-                    && (IsDebuggerCommand(_psCommand) || _pwsh.Runspace.RunspaceIsRemote)
+                    && (PowerShellExecutionOptions.AddToHistory || IsPromptCommand(_psCommand))
```
SeeminglyScience, line 71:
```suggestion
                    && (PowerShellExecutionOptions.AddToHistory || IsPromptCommand(_psCommand) || _pwsh.Runspace.RunspaceIsRemote)
```

All remote invocations still need to go through the debugger when it's stopped since we can't be on the out of process thread.

SeeminglyScience (changes requested):
One thing needs to change real quick otherwise LGTM

andschwa:
Good catch! Committed.

SeeminglyScience (approved):
LGTM!

PR #1804, opened 2022-05-17
Wrap untitled script with newlines

andschwa:
Fixes https://github.com/PowerShell/vscode-powershell/issues/3965


andschwa:
@SeeminglyScience tested on Windows. Can repro original issue, and then with fix no longer repros 🥳

SeeminglyScience (approved):
LGTM!

PR #1807, opened 2022-05-19
Fix startup bug when zero profiles are present

andschwa:
The service would fail to fully start if zero profiles were present, because that meant we attempted to invoke an empty command, which then threw `PSInvalidOperationException: No commands are specified`. This wasn't caught by our task executor, since it happened in a delegate which directly invoked the command.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
@@ -2,13 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Reflection;
 using System.Text;
 using Microsoft.Extensions.Logging;
 using Microsoft.PowerShell.EditorServices.Hosting;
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
```diff
@@ -211,7 +211,13 @@ public static void LoadProfiles(this PowerShell pwsh, ProfilePathInfo profilePat
             // NOTE: This must be set before the profiles are loaded.
             pwsh.Runspace.SessionStateProxy.SetVariable("PROFILE", profileVariable);
 
-            pwsh.InvokeCommand(psCommand);
+            // NOTE: Because it's possible there are no profiles defined, we might have an empty
+            // command. Since this is being executed directly, we can't rely on `ThrowOnError =
+            // false` to avoid an exception here. Instead, we must just not execute it.
+            if (psCommand.Commands.Count > 0)
+            {
+                pwsh.InvokeCommand(psCommand);
```
andschwa, line 219: TODO: @SeeminglyScience we should check everywhere else we directly invoke too.

SeeminglyScience (approved):
LGTM!

PR #1818, opened 2022-06-07
Exit debugger stop early if cause is PSE

SeeminglyScience:
If the `ErrorActionPreference` is set to `Break` and the user hits the
Stop button in the UI, then we need to exit early on reentrance of
`OnDebuggerStop`. Currently we deadlock until they hit stop again.

Fixes PowerShell/vscode-powershell#4010


andschwa (approved):
Approving, leaving for you to squash and merge so you can double check if it closes that issue. Thanks Patrick!!!

SeeminglyScience:
Yep! Thanks Andy!

PR #1820, opened 2022-06-09
Add Emacs regression tests for PSES

andschwa:
This adds regression tests written in Emacs to validate that PSES can be connected (and reports diagnostics) with a client that isn't VS Code, namely [Eglot](https://github.com/joaotavora/eglot), and Emacs LSP client.

### DIFF
--- a/.github/workflows/emacs-test.yml
+++ b/.github/workflows/emacs-test.yml
@@ -1,0 +1,30 @@
+name: "Emacs"
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


andschwa:
Should prevent https://github.com/PowerShell/PowerShellEditorServices/issues/1695 from recurring.

SeeminglyScience (approved):
Amazing, LGTM!

PR #1830, opened 2022-06-15
Add regression test for untitled scripts in Windows PowerShell

andschwa:
Specifically targeting the edge case where the last line of the script is a comment. Covers https://github.com/PowerShell/vscode-powershell/issues/3965

### DIFF
--- a/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
+++ b/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
@@ -152,8 +152,12 @@ namespace Microsoft.PowerShell.EditorServices.Services
         /// </summary>
         /// <param name="filePath">The file path at which the script resides.</param>
         /// <param name="scriptFile">The out parameter that will contain the ScriptFile object.</param>
-        public bool TryGetFile(string filePath, out ScriptFile scriptFile) =>
-            TryGetFile(new Uri(filePath), out scriptFile);
+        public bool TryGetFile(string filePath, out ScriptFile scriptFile)
+        {
+            scriptFile = null;
+            return Uri.IsWellFormedUriString(filePath, UriKind.RelativeOrAbsolute)
+                && TryGetFile(new Uri(filePath), out scriptFile);
+        }
 
         /// <summary>
         /// Tries to get an open file in the workspace. Returns true if it succeeds, false otherwise.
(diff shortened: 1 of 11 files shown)


SeeminglyScience (approved):
LGTM! gj figuring this one out 😁 did not sound fun!

andschwa:
@SeeminglyScience could you please re-review?

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -152,8 +152,19 @@ public ScriptFile GetFile(DocumentUri documentUri)
...
+            try
+            {
+                return TryGetFile(new Uri(filePath), out scriptFile);
+            }
+            catch (UriFormatException)
+            {
+                scriptFile = null;
+                return false;
+            }
+        }
```
andschwa, line 167: I had to change this one to catch the exception instead of using `Uri.IsWellFormedUriString(path, UriKind.RelativeOrAbsolute)` since the path we were being given was like `\\foo\\bar`, which makes me hesitant about the above too.

test/PowerShellEditorServices.Test/Session/ScriptFileTests.cs
```diff
@@ -651,6 +651,7 @@ public void DocumentUriReturnsCorrectStringForAbsolutePath()
         [InlineData("vscode-notebook-cell:/Users/me/Documents/test.ps1#0001", true)]
         [InlineData("https://microsoft.com", true)]
         [InlineData("Untitled:Untitled-1", true)]
+        [InlineData("'a log statement' > 'c:\\Users\\me\\Documents\\test.txt'\r\n", false)]
         public void IsUntitledFileIsCorrect(string path, bool expected) => Assert.Equal(expected, ScriptFile.IsUntitledPath(path));
```
andschwa, line 655: See, this test implies that the function is asking "is this path a URI that indicates we're untitled?"

andschwa: Perhaps we should expect this case to be `true` by flipping the logic with `IsWellFormedUri`. That is, if we're given something that is _not_ a URI, assume it's untitled, but I'm not sure that would work for the first two cases.

andschwa: Yeah, doesn't work, since `/Users/me/Documents/test.ps1` is not a URI.

PR #1832, opened 2022-06-16
Improve `$PROFILE` variable and profile loading test

andschwa:
Now we actually check that it is well-formed.


SeeminglyScience (approved):
LGTM!

PR #1834, opened 2022-06-17
Update processor architecture check in build script

andschwa:
Windows 11 updated and got smarter, this can now also be ARM64 (which is
more precise).


SeeminglyScience (approved):
LGTM!

PR #1836, opened 2022-06-21
Fix piping to native commands for Windows PowerShell

andschwa:
For an unknown reason, we were explicitly overriding `Console.InputEncoding`, which led to a bug where input piped to a
native command in Windows PowerShell (where the default encoding is different than our override) would fail. Instead, we just don't override the encodings at all, which lets PSES be closer to the expected experience.

The regression test covers this, though to test the failure I had to move that override outside its guard (as it wasn't being set during tests).

Also fix building on ARM64 in Windows PowerShell, which doesn't have `$IsWindows`.

Thanks @SeeminglyScience for so quickly figuring out the problem.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4021


SeeminglyScience (approved):
LGTM!

PR #1839, opened 2022-06-22
Fix logic checking for untitled or raw scripts

andschwa:
This was much more easily accomplished by checking if what we were given
exists as a file path, in which case we use it as-is (a file path, that
must exist, since we launch it directly). Otherwise we can assume that
it's a raw script block, or a URI from the client that probably
represents an untitled file.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
@@ -105,9 +105,13 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
         private async Task LaunchScriptAsync(string scriptToLaunch)
         {
             PSCommand command;
-            // Script could an actual script, or a URI to a script file (or untitled document).
-            if (!System.Uri.IsWellFormedUriString(scriptToLaunch, System.UriKind.RelativeOrAbsolute)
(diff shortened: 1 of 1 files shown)


andschwa:
Fixes a bug introduced in https://github.com/PowerShell/PowerShellEditorServices/pull/1830

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ConfigurationDoneHandler.cs
```diff
@@ -105,9 +105,13 @@ public Task<ConfigurationDoneResponse> Handle(ConfigurationDoneArguments request
         private async Task LaunchScriptAsync(string scriptToLaunch)
         {
             PSCommand command;
-            // Script could an actual script, or a URI to a script file (or untitled document).
-            if (!System.Uri.IsWellFormedUriString(scriptToLaunch, System.UriKind.RelativeOrAbsolute)
-                || ScriptFile.IsUntitledPath(scriptToLaunch))
+            if (System.IO.File.Exists(scriptToLaunch))
```
SeeminglyScience, line 108: I think this is going to break for remote files (e.g. you do an interactive `Enter-PSSession` and then `psedit ./something.ps1`) but that's also probably super broken atm anyway. Something we should revisit in the future, but definitely not a blocker to getting this fixed

SeeminglyScience (approved):
LGTM!

PR #1823, opened 2022-06-10
Utilize the `AddToHistory` delegate from PSRL proxy

SeeminglyScience:
Fixes PowerShell/vscode-powershell#3683

This PR just fixes the regression of history not including <kbd>F5</kbd>/<kbd>F8</kbd>. Opened #1822 to track making this configurable

PR #1841, opened 2022-06-23
Add command to PSReadLine history before cancellation

andschwa:
Otherwise PSReadLine thinks it's supposed to re-insert it in the buffer after execution.

This is an extension of https://github.com/PowerShell/PowerShellEditorServices/pull/1823 which fixes https://github.com/PowerShell/vscode-powershell/issues/3683, but fixes https://github.com/PowerShell/vscode-powershell/issues/4041 too.


andschwa:
TODO: Somehow add a regression test!

andschwa:
@SeeminglyScience test _thoroughly_ please! No rush, we can get this in tomorrow and cut a new preview.

SeeminglyScience:
Testing looks good!

Only one thing to note, if anyone has a custom PSRL `AddToHistoryHandler` it's probably going to deadlock since we're on a different thread. This isn't really new and it's what happened before Rob's rewrite, so I think that's fine, just something to keep in mind.

SeeminglyScience (approved):
LGTM!

PR #1850, opened 2022-07-05
Catch `OperationCanceledException` in both command loops

andschwa:
Our flaky extension command test seems to be flaky because sometimes another task gets queued, and since it runs in the foreground it cancels that task. Interactively, this happens in the first loop (with `DoOneRepl`) which catches the cancellation exception; but when under tests that is a no-op, so it happens in the second loop. Hence we should duplicate the same logic and so catch that exception.


SeeminglyScience (approved):
LGTM!

Built locally and tried a bunch of debug scenarios, no issues!

PR #1851, opened 2022-07-05
Add `Directory.Exists()` check to `SetInitialWorkingDirectoryAsync()`

### LINKED ISSUE #1849: Extension won't start with invalid (empty string) `powershell.cwd`
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Per https://github.com/PowerShell/vscode-powershell/issues/4059#issuecomment-1172690701, setting and then unsetting the `cwd` setting will result in an empty string, which the client is fine with, but causes the server to fail to start.

### Expected behavior

```console
Server starts with any setting for `powershell.cwd`.
```

### Actual behavior

```console
Server hangs.
```

### Error details

_No response_

### Environment data

```PowerShell
N/A
```

### Version

v2022.6.3

### Visuals

_No response_

andschwa:
There was an edge case where, after setting and un-setting the `powershell.cwd` property in the VS Code client, instead of being `null` the setting would be the empty string `""`. The client did not care about this, and so left the server to silently crash on startup. We previously assumed that by the time the server started, the directory would be validated by the client, but this was wrong. So now we check its existence first.

Resolves #1849 (with a regression test too).

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -475,9 +475,11 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
 
         public Task SetInitialWorkingDirectoryAsync(string path, CancellationToken cancellationToken)
         {
-            return ExecutePSCommandAsync(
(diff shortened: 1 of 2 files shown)


andschwa:
Also working on improving the client-side of this too.

SeeminglyScience (approved):
LGTM!

PR #1853, opened 2022-07-08
Sync breakpoints outside of debug sessions

SeeminglyScience:
# PR Summary

This change is paired with PowerShell/vscode-powershell#4065

Adds a client utilizing a custom notification to sync breakpoints between the client and the server at all times. All breakpoint additions, removals, and enable/disable will be synced regardless of if they occur within the console (e.g. `sbp -Variable someVar -Mode Write`) or in the UI. The only exception is when a breakpoint is enabled/disabled in the PSIC (there's no way to directly enable or disable a breakpoint in the vsc API afaict)

Setting as WIP as I still need to figure out a way to add tests, but please review carefully anyway @andschwa. Ideally we'd release a stable before merging this as I'd like a slightly longer preview period for this change if possible.


src/PowerShellEditorServices/Index.cs
```diff
@@ -0,0 +1,154 @@
+// <auto-generated>
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
SeeminglyScience, line 3: I don't 100% know what to do with this and `Index`. I may have to rip these out and stop using the range syntax for arbitrary legal reasons but I'd like to check out what other MS projects are doing to polyfill.

src/PowerShellEditorServices/Server/PsesLanguageServer.cs
```diff
@@ -151,9 +152,12 @@ public async Task StartAsync()
                                 LoadProfiles = initializationOptions?.GetValue("enableProfileLoading")?.Value<bool>() ?? true,
                                 // TODO: Consider deprecating the setting which sets this and
                                 // instead use WorkspacePath exclusively.
-                                InitialWorkingDirectory = initializationOptions?.GetValue("initialWorkingDirectory")?.Value<string>() ?? workspaceService.WorkspacePath
+                                InitialWorkingDirectory = initializationOptions?.GetValue("initialWorkingDirectory")?.Value<string>() ?? workspaceService.WorkspacePath,
+
+                                SupportsBreakpointSync = initializationOptions?.GetValue("supportsBreakpointSync")?.Value<bool>() ?? false,
```
andschwa, line 157: Can we do this in the DAP, and using a custom capability instead?

src/PowerShellEditorServices/Server/PsesLanguageServer.cs
```diff
@@ -151,9 +152,12 @@ public async Task StartAsync()
...
                             };
 
+                            languageServer.Services.GetService<BreakpointSyncService>().IsSupported = hostStartOptions.SupportsBreakpointSync;
```
andschwa, line 160: Ditto.

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+            }
+
+            existing = CreateFromServerBreakpoint(eventArgs.Breakpoint);
+            string? id = await SendToClientAsync(existing.Client, BreakpointUpdateType.Set)
+                .ConfigureAwait(false);
+
+            existing.Client.Id = id!;
+            RegisterBreakpoint(existing);
+            return;
+        }
```
andschwa, line 159: Just a little confusing to have this in an `if (!getThing(named out param))` as the out param is _not_ used (since it's not, it wasn't in the map) but we then assign to it locally and use that. I'd do `out Thing _` and then locally declare `Thing breakpoint`. Or did I read this wrong?

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+
+        if (eventArgs.UpdateType is BreakpointUpdateType.Removed)
+        {
+            UnregisterBreakpoint(existing);
+            await SendToClientAsync(existing.Client, BreakpointUpdateType.Removed).ConfigureAwait(false);
+            return;
+        }
```
andschwa, line 166: Wait does the `out` param in the `if` statement above _really_ declare it in the scope above the conditional?! That's confusing scope. Maybe my confusion would be better alleviated by doing the `TryGet` first, and then check the return result only in the `if`, so that the `out` param looks like it's declared in the scope it is actually declared in.

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+            };
+
+            UnregisterBreakpoint(existing);
+            RegisterBreakpoint(newBreakpoint);
+            return;
+        }
+
+        LogBreakpointError(
+            existing,
+            "Somehow we're syncing a new breakpoint that we've already sync'd. That's not supposed to happen.");
```
andschwa, line 187: You _know_ we're gonna see this in a user's log one day and go 😳

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+
+        List<(ClientBreakpoint Client, BreakpointTranslationInfo? Translation)>? toAdd = null;
+        List<SyncedBreakpoint>? toRemove = null;
+        List<(SyncedBreakpoint Breakpoint, bool Enabled)>? toSetActive = null;
+
+        foreach (ClientBreakpoint clientBreakpoint in clientBreakpoints)
+        {
+            if (!_map.ByGuid.TryGetValue(clientBreakpoint.Id, out SyncedBreakpoint existing))
+            {
+                (toAdd ??= new()).Add((clientBreakpoint, GetTranslationInfo(clientBreakpoint)));
```
andschwa, line 299: Love that operator 😋

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+            functionName = $"${vbp.Variable}!{mode}";
+        }
+        else if (serverBreakpoint is CommandBreakpoint cbp)
+        {
+            functionName = cbp.Command;
+        }
+        else if (serverBreakpoint is LineBreakpoint lbp)
+        {
+            location = new ClientLocation(
+                // TODO: fix the translation of this
```
andschwa, line 524: What's it cause right now to have this TODO?

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
```diff
@@ -0,0 +1,904 @@
...
+            if (nameWithSuffix is { Length: > 2 }
+                && nameWithSuffix[^1] is 'w' or 'W'
+                && nameWithSuffix[^2] is '!')
+            {
+                name = nameWithSuffix[0..^2];
+                return VariableAccessMode.Write;
+            }
+
+            name = nameWithSuffix;
+            return VariableAccessMode.ReadWrite;
```
andschwa, line 754: 🤯

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -237,6 +241,12 @@ public async Task<AttachResponse> Handle(PsesAttachRequestArguments request, Can
 
         private async Task<AttachResponse> HandleImpl(PsesAttachRequestArguments request, CancellationToken cancellationToken)
         {
+            cancellationToken.Register(() =>
+                {
+                    if (Runspace.DefaultRunspace != null)
+                    {
+                    }
+                });
```
andschwa, line 249: What's this do?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -515,6 +533,7 @@ private async Task OnExecutionCompletedAsync(Task executeTask)
 
             _debugService.IsClientAttached = false;
             _debugAdapterServer.SendNotification(EventNames.Terminated);
+            _debugStateService.Reset();
```
andschwa, line 536: I like the reset LOL.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -452,6 +458,18 @@ public IReadOnlyList<TResult> InvokePSCommand<TResult>(PSCommand psCommand, Powe
             return task.ExecuteAndGetResult(cancellationToken);
         }
 
+        void IInternalPowerShellExecutionService.UnsafeInvokePSCommand(
```
andschwa, line 461: **UNSAFE**

andyleejordan (commented):
Ok I got through _some_ of this!

src/PowerShellEditorServices/Server/PsesLanguageServer.cs
SeeminglyScience: afaict DAP only exists when a debug session is running :/

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
SeeminglyScience: Yeah it's used outside of that `if` block. The if block is the exit early condition for it not already existing.

I can for sure see how that's confusing though I'll think of a better way to write this

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
SeeminglyScience: > You _know_ we're gonna see this in a user's log one day and go 😳

oh yeah 💯 

But we're gonna have a whole lot of info when it happens! (that method writes out a bunch of info about the bp)

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
SeeminglyScience: > **UNSAFE**

Yeah it's a convention used in a few places in the BCL to just mean "make sure you know what you're doing".

Mainly I don't like that we have `ExecuteCommandAsync` which marshals to the right thread... and then we have `InvokeCommand` which assumes you're on the right thread and potentially dangerously just invokes under that assumption. 

It would be very easy to mess up our threading house of cards by forgetting this and assuming `InvokeCommand` is just "the sync version that does the same thing".

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
andschwa: I saw! It's wonderful.

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
SeeminglyScience: Alright I rewrote this to call a nested method and added a new variable instead of reusing `existing` since that was really confusing. LMK what you think!

src/PowerShellEditorServices/Services/DebugAdapter/BreakpointSyncService.cs
SeeminglyScience: No idea! Hopefully I already fixed it 🙃, removed

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
SeeminglyScience: Removed! That was there as a debugging thing and I was adding `Debug.WriteLine` or something as a test.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/BreakpointHandlers.cs
```diff
@@ -32,23 +32,46 @@ internal class BreakpointHandlers : ISetFunctionBreakpointsHandler, ISetBreakpoi
...
+            _breakpointSyncService = breakpointSyncService;
         }
 
         public async Task<SetBreakpointsResponse> Handle(SetBreakpointsArguments request, CancellationToken cancellationToken)
         {
+            if (_breakpointSyncService.IsSupported)
+            {
+                // TODO: Find some way to actually verify these. Unfortunately the sync event comes
+                // through *after* the `SetBreakpoints` event so we can't just look at what synced
+                // breakpoints were successfully set.
```
andyleejordan, line 59: Annoying!

src/PowerShellEditorServices/Services/PowerShell/IPowerShellExecutionService.cs
```diff
@@ -51,8 +51,22 @@ public interface IPowerShellExecutionService
...
+        /// has ensured that they are already running on the pipeline thread.
+        /// </summary>
+        void UnsafeInvokePSCommand(PSCommand psCommand, PowerShellExecutionOptions executionOptions, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Create and execute a <see cref="SynchronousPowerShellTask{TResult}" /> without queuing
+        /// the work for the pipeline thread. This method must only be invoked when the caller
+        /// has ensured that they are already running on the pipeline thread.
+        /// </summary>
+        IReadOnlyList<TResult> UnsafeInvokePSCommand<TResult>(PSCommand psCommand, PowerShellExecutionOptions executionOptions, CancellationToken cancellationToken);
```
andyleejordan, line 70: I wonder if we should actually use this to process user input from the prompt when we're already on the thread receiving said input...

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -527,6 +533,18 @@ public IReadOnlyList<TResult> InvokePSCommand<TResult>(PSCommand psCommand, Powe
...
+            CancellationToken cancellationToken)
+            => InvokePSCommand<PSObject>(psCommand, executionOptions, cancellationToken);
+
+        IReadOnlyList<TResult> IInternalPowerShellExecutionService.UnsafeInvokePSCommand<TResult>(
+            PSCommand psCommand,
+            PowerShellExecutionOptions executionOptions,
+            CancellationToken cancellationToken)
+            => InvokePSCommand<TResult>(psCommand, executionOptions, cancellationToken);
+
         public void InvokePSCommand(PSCommand psCommand, PowerShellExecutionOptions executionOptions, CancellationToken cancellationToken) => InvokePSCommand<PSObject>(psCommand, executionOptions, cancellationToken);
```
andyleejordan, line 548: Perhaps we should make this private and _require_ the `Unsafe` versions be called?

andyleejordan (commented):
Reviewed! Thinking we should do a team review too. But this is very exciting, let's get it into the next preview!

JustinGrote:
Hey @SeeminglyScience and @andyleejordan I can probably go thru and clean up conflicts to get this up to date, but I was wondering what is the impetus? Is there a limitation in how DAP handles breakpoints that we can't do this inside DAP? What's the ramifications for other clients e.g. neovim?

SeeminglyScience:
> Hey @SeeminglyScience and @andyleejordan I can probably go thru and clean up conflicts to get this up to date

Ty for the offer but I just went through and cherry picked to a new branch (though that may not have been super necessary 🤷) 

> Is there a limitation in how DAP handles breakpoints that we can't do this inside DAP?

DAP is only active while debugging. Which makes perfect sense for most languages, but we're *kinda* always debugging (which would be annoying UX is we actually had an active debug session all the time)

> What's the ramifications for other clients e.g. neovim?

The aim is no different than before this change.

JustinGrote:
> | What's the ramifications for other clients e.g. neovim?
> The aim is no different than before this change.

I guess more specifically, are we going to need to include implementation instructions for other language clients since this is not being handled exclusively in the DAP spec and using custom messages instead?

> DAP is only active while debugging. Which makes perfect sense for most languages, but we're kinda always debugging (which would be annoying UX is we actually had an active debug session all the time)

This doesn't have to be the case as far as I can tell, DAP can be active and persist between debugging sessions. Rather than go through a full initialization and setup on every debug launch, we can do a single initialization, and keep breakpoints in sync between launch/attach reqeusts.

Per the spec under [launch sequencing](https://microsoft.github.io/debug-adapter-protocol/overview), we currently send all our breakpoint syncs *after* a launch, but we should instead be persistently keeping it up to date *in between* launches
image

Here's the view from my DAP trace of what we currently do:
image

When what we *can* do instead is:
1. Initialize
2. Set (and continue to set) breakpoints, syncing as they are changed in vscode and changed via set-breakpoint, etc.
3. When a launch happens, we could reverify just-prior (though if we do #2 right we shouldn't have to)
4. launch/attaches continue to occur on the single persistent DAP session (might need different ones for each PSES temporary runspace/etc. but the same logic still applies)

If we do it this way then no special messaging has to be implemented in clients, the clients just have to maintain the DAP connection rather than shut it down between debug sessions.

You've been working on this a lot longer than I have so maybe there's a limitation you discovered as to why we can't do it this way, I invite your feedback.

SeeminglyScience:
Is there a specific part of the doc you linked that mentions multiple launches/terminates per initialize? Or are you inferring that from the name?

Unless it's changed, the behavior (and the way I'm reading that doc) was roughly "when a debug session is requested, send init, config, setbreakpoints/etc, launch, then wait for terminate". No messages in between debug sessions.

That's why we had to wait for some event we could hook into to even take a crack at rolling our own. Though again, maybe it's changed since then

JustinGrote:
I would say it is somewhat inferred, at the very end

JustinGrote:
I would say it is somewhat inferred at the very end under "Debug Session End"
image

The end of a debug session is signified via a `disconnect` from the client or an `exited` from the debugee. Therefore the debugee is "terminated" (in our case we only fully exit the process if it was a launch to a separate terminal or remoting, etc.) and we signify that to the client that it has, but there's nothing that says the client can't initiate a new launch at that point. It also doesn't show the debug adapter terminating in addition to the debugee.

All language of terminate and disconnect refer to the debugee, NOT the debug adapter.

Based on my reading of this a debug session is defined as a superset, and multiple launch/attach/disconnect/terminate pairings can existing within the context of a debug session.

JustinGrote:
Definition of debug session still appears to be vague https://github.com/microsoft/debug-adapter-protocol/issues/386 but `vscode-js-debug` [reuses the adapter for multiple sessions in serial](https://github.com/microsoft/vscode-js-debug/issues/1388#issuecomment-1483151502) so it is de-facto supported I would say.

Multiplexing (simultaneous session in the same adapter) however is not supported, discussion here: https://github.com/microsoft/debug-adapter-protocol/issues/329, so we would need to maintain 1 debug adapter per debugee (PSIC, standalone terminal, remoting) or only support 1 PS debug at a time, so we'd probably need some sort of event hub to register for the breakpoint syncs so all can be updated if we went this route.

JustinGrote:
Also, the `setBreakpoint` messages are still only client -> server, there's no way for the server to inform the client of a breakpoint change, only that a debug has been server-initiated with `StartDebugging`, so we would still need a custom S->C message in the case of `Set-Breakpoint` regardless.

Because we need custom messages anyways in that case then, I'd say there's no major need to change this, but maybe we can reduce the amount of custom implementation for clients in the future if some of the DAP protocol becomes more clear in this regard.

SeeminglyScience:
> Sorry for any potential thought derailment :)

Oh yeah no worries. Rest assured I would have *much* rather just hooked up existing messages. Trying to roll your own is a huge PITA and more complicated than one would think. Skipping that would have been tops

JustinGrote:
So a slight clarification, I misstated when I said there was no way for the server to inform the client of a breakpoint change, there is, [the `breakpoint` event.](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Breakpoint)

However as @SeeminglyScience stated, this only works "in" a debug session, and upon further reviewing the spec, I feel it's pretty clear that the end of a debug session means when the terminate/disconnect/exit happens, and another part states the debug adapter should terminate/disconnect, so this implies that there should only ever be one launch/attach request per debug session (aka adapter lifetime)

So it still stands that, unless we can have some sort of "background" attach session that doesn't surface in UIs and exchange messages that way, custom messages will be the way to go.

JustinGrote:
So I had a crazy idea, what about using a [custom configuration section](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) to manage breakpoints as a configuration?

There's nothing that says this has to be surfaced in vscode settings, and the type can be LSPAny which can include the breakpoint type. Client-Server is notified via Configuration Request, and Server-Client is notified via OnDidChangeConfiguration.

It would still require custom client configuration to hook the breakpoint set to the config, but it wouldn't require a custom message.

JustinGrote:
Reviewing this again, I think we might still be able to get away without custom messaging that I may explore.

1. A launch/attach request starts and `initialize/initialized` is completed
2. client sends its configured breakpoints via `setBreakpoint`
3. PSES sets the breakpoints in the process, overwriting any that already exist, and responds via `setBreakpoint` reply
4. PSES then queries the current breakpoints and finds any "additional" ones that weren't set via step 3, and spuriously reports them to the client via [Breakpoint](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Breakpoint) event as `new`
5. PSES does not send `configurationDone` until the breakpoint event above is sent. At this point the client and PSES should be fully synced, *assuming client can process breakpoint events before configurationDone which is not explicitly clear in DAP spec*
6. [Optional, probably separate PR] PSES monitors for breakpoint changes and sends new/changed/delete Breakpoint events when they occur, syncing up the client to PSES

A custom launch option for `alwaysClearBreakpoints` or whatever can be defined to override this behavior and reset whatever state is found.

This would make both launch and attach events be able to utilize existing breakpoints, and keep things in sync.

As a potential dirty workaround if spurious breakpoint events don't work during config stage, is for the `setBreakpoint` reply to include the additional extraneous breakpoints found by PSES already in the process. Less clean but no reason it shouldn't work.

@SeeminglyScience @andyleejordan feel free to poke holes in my thinking here.

PR #1860, opened 2022-07-22
Rename "Integrated Console" to "Extension Terminal"

andschwa:
Except the startup banner as its the default, and may not be for the VS Code extension but for any client that connects to PSES and supports a terminal.

Part of https://github.com/PowerShell/vscode-powershell/issues/4053.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/Configuration/EditorServicesConfig.cs
+++ b/src/PowerShellEditorServices.Hosting/Configuration/EditorServicesConfig.cs
@@ -8,7 +8,7 @@ using System.Management.Automation.Runspaces;
 namespace Microsoft.PowerShell.EditorServices.Hosting
 {
     /// <summary>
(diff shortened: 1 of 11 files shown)


andschwa:
@JustinGrote did I miss anything here too?

src/PowerShellEditorServices.Hosting/Configuration/EditorServicesConfig.cs
```diff
@@ -117,7 +117,7 @@ public sealed class EditorServicesConfig
 
         public string StartupBanner { get; set; } = @"
 
-                  =====> PowerShell Integrated Console <=====
+                  =====> PowerShell Editor Services <=====
```
andschwa, line 120: This is the default banner, not the one used by the extension.

SydneyhSmith (approved):

SeeminglyScience (approved):
LGTM!

PR #1864, opened 2022-07-28
Use `HostInfo.BundledModulePath` to find PSScriptAnalyzer

andschwa:
Also refactor a bit so we're not triplicating this logic.

Resolves https://github.com/PowerShell/vscode-powershell/issues/2697.

I need to add some tests which just ensure PSSA loads and has rules.


SeeminglyScience (approved):
LGTM!

andschwa:
Simplified the test because the `Write-Host` warning I guess doesn't show up with Windows PowerShell?!

PR #1866, opened 2022-08-02
Fix bug where error in `prompt` function crashed REPL loop

andschwa:
Double-whammy fix by both setting `ThrowOnError` to false, and catching a more generic PowerShell `RuntimeException`. Plus a regression test!

Fixes https://github.com/PowerShell/vscode-powershell/issues/4073.


SeeminglyScience (approved):
LGTM!

Issue #1699, opened 2022-02-03
Crash when `prompt` is undefined in the runspace


dkattan:
When "prompt" isn't defined in the runspace the fallback to DefaultPrompt doesn't work because we aren't catching the CommandNotFoundException

https://github.com/PowerShell/PowerShellEditorServices/blob/b433b0f2d1b0e9b93466b5e307a202bfe290c841/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L662-L664

JustinGrote:
Thanks! I don't want to know how you discovered this :P

PR #1867, opened 2022-08-03
Add regression test for when `prompt` is undefined

andschwa:
Also improve regression test for when it throws.

Regression test for https://github.com/PowerShell/PowerShellEditorServices/issues/1699.

### DIFF
--- a/test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs
+++ b/test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs
@@ -190,7 +190,7 @@ namespace PowerShellEditorServices.Test.Session
         public async Task CanHandleUndefinedPrompt()
         {
             Assert.Empty(await psesHost.ExecutePSCommandAsync<PSObject>(
-                new PSCommand().AddScript("Remove-Item function:prompt; Get-Item function:prompt"),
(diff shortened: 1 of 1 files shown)


test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs
```diff
@@ -170,10 +170,29 @@ public async Task CanHandleNoProfiles()
...
                 CancellationToken.None).ConfigureAwait(true);
 
+            Assert.Equal(PsesInternalHost.DefaultPrompt, prompt);
+        }
+
+        [Fact]
+        public async Task CanHandleUndefinedPrompt()
+        {
+            Assert.Empty(await psesHost.ExecutePSCommandAsync<PSObject>(
+                new PSCommand().AddScript("Remove-Item function:prompt; Get-Item function:prompt"),
```
SeeminglyScience, line 193:
```suggestion
                new PSCommand().AddScript("Remove-Item function:prompt; Get-Item function:prompt -ErrorAction Ignore"),
```

This (ideally) writes a non-terminating error. Since we don't care about it, we should set error action ignore in case we ever want to set error action for tests.

SeeminglyScience (approved):
LGTM!

PR #1871, opened 2022-08-04
Delete out-of-date `devcontainer` setup

andschwa:
We'd welcome an up-to-date and tested `devcontainer` in the client repo that works with the existing VS Code workspace, but this one is confusing as it's unused and so untested.

Related to https://github.com/PowerShell/vscode-powershell/issues/3856.


SeeminglyScience (approved):
LGTM!

PR #1873, opened 2022-08-05
Fix (and test) regression with PSScriptAnalyzer default rules

andschwa:
When we swapped a ternary for a null-coalescing operator we ran into a problem because the two constructor overloads took `string[]` and `object`, and we were now calling the constructor first, which meant the `object` version was always being called. This erroneously sent the rule list to the settings object parameter.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4112.


SeeminglyScience (approved):
LGTM!

Issue #1877, opened 2022-08-11
`NullReferenceException` in CI/CD since .NET Core Runtime version 3.1.28


andschwa:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Our last successful CI/CD pipeline used [.NET Core Runtime version 3.1.27](https://dev.azure.com/powershell/PowerShellEditorServices/_build/results?buildId=106187&view=logs&j=f4f26a0c-431b-509f-f476-233b61264efc&t=5c6a8cc2-de2a-5af6-fee5-724b46399078&l=12), and since the update to 3.1.28 it has consistently failed in the completion tests. The delegate which calls `TabExpansion2` is now failing, resulting in a null reference exception (as it's not expected to throw). The inner exception is a file not found exception, where [Microsoft.Bcl.AsyncInterfaces v1.0.0.0](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) and [System.Threading.Tasks.Extensions v4.2.0.0](https://www.nuget.org/packages/System.Threading.Tasks.Extensions/) aren't being found.

Specifically this happens with PowerShell 5.1, which means when we're targeting .NET Desktop Framework 4.6.2.

I am hesitant to blame the .NET update...and yet I was able to immediately repro this locally as soon as I installed the same update.

**This is currently blocking new releases.**

### Expected behavior

```console
N/A
```

### Actual behavior

```console
N/A
```

### Error details

```console
Exception thrown: 'System.Management.Automation.MethodInvocationException' in System.Management.Automation.dll: 'Exception calling "CompleteInput" with "4" argument(s): "Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information."' Inner exceptions found, see $exception in variables window for more details. Innermost exception System.Reflection.ReflectionTypeLoadException : Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
[...]

PR #1878, opened 2022-08-11
Pin .NET to 6.0.302 and 3.1.27

andschwa:
The latest update has blocked the CI/CD pipeline, and so while we investigate that, we'll pin to the last known working version.

Works around https://github.com/PowerShell/PowerShellEditorServices/issues/1877 for now.


SeeminglyScience (approved):
LGTM!

PR #1881, opened 2022-08-11
Only run Windows PowerShell 5.1 tests on CI with `powershell.exe`

andschwa:
And only PowerShell Core on CI with `pwsh.exe`, and re-enable the PowerShell 7.0 tests (oops).


SeeminglyScience (approved):
LGTM!

PR #1894, opened 2022-08-15
Fix stepping while watch expressions or interactive pipeline is running

SeeminglyScience:
# PR Summary

Trying to step when any nested command that wasn't PSRL was running caused us to essentially ignore the request. This caused VSCode to think we were "running" while we didn't actually cancel anything.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
@@ -382,7 +382,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution
 
         private void CancelDebugExecution()
         {
-            if (_pwsh.Runspace.RunspaceStateInfo.IsUsable())
+            if (!_pwsh.Runspace.RunspaceStateInfo.IsUsable())
             {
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
```diff
@@ -382,7 +382,7 @@ internal void MaybeAddToHistory()
 
         private void CancelDebugExecution()
         {
-            if (_pwsh.Runspace.RunspaceStateInfo.IsUsable())
+            if (!_pwsh.Runspace.RunspaceStateInfo.IsUsable())
```
andschwa, line 385: Was this just a big oops?

andschwa (approved):

andschwa:
@SeeminglyScience any chance you could open an issue just so we have this close a bug out?

src/PowerShellEditorServices/Services/PowerShell/Execution/SynchronousPowerShellTask.cs
SeeminglyScience: Oh yeah, I meant to add a review comment pointing this out

BIG ol' oops there

PR #1898, opened 2022-08-16
Add artificial stack frame to represent contexts without one

### LINKED ISSUE #1897: Account for the engine not creating a call stack frame when stepping into param block default value
Tracking for PowerShell/vscode-powershell#4103 on this side

SeeminglyScience:
Fixes #1897 
Fixes PowerShell/vscode-powershell#4103

# PR Summary

When stepping into certain contexts like a `param` block's default value expression, the engine does not provide a call stack frame to represent it. In this scenario we want to create an artificial call stack frame to represent the context we've stepped into.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -994,10 +994,45 @@ namespace Microsoft.PowerShell.EditorServices.Services
                     // Augment the top stack frame with details from the stop event
                     if (invocationTypeScriptPositionProperty.GetValue(e.InvocationInfo) is IScriptExtent scriptExtent)
                     {
-                        stackFrameDetails[0].StartLineNumber = scriptExtent.StartLineNumber;
-                        stackFrameDetails[0].EndLineNumber = scriptExtent.EndLineNumber;
-                        stackFrameDetails[0].StartColumnNumber = scriptExtent.StartColumnNumber;
-                        stackFrameDetails[0].EndColumnNumber = scriptExtent.EndColumnNumber;
+                        StackFrameDetails targetFrame = stackFrameDetails[0];
+
+                        // Certain context changes (like stepping into the default value expression
+                        // of a parameter) do not create a call stack frame. In order to represent
+                        // this context change we create a fake call stack frame.
+                        if (!string.IsNullOrEmpty(scriptExtent.File)
+                            && !PathUtils.IsPathEqual(scriptExtent.File, targetFrame.ScriptPath))
+                        {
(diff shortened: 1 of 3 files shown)


src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -994,10 +994,45 @@ internal async void OnDebuggerStopAsync(object sender, DebuggerStopEventArgs e)
...
-                        stackFrameDetails[0].EndColumnNumber = scriptExtent.EndColumnNumber;
+                        StackFrameDetails targetFrame = stackFrameDetails[0];
+
+                        // Certain context changes (like stepping into the default value expression
+                        // of a parameter) do not create a call stack frame. In order to represent
+                        // this context change we create a fake call stack frame.
+                        if (!string.IsNullOrEmpty(scriptExtent.File)
+                            && !PathUtils.IsPathEqual(scriptExtent.File, targetFrame.ScriptPath))
+                        {
+                            await debugInfoHandle.WaitAsync().ConfigureAwait(false);
```
SeeminglyScience, line 1005: Normally I'd prefer to push this into the code for `FetchStackFramesAndVariablesAsync` but that would require giving it more context than it typically needs. Since this should be an edge case rather than the norm, I lean towards this being preferable.

That said I'm open to switching it up and just passing more context to fetch.

src/PowerShellEditorServices/Utility/PathUtils.cs
```diff
@@ -29,13 +30,48 @@ internal static class PathUtils
         /// </summary>
         internal static readonly char AlternatePathSeparator = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? '/' : '\\';
 
+        /// <summary>
+        /// The <see cref="StringComparison" /> value to be used when comparing paths. Will be
+        /// <see cref="StringComparison.Ordinal" /> for case sensitive file systems and <see cref="StringComparison.OrdinalIgnoreCase" />
+        /// in case insensitive file systems.
+        /// </summary>
+        internal static readonly StringComparison PathComparison = RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
```
SeeminglyScience, line 38: Platform checks are sort of littered throughout the project. It'd be nice to go through and replace these with centralized logic. (It's optimized by the JIT anyway iirc so it's not a perf boost at least outside of Windows PowerShell, but nice for maintenance)

andschwa: 💯 agree!

andschwa (approved):
LGTM!

andschwa:
Ah ha! I figured out how to rebase it myself.

PR #1900, opened 2022-08-17
Add setting to control references code lens

SeeminglyScience:
# PR Summary

This relies on PowerShell/vscode-powershell#

Adds a simple setting and update handler that adds/removes the references code lens. 

## PR Context

This gives folks the ability to disable the high perf impact code lens without disabling all other code lenses.

### DIFF
--- a/src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
+++ b/src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
@@ -22,6 +22,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Configuration
         public CodeFoldingSettings CodeFolding { get; set; }
         public PesterSettings Pester { get; set; }
         public string Cwd { get; set; }
+        public bool EnableReferencesCodeLens { get; set; }
 
         public LanguageServerSettings()
         {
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
```diff
@@ -22,6 +22,7 @@ internal class LanguageServerSettings
         public CodeFoldingSettings CodeFolding { get; set; }
         public PesterSettings Pester { get; set; }
         public string Cwd { get; set; }
+        public bool EnableReferencesCodeLens { get; set; }
```
andschwa, line 25: LMAO this defaulted to `false` which caused the unit test `CanSendReferencesCodeLensRequestAsync` to fail. Should we just default to `true` for now?

andschwa (approved):

src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
andschwa, line 25:
```suggestion
        public bool EnableReferencesCodeLens { get; set; } = true;
```

andschwa (approved):

src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs
SeeminglyScience: Oops! missed this part :) ty!

PR #1886, opened 2022-08-13
Add symbols for class-types and configuration

fflaten:
# PR Summary

Adds support for class-related symbols:
- Class
- Enum
- Property and EnumMember (no references) 
- Methods (no references)
- Constructors (no references)
- Configuration (no references/definition and highlight)

All symbol types have document symbols (outline + symbol search), symbol highlight and hover.
Only class and enum has code lense and references.

Methods and Constructors are shown with parameters in outline + workspace symbol search.

Also:
- Fixes symbol-detection when newline between function-keyword and name (#1905)
- Comment and code cleanup in relevant parts

TODO:
- [x] Manual test
- [x] Add tests

## PR Context

Easier navigation, search and sticky scroll.

Fix #1683
Fix #1905
Related #14 (fixed also? what is parameter hints?)
Related PowerShell/vscode-powershell#3 (missing references for class members)

PR #1911, opened 2022-08-26
Add document symbols for #region

fflaten:
# PR Summary

Adds symbols for regions for easier navigation. Currently available in outline + breadcrumbs and "go to symbol in current editor".

image

## PR Context

Fix PowerShell/vscode-powershell#3604
Fix PowerShell/vscode-powershell#3210

### DIFF
--- a/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
@@ -79,6 +79,7 @@ internal sealed class ReferenceTable
         }
 
         _parent.ScriptAst.Visit(new SymbolVisitor(_parent, AddReference));
+        RegionVisitor.FindRegionsInDocument(_parent, AddReference);
     }
 
     private AstVisitAction AddReference(SymbolReference symbol)
--- a/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
@@ -79,7 +79,6 @@ internal sealed class ReferenceTable
         }
 
         _parent.ScriptAst.Visit(new SymbolVisitor(_parent, AddReference));
-        RegionVisitor.GetRegionsInDocument(_parent, AddReference);
     }
 
     private AstVisitAction AddReference(SymbolReference symbol)
--- a/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
@@ -79,7 +79,7 @@ internal sealed class ReferenceTable
         }
 
         _parent.ScriptAst.Visit(new SymbolVisitor(_parent, AddReference));
-        RegionVisitor.FindRegionsInDocument(_parent, AddReference);
+        RegionVisitor.GetRegionsInDocument(_parent, AddReference);
     }
 
     private AstVisitAction AddReference(SymbolReference symbol)
--- a/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
@@ -160,7 +160,8 @@ namespace Microsoft.PowerShell.EditorServices.Services
             SymbolReference symbol,
             CancellationToken cancellationToken = default)
         {
-            if (symbol is null)
+            // Never scan for regions as they're only useful in current document
+            if (symbol is null || symbol.Type is SymbolType.Region)
             {
                 return Enumerable.Empty<SymbolReference>();
             }
--- a/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
@@ -161,7 +161,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
             CancellationToken cancellationToken = default)
         {
             // Never scan for regions as they're only useful in current document
-            if (symbol is null || symbol.Type is SymbolType.Region)
+            if (symbol is null)
             {
                 return Enumerable.Empty<SymbolReference>();
             }
--- a/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
@@ -160,7 +160,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
             SymbolReference symbol,
             CancellationToken cancellationToken = default)
         {
-            // Never scan for regions as they're only useful in current document
             if (symbol is null)
             {
                 return Enumerable.Empty<SymbolReference>();
--- a/src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
 {
     internal static class RegionVisitor
     {
-        internal static void GetRegionsInDocument(ScriptFile file, Func<SymbolReference, AstVisitAction> action)
+        internal static IEnumerable<SymbolReference> GetRegionsInDocument(ScriptFile file)
         {
             Stack<Token> tokenCommentRegionStack = new();
             Token[] tokens = file.ScriptTokens;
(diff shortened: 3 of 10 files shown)


fflaten:
Consider waiting for #1886 as I'd rather get merge conflicts here 🙂

andschwa:
Heh, ok @fflaten things are so cleaned up now I think you can do this again (much more easily!)

fflaten:
Rewrote to use ReferenceTable for caching, but still a WIP.
@andschwa: Does it make sense to use ReferenceTable here? It'll make it available for handlers incl. find all references. And it isn't something we'd need to track in the workspace.

Todo:
- [ ] Test with full scriptrange (track endregion)
  - How does outline behave when endregion occurs in the middle of a function/class etc?
- [ ] Validate regex behavior against PowerShell ISE
- [ ] Move out of ReferenceTable?
- [ ] Ignore symbol in scenarios like highlight and hover?

andschwa:
> Does it make sense to use ReferenceTable here?

I'd keep using it, it'll make everything easier. We can add "region" as a symbol type and just exclude them where appropriate.

andschwa:
> > Does it make sense to use ReferenceTable here?
> 
> I'd keep using it, it'll make everything easier. We can add "region" as a symbol type and just exclude them where appropriate.

Hm I see what you mean now, since it's not done via AST but simply RegEx, and is _only_ useful for the current document...I'll ponder, I'm sure there's a good way to handle that.

fflaten:
> * [ ]  Test with full scriptrange (track endregion)
>   
>   * How does outline behave when endregion occurs in the middle of a function/class etc?

It didn't break as I feared, but only wraps symbols with both start and end position inside. Ex.
image

Updated to include it and returning only regions with matching end just like folding.
I've disabled hover, highlight (on click) and reference scanning, but left go to symbol for now.

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -79,6 +79,7 @@ internal void EnsureInitialized()
         }
 
         _parent.ScriptAst.Visit(new SymbolVisitor(_parent, AddReference));
+        RegionVisitor.GetRegionsInDocument(_parent, AddReference);
```
andschwa, line 82: @fflaten I'm contemplating a way of moving this out of the reference table. Like ideally I want to call `GetRegionsInDocument` _only_ in `DocumentSymbolHandler` and get the region symbols separately. This would also simplify the rest of the PR because we wouldn't have to special case excluding them.

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -160,7 +160,8 @@ private static string[] GetIdentifiers(string symbolName, SymbolType symbolType,
             SymbolReference symbol,
             CancellationToken cancellationToken = default)
         {
-            if (symbol is null)
+            // Never scan for regions as they're only useful in current document
+            if (symbol is null || symbol.Type is SymbolType.Region)
```
andschwa, line 164: So this would be able to go away.

src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+using System.Management.Automation.Language;
+using Microsoft.PowerShell.EditorServices.Services.TextDocument;
+
+namespace Microsoft.PowerShell.EditorServices.Services.Symbols
+{
+    internal static class RegionVisitor
+    {
+        internal static void GetRegionsInDocument(ScriptFile file, Func<SymbolReference, AstVisitAction> action)
```
andschwa, line 13: I like this a lot, makes reworking it much easier!

src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            Stack<Token> tokenCommentRegionStack = new();
+            Token[] tokens = file.ScriptTokens;
+
+            for (int i = 0; i < tokens.Length; i++)
```
andschwa, line 18: Heh, I hate seeing this over `foreach` but I see `IsBlockComment` takes the index for some reason...

src/PowerShellEditorServices/Services/TextDocument/Handlers/DocumentHighlightHandler.cs
```diff
@@ -53,6 +53,11 @@ protected override DocumentHighlightRegistrationOptions CreateRegistrationOption
             List<DocumentHighlight> highlights = new();
             foreach (SymbolReference occurrence in occurrences)
             {
+                if (occurrence.Type is SymbolType.Region)
```
andschwa, line 56: And this would go away.

src/PowerShellEditorServices/Services/TextDocument/Handlers/HoverHandler.cs
```diff
@@ -52,7 +52,7 @@ public override async Task<Hover> Handle(HoverParams request, CancellationToken
                         request.Position.Line + 1,
                         request.Position.Character + 1).ConfigureAwait(false);
 
-            if (symbolDetails is null)
+            if (symbolDetails is null || symbolDetails.SymbolReference.Type is SymbolType.Region)
```
andschwa, line 55: And this too.

test/PowerShellEditorServices.Test/Language/SymbolsServiceTests.cs
```diff
@@ -821,6 +821,13 @@ public void FindsSymbolsInFile()
             Assert.Equal("prop AValue", symbol.Id);
             Assert.Equal("AValue", symbol.Name);
             Assert.True(symbol.IsDeclaration);
+
+            symbol = Assert.Single(symbols.Where(i => i.Type == SymbolType.Region));
+            Assert.Equal("region my region 123", symbol.Id);
+            Assert.Equal("#region my region 123", symbol.Name);
+            AssertIsRegion(symbol.NameRegion, 50, 1, 50, 22);
+            AssertIsRegion(symbol.ScriptRegion, 50, 1, 52, 11);
+            Assert.True(symbol.IsDeclaration);
```
andschwa, line 830: Thank you 😉

andschwa (changes requested):
If you want to take a crack at what I'm contemplating, go for it! Otherwise I'll try to get to it later today.

andschwa:
> > Does it make sense to use ReferenceTable here?
> 
> I'd keep using it, it'll make everything easier. We can add "region" as a symbol type and just exclude them where appropriate.

I was wrong here 😆

src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
fflaten: It uses index to verify the previous token was a new line.

src/PowerShellEditorServices/Services/Symbols/Visitors/RegionVisitor.cs
fflaten: I assume you meant the separate class, right?
Because the Func-parameter won't make much sense when moving out of ReferenceTable, would it? Or would you prefer a `Action<SymbolReference>` over ex. IEnumerable + LINQ `Concat()`?

src/PowerShellEditorServices/Services/TextDocument/Handlers/DocumentSymbolHandler.cs
```diff
@@ -50,10 +50,7 @@ public override async Task<SymbolInformationOrDocumentSymbolContainer> Handle(Do
             ScriptFile scriptFile = _workspaceService.GetFile(request.TextDocument.Uri);
 
             IEnumerable<SymbolReference> foundSymbols = ProvideDocumentSymbols(scriptFile);
-            if (foundSymbols is null)
-            {
-                return null;
-            }
+            foundSymbols = foundSymbols.Concat(RegionVisitor.GetRegionsInDocument(scriptFile));
```
fflaten, line 54: I didn't see a reason to keep the null-check. The local method `ProvideDocumentSymbols()` always returns a list, right?

test/PowerShellEditorServices.Test/Language/SymbolsServiceTests.cs
```diff
@@ -952,5 +959,30 @@ public void FindsSymbolsInNoSymbolsFile()
             IEnumerable<SymbolReference> symbolsResult = FindSymbolsInFile(FindSymbolsInNoSymbolsFile.SourceDetails);
             Assert.Empty(symbolsResult);
         }
+
+        [Fact]
+        public void FindsRegionSymbolsInFile()
```
fflaten, line 964: This technically doesn't belong here as SymbolsService isn't used for DocumentSymbolHandler. Move to own test-file? Suggestion?

andschwa:
Closing since I integrated this, thanks again so much for your work!

PR #1915, opened 2022-09-01
Change CI build script to only build

andschwa:
Looking at CodeQL build.


andschwa:
That fix for CodeQL is per https://github.com/github/codeql-cli-binaries/issues/131

SeeminglyScience (approved):
LGTM!

PR #1918, opened 2022-09-13
Created a nested PowerShell for the top-level loop

andschwa:
So that events registered with the `OnIdle` engine event handler will always work.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4048

I added the following code to my profile:

```powershell
Register-EngineEvent -SourceIdentifier PowerShell.OnIdle -MaxTriggerCount 10 -Action {
  Write-Host "OnIdle"
}
```
and was able to repro the linked issue before this change, and cannot repro it afterwards.

There may be other ramifications that we're not seeing though, and I'd be curious about a regression test for this.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -475,8 +475,18 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
                 (PowerShell pwsh, RunspaceInfo localRunspaceInfo, EngineIntrinsics engineIntrinsics) = CreateInitialPowerShellSession();
                 _mainRunspaceEngineIntrinsics = engineIntrinsics;
                 _localComputerName = localRunspaceInfo.SessionDetails.ComputerName;
-                _runspaceStack.Push(new RunspaceFrame(pwsh.Runspace, localRunspaceInfo));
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -475,8 +475,19 @@ private void Run()
...
+                // handler, we have to have our top-level PowerShell instance be nested (otherwise
+                // we get a PSInvalidOperationException because pipelines cannot be run
+                // concurrently). Specifically this bug cropped up when a profile loaded code which
+                // registered (and subsequently ran) on the OnIdle handler since it was hitting the
+                // non-nested PowerShell instance. So now we just start with a nested instance.
+                PowerShell nestedPwsh = CreateNestedPowerShell(localRunspaceInfo);
+                _runspaceStack.Push(new RunspaceFrame(nestedPwsh.Runspace, localRunspaceInfo));
+
+                // TODO: Should this now be marked as a Nested frame, even though for our purposes
+                // it isn't really nested?
```
SeeminglyScience, line 489: nah. Two very different meanings of nested. For us we mean nested executions, for PowerShell it basically equates to "on the pipeline thread"

SeeminglyScience (approved):
LGTM! Nice approach ❤️ my take was going to be a little bit more hacky since it would happen earlier in start up. This is better

andschwa:
The plus side of spending that feverish day looking at the `Get-Variable` code was that I got a better understanding of the runspace/nested/availability stuff. It's still pretty obtuse to me, but hey this works.

PR #1924, opened 2022-09-28
Re-implement indicator when running registered editor commands

andyleejordan:
Necessary wiring for https://github.com/PowerShell/vscode-powershell/pull/4187.

### DIFF
--- a/src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
+++ b/src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
@@ -14,6 +14,37 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Server;
 
 namespace Microsoft.PowerShell.EditorServices.Services.Extension
 {
+    /// Enumerates the possible execution results that can occur after
+    /// executing a command or script.
+    /// </summary>
+    internal enum ExecutionStatus
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
```diff
@@ -140,9 +173,27 @@ public Task InvokeCommandAsync(string commandName, EditorContext editorContext,
...
+                        {
+                            status = ExecutionStatus.Completed;
+                        }
+                        else if (executeTask.IsCanceled)
+                        {
+                            status = ExecutionStatus.Aborted;
+                        }
+                        else if (executeTask.IsFaulted)
+                        {
+                            status = ExecutionStatus.Failed;
```
SeeminglyScience, line 189: `ThrowOnError` is false do I don't think this branch is reachable.

...that said we currently don't have a way to surface whether something had errors, so 🤷 I'm cool with leaving it until we figure that out 😁

SeeminglyScience (approved):
LGTM!

src/PowerShellEditorServices/Services/Extension/ExtensionService.cs
andschwa: TBH I have no idea how that would interact with the `Task.IsFaulted` field.

Issue #1855, opened 2022-07-13
Simplify PSES CLI so startup with non-Code editors is easier


andyleejordan:
It takes _way_ too many arguments to launch PSES correctly, which is blocking #1829. Once we simplify this, we could conceivably add a `-lsp` flag to PowerShell itself that would simplify it even further.

smartguy1196:
How definitive is this? Do you intend to completely rewrite the CLI/introduce "breaking changes" with this? I imagine that any changes you make would be backward compatible, but I'm asking, because my PSES manual controller is dependent on how your CLI works.

andschwa:
I don't _intend_ to make breaking changes. The best coverage I have for that is

https://github.com/PowerShell/PowerShellEditorServices/blob/9416a8ae9c7df5a2539731388753e558193c37fb/test/emacs-test.el#L48-L54

and 

https://github.com/PowerShell/PowerShellEditorServices/blob/9416a8ae9c7df5a2539731388753e558193c37fb/test/vim-test.vim#L6-L13

fai8146767:
### ****

PR #1932, opened 2022-10-14
Add example of starting PSES in a script to readme

smartguy1196:
# Problems Solved:

## Shell Hang:

Executing pwsh with the original command (the older one on top) from the readme causes shells to hang.
Not script or automation friendly.

## Poor Doc Readability
The arguments list is also long enough to scroll off the screen on most devices.
Spacing them out over an array is a screen-friendly alternative.

- However, keeping the old one is good for users who just want a quick copy-and-paste one-liner.

# PR Summary

Added an automation-friendly powershell script for starting the service that doesn't hang the shell.

## PR Context

The original command is long (longer than the width of most device screens), so it's not great in documentation, since it needs to be readable. The old one is being left there, so that users have access to a quick and easy copy-and-paste one-liner.

It also hangs the shell. If you run it from the command line, you have to wait for the process to exit before running anything else. Since some users may want to automate the service from PowerShell itself, it is a better practice to call the command with `Start-Process` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,30 @@ pwsh -NoLogo -NoProfile -Command "$PSES_BUNDLE_PATH/PowerShellEditorServices/Sta
 > - `$PSES_BUNDLE_PATH` is the root of the PowerShellEditorServices.zip downloaded from the GitHub releases.
 > - `$SESSION_TEMP_PATH` is the folder path that you'll use for this specific editor session.
 
+If you are trying to automate the service in PowerShell, You can also run it under `Start-Process` to prevent hanging your script. It also gives you access to Process/PID automation features like `$process.Close()` or `$process.Kill()`
+
+```powershell
+$command = @(
+    "$PSES_BUNDLE_PATH/PowerShellEditorServices/Start-EditorServices.ps1",
+        "-BundledModulesPath $PSES_BUNDLE_PATH",
+        "-LogPath $SESSION_TEMP_PATH/logs.log",
+        "-SessionDetailsPath $SESSION_TEMP_PATH/session.json",
+        "-FeatureFlags @()",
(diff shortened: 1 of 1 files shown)


andschwa (approved):
FWIW I also intend (at some point soon-ish) to make this whole thing a heck of a lot easier: https://github.com/PowerShell/PowerShellEditorServices/issues/1855

smartguy1196:
@andschwa I'm currently working on a script that acts as a manual controller for the service. The objective of the script will be to allow end users to manually control the service as well as introduce an opportunity to learn LSP.

My current idea is to setup a basic process and pipe controller, then add functions that allow users to load JSON-RPC messages from files and send them to the server to see what kind of response they get. This also gives the end user the opportunity to learn how to write LSP messages (up to and including the message header)

The interface for the controller will have 2 parts. One, a shell like interface that users can enter commands into, and two, a simple scripting engine for automating those commands.

I got the idea from a different project that I'm working on. It's a powershell command palette that minimizes to the system tray on WIndows/Ubuntu.

Here are the referenced projects and their Hackaday counterparts:

- [smartguy1196/PSES-Controller](https://github.com/smartguy1196/PSES-Controller/tree/test-pses)
  - [hackaday project](https://hackaday.io/project/187779-pses-controller)
- [smartguy1196/PowerShell-Quick-Tray](https://github.com/smartguy1196/PowerShell-Quick-Tray)
  - [hackaday list of projects](https://hackaday.io/list/187741-powershell-quick-tray-journeys)

Once, I finish writing the controller, I might even request to have it referenced within this project so that users can get a tool to playtest LSP and PSES

smartguy1196:
I pulled that bit of code from the controller. Thought it would be useful to other people

PR #1933, opened 2022-10-21
Fix release build

andschwa:
Add assertion to build that bits are built in release configuration and invoke tests in release configuration. Resolves https://github.com/PowerShell/vscode-powershell/issues/4218 by (correctly and how it was expected in the first place) preventing the developer option to wait for debugger to be in the binary.


SeeminglyScience (approved):
LGTM! ❤️

PR #1936, opened 2022-10-24
Fix regression around `OnIdle` handler

andschwa:
Unfortunately the (admittedly hairy) change made in https://github.com/PowerShell/PowerShellEditorServices/pull/1918 to fix:

> System.Management.Automation.PSInvalidOperationException: The pipeline was not run because a pipeline is already running. Pipelines cannot be run concurrently.

has resulted in https://github.com/PowerShell/vscode-powershell/issues/4219:

> System.Management.Automation.PSInvalidOperationException: You should only run a nested pipeline from within a running pipeline.

because we had hacked in that _all_ our pipelines are nested (to allow the concurrency check to pass), but when loading profiles, we do not actually have a running pipeline, so our  lie about being nested fails another check.

So I think what happened was that https://github.com/PowerShell/vscode-powershell/issues/4048 was already no longer reproducing by happenstance (as the Az.Tools.Predictor module presumably updated and removed their `OnIdle` handler). I failed to check that _before_ attempting to fix it following the last result of our investigation. The "fix" inadvertently caused nearly the same `PSInvalidOperationException`, and by mistake I confused the two scenarios. With this PR, and regression tests, I can confirm that the `OnIdle` handler works both in a profile and at the regular prompt after reverting the breaking change. Furthermore, I can no longer repro the original issue with Az.Tools.Predictor whatsoever, in any version, hence my conclusion their module updated.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4219


SeeminglyScience (approved):
LGTM!

PR #1938, opened 2022-10-27
Remove `/p:UseSharedCompilation=false` injection

andschwa:
This was introduced in f234a8e to workaround an upstream bug in `dotnet`. That issue has since been fixed and released (so this flag is auto-injected), hence we can remove it.


andschwa:
Resolves https://github.com/github/codeql-cli-binaries/issues/131

SeeminglyScience (approved):
LGTM!

PR #1940, opened 2022-11-02
Wrap script paths with single instead of double quotes

andschwa:
In order to support funny path names with dollar signs. Also this is what PowerShell does by default (e.g. through tab-completion).

Resolves https://github.com/PowerShell/vscode-powershell/issues/4238.


SeeminglyScience (approved):
LGTM!

PR #1957, opened 2022-11-17
Escape single quotes when launching a script by path

andschwa:
Since we surround it with single quotes.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4269 with tests.


SeeminglyScience (approved):
LGTM!

PR #1958, opened 2022-11-17
Enable VS Code's shell integration

andschwa:
It seems to work, but needs more testing.

I don't particularly like copy/pasting https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/media/shellIntegration.ps1 with modifications to embed it, but not sure what else to do here.

Resolves https://github.com/PowerShell/vscode-powershell/issues/3901 along with https://github.com/PowerShell/vscode-powershell/pull/4271.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -77,6 +77,8 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
 
         private string _localComputerName;
 
+        private bool _shellIntegrationEnabled;
+
         private ConsoleKeyInfo? _lastKey;
 
         private bool _skipNextPrompt;
@@ -252,6 +254,18 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
                 _logger.LogDebug("Loading profiles...");
                 await LoadHostProfilesAsync(cancellationToken).ConfigureAwait(false);
                 _logger.LogDebug("Profiles loaded!");
+            }
+
+            if (startOptions.ShellIntegrationEnabled)
+            {
+                _logger.LogDebug("Enabling shell integration...");
+                _shellIntegrationEnabled = true;
+                await EnableShellIntegrationAsync(cancellationToken).ConfigureAwait(false);
+                _logger.LogDebug("Shell integration enabled!");
+            }
+            else
+            {
+                _logger.LogDebug("Shell integration not enabled!");
             }
 
             if (startOptions.InitialWorkingDirectory is not null)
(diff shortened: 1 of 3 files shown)


andschwa:
Per https://github.com/PowerShell/vscode-powershell/issues/3901#issuecomment-1086104834 we may also want a separate setting to enable/disable this just for the Extension Terminal.

andschwa:
@Tyriar hey we got it! Though like I said, I really don't like my copy/paste method, but I'm unsure if anything else is _better_, we may just have to settle for it.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -962,8 +1065,23 @@ private string InvokeReadLine(CancellationToken cancellationToken)
...
+                    InvokePSCommand(
+                        new PSCommand().AddScript("Write-Host -NoNewLine \"`e]633;C`a\""),
+                        new PowerShellExecutionOptions
+                        {
+                            AddToHistory = false,
+                            ThrowOnError = false,
+                            WriteOutputToHost = true,
+                            FromRepl = true,
+                        },
+                        cancellationToken);
```
SeeminglyScience, line 1082:
```suggestion
                    Console.Write("\x1b]633;C\a");
```

Optional nit, we can just directly write instead of involving the host or PowerShell at all

SeeminglyScience (approved):
LGTM with one optional suggestion

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
andschwa: Brilliant!

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -487,6 +501,96 @@ internal Task LoadHostProfilesAsync(CancellationToken cancellationToken)
                 cancellationToken);
         }
 
+        private Task EnableShellIntegrationAsync(CancellationToken cancellationToken)
+        {
+            // Imported on 11/17/22 from
+            // https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/media/shellIntegration.ps1
+            // with quotes escaped, `__VSCodeOriginalPSConsoleHostReadLine` removed (as it's done
+            // in our own ReadLine function), and `[Console]::Write` replaced with `Write-Host`.
+            // TODO: We can probably clean some of this up.
```
Tyriar, line 510: Clean up PRs to the source are welcome 🙂

andschwa: I wanted to touch it as _little_ as possible for now to make sure we don't break anything.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -487,6 +501,96 @@ internal Task LoadHostProfilesAsync(CancellationToken cancellationToken)
...
+            const string shellIntegrationScript = @"
+# Prevent installing more than once per session
+if (Test-Path variable:global:__VSCodeOriginalPrompt) {
+	return;
+}
+
+# Disable shell integration when the language mode is restricted
```
dkattan, line 517: Hey guys, I know I'm late to the party but I was wondering what the risk of enabling this in Constrained Language mode would be? I run PowerShell in Constrained Language Mode in my web application and use xterm.js and would love to have proper terminal support.

Tyriar: Certain things that we need don't work when not in this mode. 

> in my web application and use xterm.js and would love to have proper terminal support.

There's a bunch of code in vscode to handle shell integration, you wouldn't be able to just drop this script in an xterm.js frontend and have it work

SeeminglyScience: Since we control the host we could move a lot of these VT sequences to C# (e.g. just do `Console.Write` before invoking prompt).

But that would be more explicitly depending on the implementation detail of the script rather than copy pasting it, and as @Tyriar points out it wouldn't help the target scenario.

PR #1961, opened 2022-11-23
Replace backtick-e with `$([char]0x1b)`

andschwa:
By replacing backtick-e with `$([char]0x1b)`, we allow shell integration to work with Windows PowerShell, as the only problem seemed to be use of backtick-e, which was an escape character added only to PowerShell Core, but the `[char]` equivalent works fine.

Same change as proposed upstream in https://github.com/microsoft/vscode/pull/167023, resolves https://github.com/PowerShell/vscode-powershell/issues/4279.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -531,12 +531,12 @@ function Global:Prompt() {
 	if ($Global:__LastHistoryId -ne -1) {
 		if ($LastHistoryEntry.Id -eq $Global:__LastHistoryId) {
 			# Don't provide a command line or exit code if there was no history entry (eg. ctrl+c, enter on no command)
-			$Result  = ""`e]633;E`a""
-			$Result += ""`e]633;D`a""
+			$Result  = ""$([char]0x1b)]633;E`a""
+			$Result += ""$([char]0x1b)]633;D`a""
 		} else {
 			# Command finished command line
 			# OSC 633 ; A ; <CommandLine?> ST
-			$Result  = ""`e]633;E;""
+			$Result  = ""$([char]0x1b)]633;E;""
 			# Sanitize the command line to ensure it can get transferred to the terminal and can be parsed
 			# correctly. This isn't entirely safe but good for most cases, it's important for the Pt parameter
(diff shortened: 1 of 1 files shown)


andschwa:
Thanks @SeeminglyScience for the spot!

SeeminglyScience (approved):
LGTM!

Hrxn:
Do I get this right, this is only a change for the sake of backward compatibility?

msftrncs:
@Hrxn, Not necessarily backward compatible, more of a cross platform compatibility.  Windows PowerShell is a still active currently supported platform.  Consider that PowerShell Core is an enhanced platform, rather than a 'newer' version.

SeeminglyScience:
> Do I get this right, this is only a change for the sake of backward compatibility?

That's correct, the syntax `` `e`` only works in newer versions of PowerShell.


> @Hrxn, Not necessarily backward compatible, more of a cross platform compatibility. Windows PowerShell is a still active currently supported platform. Consider that PowerShell Core is an enhanced platform, rather than a 'newer' version.

Windows PowerShell receives primarily only security updates. PowerShell 7 is definitely considered a newer version.

Hrxn:
Thank you for this explanation, appreciate it.
What I've always wondered, is there also some kind of performance difference between using
 `"$([char]0x1b)<ANSI_Esc_Sequence>"` and ` "backtick-e<ANSI_Esc_Sequence>" ` ?

I've done some test here on my machine with `Measure-Command`, but I could not find any conclusive result somehow 🤷‍♂️

SeeminglyScience:
Technically `` `e`` will be faster as it'll be emitted by the compiler as a constant. Not enough to really matter though, use `` `e`` where you don't need to support 5.1 since it's easier to read. Otherwise don't stress it

Hrxn:
Thanks, glad that I finally found an answer for what has been a pretty longstanding question about powershell for me, to be honest.. 😉
Will use `` `e `` from now on, because I don't need/want to support 5.1

SeeminglyScience:
> Thanks, glad that I finally found an answer for what has been a pretty longstanding question about powershell for me, to be honest.. 😉 Will use `` `e `` from now on, because I don't need/want to support 5.1

If you aren't already there, check out the [PowerShell Discord](https://aka.ms/psdiscord) server. Plenty of folks hanging out ready to answer fun questions like that 😁

PR #1962, opened 2022-11-28
Revert manual pin of Newtonsoft.Json

andschwa:
Pinning this in https://github.com/PowerShell/PowerShellEditorServices/pull/1947 before getting our transitive dependencies in order seems to be the root cause of https://github.com/PowerShell/vscode-powershell/issues/4278 and possibly also https://github.com/PowerShell/vscode-powershell/issues/4283. We're going to revert until we get OmniSharp's dependency sorted, and test in preview with users if those issues are resolved.


SeeminglyScience (approved):
LGTM!

PR #1964, opened 2022-12-02
Remove unnecessary `PowerShellProcessArchitecture`

andschwa:
This simply wasn't being used and was overly complicated. The only time we want the architecture is when queried via LSP so that the VS Code client can determine which installer to download for its auto-update feature. If we can, we should deduplicate version logic in the loader and the `VersionUtils` class.

Simplification found when investigating https://github.com/PowerShell/vscode-powershell/issues/3435.


SeeminglyScience (approved):
LGTM!

PR #1965, opened 2022-12-06
Send `GitCommitId` over `GetVersionHandler`

andschwa:
So the client can skip updates for daily and dev builds. Also clean up a little more, the commit ID was not actually being used in the rest of the codebase, and the `DisplayVersion` was dumb.

Required for https://github.com/PowerShell/vscode-powershell/pull/4306.


SeeminglyScience (approved):
LGTM!

Issue #1879, opened 2022-08-11
Deprecate PowerShell 7.0/7.1 and .NET Core 3.1


andschwa:
### Summary of the new feature / enhancement

When these dependencies hit end-of-life [this December](https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/) we need to remove them from the build.

andschwa:
* https://github.com/PowerShell/PowerShellEditorServices/pull/1952
* https://github.com/PowerShell/PowerShellEditorServices/pull/1950

andschwa:
Following https://learn.microsoft.com/en-us/aspnet/core/migration/31-to-60?view=aspnetcore-6.0&tabs=visual-studio

PR #1967, opened 2022-12-06
Drop support for end-of-life PowerShell 7.0

andschwa:
Which more importantly means we can drop `netcoreapp3.1`. Resolves https://github.com/PowerShell/PowerShellEditorServices/issues/1879.


SeeminglyScience (approved):
LGTM!

PR #1970, opened 2022-12-09
Fix `ShowHelpHandler` by running with `RequiresForeground`

andschwa:
This was broken after the rewrite because we'd forgotten to specify that it needs the foreground (which implies too that it must run immediately). Now the script actually executes, either showing the help in the Extension Terminal (or...I guess nowhere for clients without it) or opening the URL it found. This should still be rewritten properly to return the help over LSP and do something better with it, but at least it's no longer broken.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4212.


SeeminglyScience (approved):
LGTM!

PR #1977, opened 2022-12-20
Remove supposedly unnecessary lambdas from `Task.Run`

andschwa:
Except I think this might be causing Windows PowerShell to fail.


andschwa:
Well, wasn't this.

PR #1981, opened 2022-12-28
Make `Set-ScriptExtent` not slow

SeeminglyScience:
This was some of the first code I ever contributed...and it shows. It's actually really easy to solve the problem of "when editing many parts of the same document, offsets no longer match up". You just go in reverse, and then you don't have that problem anymore.

So this gets rid of the "waiting for the document" to update which takes forever.

### DIFF
--- a/module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
@@ -5,27 +5,29 @@ function Set-ScriptExtent {
     <#
     .EXTERNALHELP ..\PowerShellEditorServices.Commands-help.xml
     #>
-    [CmdletBinding(PositionalBinding=$false, DefaultParameterSetName='__AllParameterSets')]
+    [CmdletBinding(PositionalBinding = $false, DefaultParameterSetName = '__AllParameterSets')]
     param(
-        [Parameter(Position=0, Mandatory)]
-        [psobject]
-        $Text,
+        [Parameter(Position = 0, Mandatory)]
+        [psobject] $Text,
 
-        [Parameter(Mandatory, ParameterSetName='AsString')]
+        [Parameter(Mandatory, ParameterSetName = 'AsString')]
         [switch]
         $AsString,
 
-        [Parameter(Mandatory, ParameterSetName='AsArray')]
-        [switch]
-        $AsArray,
+        [Parameter(Mandatory, ParameterSetName = 'AsArray')]
+        [switch] $AsArray,
 
         [Parameter(ValueFromPipeline, ValueFromPipelineByPropertyName)]
-        [System.Management.Automation.Language.IScriptExtent]
-        $Extent = (Find-Ast -AtCursor).Extent
+        [System.Management.Automation.Language.IScriptExtent] $Extent = (Find-Ast -AtCursor).Extent
     )
     begin {
         $fileContext = $psEditor.GetEditorContext().CurrentFile
-        $extentList = [System.Collections.Generic.List[[Microsoft.PowerShell.EditorServices.Extensions.FileScriptExtent, Microsoft.PowerShell.EditorServices]]]::new()
+        $descendingComparer = [System.Collections.Generic.Comparer[int]]::Create{
+            param($x, $y) return $y.CompareTo($x)
+        }
+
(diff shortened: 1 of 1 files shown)


module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
```diff
@@ -5,27 +5,29 @@ function Set-ScriptExtent {
...
+        [switch] $AsArray,
 
         [Parameter(ValueFromPipeline, ValueFromPipelineByPropertyName)]
-        [System.Management.Automation.Language.IScriptExtent]
-        $Extent = (Find-Ast -AtCursor).Extent
+        [System.Management.Automation.Language.IScriptExtent] $Extent = (Find-Ast -AtCursor).Extent
     )
     begin {
         $fileContext = $psEditor.GetEditorContext().CurrentFile
-        $extentList = [System.Collections.Generic.List[[Microsoft.PowerShell.EditorServices.Extensions.FileScriptExtent, Microsoft.PowerShell.EditorServices]]]::new()
```
SeeminglyScience, line 28: Before this point is all style fixes

module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
```diff
@@ -34,11 +36,11 @@ function Set-ScriptExtent {
...
                 $Extent.EndOffset)
         }
-        $extentList.Add($Extent)
+
+        $extentList.Add($Extent.StartOffset, $Extent)
     }
-    # Currently this kills the pipeline because we need to keep track and edit all extents for position tracking.
-    # TODO: Consider queueing changes in a static property and adding a PassThru switch.
     end {
+        $needsIndentFix = $false
```
SeeminglyScience, line 43: Defensively assigning to avoid potentially getting a variable with the same name from a parent scope

module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
```diff
@@ -34,11 +36,11 @@ function Set-ScriptExtent {
                 $Extent.StartOffset,
                 $Extent.EndOffset)
         }
-        $extentList.Add($Extent)
+
+        $extentList.Add($Extent.StartOffset, $Extent)
     }
-    # Currently this kills the pipeline because we need to keep track and edit all extents for position tracking.
-    # TODO: Consider queueing changes in a static property and adding a PassThru switch.
```
SeeminglyScience, line 40: It doesn't kill the pipeline, it just stops streaming behavior. Doesn't really need a comment.

Also the `TODO` doesn't make any sense

module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
```diff
@@ -55,41 +57,16 @@ function Set-ScriptExtent {
...
+        foreach ($kvp in $extentList.GetEnumerator()) {
+            $aExtent = $kvp.Value
             $aText = $Text
 
             if ($needsIndentFix) {
-                # I'd rather let PSSA handle this when there are more formatting options.
                 $indentOffset = ' ' * ($aExtent.StartColumnNumber - 1)
-                $aText = $aText -split '\r?\n' `
-                                -join ([Environment]::NewLine + $indentOffset)
+                $aText = $aText -split '\r?\n' -join ([Environment]::NewLine + $indentOffset)
```
SeeminglyScience, line 66: This is just formatting. The rest of this should probably be revisited at some point as well, but not actively bad afaik

andschwa (approved):
Nice work.

module/PowerShellEditorServices/Commands/Public/Set-ScriptExtent.ps1
```diff
@@ -5,27 +5,29 @@ function Set-ScriptExtent {
...
-        [Parameter(Position=0, Mandatory)]
-        [psobject]
-        $Text,
+        [Parameter(Position = 0, Mandatory)]
+        [psobject] $Text,
 
-        [Parameter(Mandatory, ParameterSetName='AsString')]
+        [Parameter(Mandatory, ParameterSetName = 'AsString')]
         [switch]
         $AsString,
```
Hrxn, line 15: Is it intentional that `$AsString,` is still kept on its own line here?

SeeminglyScience: Nah not intentional, just missed it

PR #1982, opened 2023-01-03
Update VS Code shell integration script

andschwa:
Updates our embedded copy with changes made upstream, see https://github.com/microsoft/vscode/pull/170516.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4343.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -503,11 +503,10 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
 
         private Task EnableShellIntegrationAsync(CancellationToken cancellationToken)
         {
-            // Imported on 11/17/22 from
+            // Imported on 01/03/23 from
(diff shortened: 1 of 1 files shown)


andschwa:
@SeeminglyScience @tyriar I think the change we make from `Console::Write` to `Write-Host -NoNewLine` could be shipped upstream, and I'm wondering if we should change:

```pwsh
# Only send the command executed sequence when PSReadLine is loaded, if not shell integration should
# still work thanks to the command line sequence
if (Get-Module -Name PSReadLine) {
	$__VSCodeOriginalPSConsoleHostReadLine = $function:PSConsoleHostReadLine
...
}
```

to

```pwsh
if (Get-Module -Name PSReadLine -and $Host.Name -ne "Visual Studio Code Host") {
	$__VSCodeOriginalPSConsoleHostReadLine = $function:PSConsoleHostReadLine
...
}
```

then at least I won't have to change those two things when importing the script, and that gets us closer to somehow doing what I think is the "Right Thing" and getting either PSES or VS Code to load the upstream shell integration script even for PSES Extension Terminals.

SeeminglyScience (approved):
LGTM!

Issue #1889, opened 2022-08-14
Function reference not counted in codelens when on same line as definition


fflaten:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Create a function and invoke it on the same line. See image.

### Expected behavior

```console
Invocation is counted in codelens
```

### Actual behavior

```console
Invocation isn't counted in codelense.
Go to reference detects the invocation as expected.
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.2.6
PSEdition                      Core
GitCommitId                    7.2.6
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

2022.8.2 preview

### Visuals

![image](https://user-images.githubusercontent.com/3436158/184547912-b7f4540a-2af6-4aa7-a6e9-a65a26824e38.png)

This issue also means this will show 2 references in code lense because it counts the function-name in definition itself:

```powershell
function
funcNewline {
    'hello'
}

funcNewline
```

fflaten:
Looks like the problem is here where it only compares file and startline.
https://github.com/PowerShell/PowerShellEditorServices/blob/948dffe080cb93b6c2d6b8fb5544788d2bcb5a83/src/PowerShellEditorServices/Services/CodeLens/ReferencesCodeLensProvider.cs#L170-L183

Suggestions
1) **Quick fix:** Compare reference startcolumn not equal

2) Should CommandAst use the same `SymbolType.Function` as definitions? Not extend with a new `SymbolType.Command` for invocations? Then we could use that for found references in [FindReferencesVisitor.VisitCommand(CommandAst)](https://github.com/PowerShell/PowerShellEditorServices/blob/948dffe080cb93b6c2d6b8fb5544788d2bcb5a83/src/PowerShellEditorServices/Services/Symbols/Vistors/FindReferencesVisitor.cs#L96-L104)? *I assume that's the source, haven't stepped through yet*

andschwa:
Theoretically fixed by https://github.com/PowerShell/PowerShellEditorServices/pull/1984

Issue #1905, opened 2022-08-21
Outline, symbol search and go to definition should jump to symbol name


fflaten:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

```powershell
function
MyFunc123 {
    'some code'
}

MyFunc123
```

- Select the last line and press F12
- Find `MyFunc123` in symbol search
- Click in outline

### Expected behavior

```console
Jumps to line 2 `MyFun123` like other languages like C# and TypeScript
```

### Actual behavior

```console
Jumps to `function`
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.2.6
PSEdition                      Core
GitCommitId                    7.2.6
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

2022.8.3 preview

andschwa:
I think fixed by https://github.com/PowerShell/PowerShellEditorServices/pull/1984

Issue #1904, opened 2022-08-21
CodeLens range not following LSP-spec


fflaten:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Debug extension or see LSP log while it's looking up code lenses for a function definition, Pester block etc.

### Expected behavior

```console
interface CodeLens {
	/**
	 * The range in which this code lens is valid. Should only span a single
	 * line.
	 */
	range: Range;
...

Source: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens
```

### Actual behavior

```console
Range is for full definition, spanning multiple lines.
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.2.5
PSEdition                      Core
GitCommitId                    7.2.5
OS                             Linux 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

2022.8.3 preview

### Visuals

![image](https://user-images.githubusercontent.com/3436158/185794657-15605eae-52ee-4857-98ee-d3d7071ab5b2.png)

fflaten:
Code lens should also follow the symbol name. Currently it's placed above extent start (function keyword), but should follow symbol name in case function-keyword is followed by a linebreak.

Now:
![image](https://user-images.githubusercontent.com/3436158/185795272-fd8688b9-3538-43db-90fd-f94eb0fab5e9.png)

Suggestion:
![image](https://user-images.githubusercontent.com/3436158/185795376-5fdb251b-7c57-48d2-bf07-a36d11bff11d.png)

Like C#:
![image](https://user-images.githubusercontent.com/3436158/185795481-aaf4e00b-b1c7-4bc7-a27b-91cdcc096f46.png)

andschwa:
I think fixed by https://github.com/PowerShell/PowerShellEditorServices/pull/1984

PR #1984, opened 2023-01-12
Integrating class symbol support

andschwa:
This is the full (admittedly still a bit of a work-in-progress PR) that integrates https://github.com/PowerShell/PowerShellEditorServices/pull/1886 into `main`.


src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -19,32 +21,46 @@ internal sealed class ReferenceTable
...
+    {
+        EnsureInitialized();
+        List<SymbolReference> allReferences = new();
+        foreach (ConcurrentBag<SymbolReference> bag in _symbolReferences.Values)
+        {
+            allReferences.AddRange(bag);
+        }
+        return allReferences;
+    }
+
```
andschwa, line 63: @SeeminglyScience is there a better way to do this?

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -55,62 +71,228 @@ internal void EnsureInitialized()
         _parent.ScriptAst.Visit(new ReferenceVisitor(this));
     }
 
-    private void AddReference(string symbol, IScriptExtent extent)
+    private static bool ExtentIsEmpty(IScriptExtent e) => string.IsNullOrEmpty(e.File) &&
+        e.StartLineNumber == 0 && e.StartColumnNumber == 0 &&
+        e.EndLineNumber == 0 && e.EndColumnNumber == 0 &&
+        string.IsNullOrEmpty(e.Text);
```
andschwa, line 77: I pulled this from `IsEmptyExtent` in PowerShell's own codebase.

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -55,62 +71,228 @@ internal void EnsureInitialized()
...
             });
     }
 
-    private sealed class ReferenceVisitor : AstVisitor
+    // TODO: Reconstruct this to take an action lambda that returns a visit action and accepts a
+    // symbol. Then ReferenceTable can add a reference and find symbol can just stop.
+    //
+    // TODO: Have a symbol name and a separate display name, the first minimally the text so the
+    // buckets work, the second usually a more complete signature for e.g. outline view.
+    private sealed class ReferenceVisitor : AstVisitor2
```
andschwa, line 103: Big TODOs really.

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -55,62 +71,228 @@ internal void EnsureInitialized()
...
+        {
+            string? memberName = methodCallAst.Member is StringConstantExpressionAst stringConstant ? stringConstant.Value : null;
+            if (string.IsNullOrEmpty(memberName))
+            {
+                return AstVisitAction.Continue;
+            }
+
+            _references.AddReference(
+                SymbolType.Method,
+                memberName,
```
andschwa, line 280: I literally check `IsNullOrEmpty` so IDK why it thinks it could be null here.

andschwa (commented):
Some quick comments but there's more, I just need to eat lunch.

fflaten:
This is coming along very well! I'll try to consolidate my thoughts in this PR 🙂 

- Need to filter on symboltype when searching symbols. Function name and and class name match, Go to definition on `[type]` shows constructors etc. See https://github.com/fflaten/PowerShellEditorServices/pull/16#discussion_r1070663281
image

- Since it matches `InvokeMemberExpressions` with methods, maybe consider matching `[classname]::new()` with `classname()` constructors? https://github.com/fflaten/PowerShellEditorServices/pull/16#discussion_r1083286012

- Missing method overload in outline and workspace symbol search
image
  - Personally I'd also filter out the method parameters shown as variables atm. 

- Only show variable assignment symbols for top-level assignments like stable? To remove noise in outline and workspace symbols - at least until we're able to handle scopes better. Related discussion https://github.com/fflaten/PowerShellEditorServices/pull/16#issuecomment-1383191404
  - Also for go to definition? Will currently list reassignment inside of out-of-scope function

- Outline, workspace symbol search and go to definition jumps to function-keyword and not name (using scriptrange atm?). So this undoes fix for https://github.com/PowerShell/PowerShellEditorServices/issues/1905 in original PR.

- References codelens is painfully slow to show up or update at times. 
  Repro: Create a new tab, select PowerShell as language and paste in:
  ```ps1
  function
  funcNewline2 {
      'hello'
  }
  
  funcNewline2
  ```
  This takes <1sec in latest preview-version, but ~10sec in PR (debug mode)

This will also fix #1889 and #1904 🎉
image

fflaten:
Latest commit fixed the reference-issue with large files 🎉 Update previous post

andschwa:
Ok @fflaten and @SeeminglyScience try it out!

> Need to filter on symboltype when searching symbols.

I'm not quite sure how to do this so that it's totally right. Want to punt until after this PR and come fix it for me? 😄 

> Consider matching [classname]::new() with classname() constructors?

Definitely punting this.

> Only show variable assignment symbols for top-level assignments like stable?

Any idea how to do this? I talked with Patrick and there's not really a right way to distinguish between the initial assignment and later assignments, they all just trigger `VisitVariableExpression` and then we set `isDeclaration: variableExpressionAst.Parent is AssignmentStatementAst or ParameterAst`.

> Outline, workspace symbol search and go to definition jumps

I set workspace and definition locations to `NameRegion` and multiple-definitions still work (I kinda liked it the other way around lol) but am not sure how to fix jumping from the outline view because it needs the whole range in order to figure out what's a child of what. Once we backlink symbols to their parents in the future, that's easily fixed, but for now 🤷

I just want to get the tests passing and get this merged at this point so we can put it in preview, and you can open PRs all you want to it 🥲

src/PowerShellEditorServices/Services/TextDocument/Handlers/DocumentSymbolHandler.cs
```diff
@@ -44,44 +44,66 @@ protected override DocumentSymbolRegistrationOptions CreateRegistrationOptions(D
...
+                //     SelectionRange = r.NameRegion.ToRange()
+                // }));
+                symbols.Add(new SymbolInformationOrDocumentSymbol(new SymbolInformation
+                {
+                    ContainerName = containerName,
+                    Kind = SymbolTypeUtils.GetSymbolKind(r.SymbolType),
+                    Location = new Location
+                    {
+                        Uri = DocumentUri.From(r.FilePath),
+                        Range = r.ScriptRegion.ToRange() // The whole thing, not just the name.
```
fflaten, line 95:
```suggestion
                        Range = new Range(r.NameRegion.ToRange().Start, r.ScriptRegion.ToRange().End) // Jump to name start, but keep whole range to support symbol tree in outline
```

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -106,288 +107,178 @@ internal class SymbolsService
...
 
-            if (foundSymbol == null)
+            if (symbol is null)
             {
                 return null;
             }
 
-            return AstOperations.FindReferencesOfSymbol(file.ScriptAst, foundSymbol).ToArray();
+            scriptFile.References.TryGetReferences(symbol.SymbolName, out ConcurrentBag<SymbolReference>? references);
+            return references;
```
fflaten, line 219:
```suggestion
            return references.Where((r) =>
                    symbol.SymbolType == r.SymbolType ||
                    (symbol.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type
                        && r.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type) ||
                    (symbol.SymbolType is SymbolType.EnumMember or SymbolType.Property
                        && r.SymbolType is SymbolType.EnumMember or SymbolType.Property)
                );
```

As a non-professional developer I'm not above this, except maybe extract it. 😅 
Will need something similar for Go to Definition.

fflaten:
> Ok @fflaten and @SeeminglyScience try it out!

Awesome. Just had a few minutes now, but will give it a good try later.

> > Need to filter on symboltype when searching symbols.
> 
> I'm not quite sure how to do this so that it's totally right. Want to punt until after this PR and come fix it for me? 😄

See comment. Can also be fixed later 👍 

> 
> > Only show variable assignment symbols for top-level assignments like stable?
> 
> Any idea how to do this? I talked with Patrick and there's not really a right way to distinguish between the initial assignment and later assignments...

Agreed. Wouldn't be able to distinguish those, but in the handler we could filter out variable-symbols with a `Parent`-reference.

Had a PoC (probably horrible solution), but I might need to adjust it after the last changes. TBC.

> 
> > Outline, workspace symbol search and go to definition jumps
> 
> I set workspace and definition locations to `NameRegion` and multiple-definitions still work (I kinda liked it the other way around lol) but am not sure how to fix jumping from the outline view because it needs the whole range in order to figure out what's a child of what. Once we backlink symbols to their parents in the future, that's easily fixed, but for now 🤷

👍 Outline just needs to start on the name. There's no children between keyword and symbolname so should be fine as temp solution. See comment.

> I just want to get the tests passing and get this merged at this point so we can put it in preview, and you can open PRs all you want to it 🥲

I just got a deja vu-feeling from writing my PR - and I completely suppport this. 😄 
We'll be able to iron out the rest in preview or before.

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -408,123 +299,56 @@ static string[] GetIdentifiers(string symbolName, SymbolType symbolType, Command
...
-                            foundSymbol);
-
-                    filesSearched.Add(file);
-                    if (foundDefinition is not null)
-                    {
-                        foundDefinition.FilePath = file;
-                        break;
-                    }
-                }
+                return declarations;
```
fflaten, line 328:
```suggestion
                return declarations.Where((d) =>
                    symbol.SymbolType == d.SymbolType ||
                    (symbol.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type
                        && d.SymbolType is SymbolType.Class or SymbolType.Enum) ||
                    (symbol.SymbolType is SymbolType.EnumMember or SymbolType.Property
                        && d.SymbolType is SymbolType.EnumMember)
                );
```

Fix go to definition symbol type mismatch (part 1)

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -408,123 +299,56 @@ static string[] GetIdentifiers(string symbolName, SymbolType symbolType, Command
...
-                        foundSymbol);
+                if (foundReference.IsDeclaration)
+                {
+                    _logger.LogDebug($"Found possible declaration in workspace ${foundReference}");
+                    declarations.Add(foundReference);
+                }
             }
 
-            return foundDefinition;
+            return declarations;
```
fflaten, line 341:
```suggestion
            return declarations.Where((d) =>
                    symbol.SymbolType == d.SymbolType ||
                    (symbol.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type
                        && d.SymbolType is SymbolType.Class or SymbolType.Enum) ||
                    (symbol.SymbolType is SymbolType.EnumMember or SymbolType.Property
                        && d.SymbolType is SymbolType.EnumMember)
                );
```

Fix go to definition symbol type mismatch (part 2)

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
```diff
@@ -106,288 +107,178 @@ internal class SymbolsService
...
                     }
 
-                    await Task.Yield();
-                    cancellationToken.ThrowIfCancellationRequested();
+                    symbols.AddRange(references);
                 }
             }
 
-            return symbolReferences;
+            return symbols;
```
fflaten, line 202:
```suggestion
            return symbols.Where((reference) =>
                    symbol.SymbolType == reference.SymbolType ||
                    (symbol.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type
                        && reference.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type) ||
                    (symbol.SymbolType is SymbolType.EnumMember or SymbolType.Property
                        && reference.SymbolType is SymbolType.EnumMember or SymbolType.Property)
                );
```

Fix code lense reference count

src/PowerShellEditorServices/Services/Symbols/SymbolsService.cs
fflaten, line 219:
```suggestion
            return references.Where((r) =>
                    symbol.SymbolType == r.SymbolType ||
                    (symbol.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type
                        && r.SymbolType is SymbolType.Class or SymbolType.Enum or SymbolType.Type) ||
                    (symbol.SymbolType is SymbolType.EnumMember or SymbolType.Property
                        && r.SymbolType is SymbolType.EnumMember or SymbolType.Property)
                );
```

Fix symbol highlight etc.

fflaten (commented):
Added a few ideas to match symbol type.
As a non-professional developer I'm not above this method, except maybe extract and reuse the duplicates (2-3). 😅

src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -0,0 +1,231 @@
...
+            _file,
+            isDeclaration: true));
+    }
+
+    public override AstVisitAction VisitTypeExpression(TypeExpressionAst typeExpressionAst)
+    {
+        return _action(new SymbolReference(
+            SymbolType.Type,
+            typeExpressionAst.TypeName.Name,
+            typeExpressionAst.TypeName.Name,
```
fflaten, line 122: Nit. Same as `VisitTypeConstraint` or change both to `"type" + typeExpressionAst.TypeName.Name`? No preference, but `symbol name<overload>` as pattern in general for displaynames?

src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            return AstVisitAction.Continue;
+        }
+
+        return _action(new SymbolReference(
+            SymbolType.Property,
+#pragma warning disable CS8604 // Possible null reference argument.
+            memberName,
+#pragma warning restore CS8604
+            memberExpressionAst.Member.Extent.Text,
```
fflaten, line 192: Nit, `symboltype name`?

src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            return AstVisitAction.Continue;
+        }
+
+        return _action(new SymbolReference(
+            SymbolType.Method,
+#pragma warning disable CS8604 // Possible null reference argument.
+            memberName,
+#pragma warning restore CS8604
+            methodCallAst.Member.Extent.Text,
```
fflaten, line 212: Nit, same as above.

fflaten:
Tested (no code review). I think it looks good - excellent work @andschwa !

I'd prefer to get the symbol type matching in now as that's a straight up bug. See review comment, incl. update about using the last where-epxression for all four places.

As for the variable assignment-discussion we can see what people say in preview-release. I'm curious and will look at python extension to see how they've approached it 🙂

andschwa:
Ok @fflaten got that taken care of, and I fixed a bug I found when @SeeminglyScience decided to skip children when visiting parameters. I removed that and instead skip parameters in the variable expression visitor (so they're not double visited), but we _do_ want to visit the children (specifically the type constraints!).

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
```diff
@@ -19,30 +20,55 @@ internal sealed class ReferenceTable
...
-        return _symbolReferences.TryGetValue(command, out references);
+        return symbol is not null
+            && _symbolReferences.TryGetValue(symbol.SymbolName, out ConcurrentBag<SymbolReference>? bag)
+                ? bag.Where(i => SymbolTypeUtils.SymbolTypeMatches(symbol.SymbolType, i.SymbolType))
+                : Enumerable.Empty<SymbolReference>();
+    }
+
+    // Gets symbol whose name contains the position
+    internal SymbolReference? TryGetSymbolAtPosition(int line, int column) => GetAllReferences()
+        .FirstOrDefault(i => i.NameRegion.ContainsPosition(line, column));
```
SeeminglyScience, line 56: Ah so this is what you were talkin' about, I see what you mean now. Yeah ideally we're revisit this sometime after this PR because I can't imagine it scales well. We'll see if anything pops up in preview though

src/PowerShellEditorServices/Services/Symbols/SymbolType.cs
```diff
@@ -41,6 +43,74 @@ internal enum SymbolType
...
-        HashtableKey
+        HashtableKey,
+
+        /// <summary>
+        /// The symbol is a class
+        /// </summary>
+        Class,
+
+        /// <summary>
+        /// The symbol is a enum
```
SeeminglyScience, line 54: does that mean `[SomeEnum]` or does it mean `enum SomeEnum`?

andschwa: `enum SomeEnum` since `[SomeEnum]` is a type constraint (and so is a `SymbolType.Type`).

src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
andschwa: Eh, it hasn't seemed bad at all. It's faster than the previous logic which was to revisit the AST, checking each (newly found and created symbol)'s extent to see if it contained the position. We cut 3/4's of that out, since now we just check essentially a cached set of symbols.

src/PowerShellEditorServices/Services/TextDocument/ScriptRegion.cs
```diff
@@ -12,64 +14,42 @@ namespace Microsoft.PowerShell.EditorServices.Services.TextDocument
     /// </summary>
     public sealed class ScriptRegion : IScriptExtent
     {
-        #region Static Methods
+        public TextEdit ToTextEdit() => new() { NewText = Text, Range = ToRange() };
```
SeeminglyScience, line 17:
```suggestion
        internal TextEdit ToTextEdit() => new() { NewText = Text, Range = ToRange() };
```

This class is public and surfaced, I'd recommend keeping these internal unless you were explicitly intending on providing it as an extension point

src/PowerShellEditorServices/Services/TextDocument/ScriptRegion.cs
```diff
@@ -12,64 +14,42 @@ namespace Microsoft.PowerShell.EditorServices.Services.TextDocument
...
+        {
+            return StartLineNumber == 0 && StartColumnNumber == 0
+                && EndLineNumber == 0 && EndColumnNumber == 0
+                && string.IsNullOrEmpty(File)
+                && string.IsNullOrEmpty(Text);
         }
 
-        internal static TextEdit ToTextEdit(ScriptRegion scriptRegion)
+        // Do not use PowerShell's ContainsLineAndColumn, it's nonsense.
+        public bool ContainsPosition(int line, int column)
```
SeeminglyScience, line 46:
```suggestion
        internal bool ContainsPosition(int line, int column)
```

src/PowerShellEditorServices/Services/TextDocument/ScriptRegion.cs
```diff
@@ -12,64 +14,42 @@ namespace Microsoft.PowerShell.EditorServices.Services.TextDocument
...
-                scriptExtent.File,
-                scriptExtentText,
-                scriptExtent.StartLineNumber,
-                scriptExtent.StartColumnNumber,
-                scriptExtent.StartOffset,
-                scriptExtent.EndLineNumber,
-                scriptExtent.EndColumnNumber,
-                scriptExtent.EndOffset);
+        // Same as PowerShell's EmptyScriptExtent
+        public bool IsEmpty()
```
SeeminglyScience, line 37:
```suggestion
        internal bool IsEmpty()
```

src/PowerShellEditorServices/Services/TextDocument/ScriptRegion.cs
```diff
@@ -12,64 +14,42 @@ namespace Microsoft.PowerShell.EditorServices.Services.TextDocument
...
-        /// Creates a new instance of the ScriptRegion class from an
-        /// instance of an IScriptExtent implementation.
-        /// </summary>
-        /// <param name="scriptExtent">
-        /// The IScriptExtent to copy into the ScriptRegion.
-        /// </param>
-        /// <returns>
-        /// A new ScriptRegion instance with the same details as the IScriptExtent.
-        /// </returns>
-        public static ScriptRegion Create(IScriptExtent scriptExtent)
```
SeeminglyScience, line 27: No issues with adding the ctor but we need to keep the public method otherwise it's a binary breaking change.

src/PowerShellEditorServices/Utility/VisitorUtils.cs
```diff
@@ -10,42 +16,336 @@ namespace Microsoft.PowerShell.EditorServices.Utility
...
+
+        /// <summary>
+        /// Calculates the start line and column of the actual method/constructor name in a function member AST.
+        /// </summary>
+        /// <param name="functionMemberAst">A FunctionMemberAst object in the script's AST</param>
+        /// <returns>A tuple with start column and line for the method/constructor name</returns>
+        internal static (int startColumn, int startLine) GetNameStartColumnAndLineFromAst(FunctionMemberAst functionMemberAst)
+        {
+            // find name index to get offset even with attributes, static, hidden ++
+            int nameStartIndex = functionMemberAst.Extent.Text.LastIndexOf(string.Concat(functionMemberAst.Name, '('), StringComparison.OrdinalIgnoreCase);
```
SeeminglyScience, line 100: Does `LastIndexOf` make sense here? Does that not include the body of the function? If it does, then recursive functions might cause some errors here. Or even something like:

```powershell
class test {
    [string] GetThing() {
        return 'GetThing('
    }
}
```

Ideally we'd be using tokens for this

src/PowerShellEditorServices/Utility/VisitorUtils.cs
```diff
@@ -10,42 +16,336 @@ namespace Microsoft.PowerShell.EditorServices.Utility
...
+        /// Calculates the start line and column of the actual property name in a property member AST.
+        /// </summary>
+        /// <param name="propertyMemberAst">A PropertyMemberAst object in the script's AST</param>
+        /// <param name="isEnumMember">A bool indicating this is a enum member</param>
+        /// <returns>A tuple with start column and line for the property name</returns>
+        internal static (int startColumn, int startLine) GetNameStartColumnAndLineFromAst(PropertyMemberAst propertyMemberAst, bool isEnumMember)
+        {
+            // find name index to get offset even with attributes, static, hidden ++
+            string searchString = isEnumMember ? propertyMemberAst.Name : string.Concat('$', propertyMemberAst.Name);
+            int nameStartIndex = propertyMemberAst.Extent.Text.LastIndexOf(searchString, StringComparison.OrdinalIgnoreCase);
```
SeeminglyScience, line 114: Same here as `InitialValue` can have any arbitrary text

SeeminglyScience (changes requested):
I'm probably about 2/3's done reviewin' but here's what I got so far. Very exciting changes though! ❤️

src/PowerShellEditorServices/Services/TextDocument/Handlers/SignatureHelpHandler.cs
```diff
@@ -89,7 +89,7 @@ private static ParameterInformation CreateParameterInfo(ParameterInfo parameterI
             return new ParameterInformation
             {
                 Label = parameterInfo.Name,
-                Documentation = string.Empty
+                Documentation = parameterInfo.HelpMessage
```
andschwa, line 92: @SeeminglyScience Not a whole lot comes from this it seems.

src/PowerShellEditorServices/Services/TextDocument/ScriptFile.cs
```diff
@@ -53,6 +52,7 @@ internal sealed class ScriptFile
         /// Gets or sets a boolean that determines whether
         /// semantic analysis should be enabled for this file.
         /// For internal use only.
+        /// TODO: Actually use and respect this property to avoid built-in files from being analyzed.
```
andschwa, line 55: And we stopped opening built-ins, so actually `TODO: We can remove this property.`

src/PowerShellEditorServices/Services/TextDocument/ScriptRegion.cs
andschwa: LOL but who on _earth_ would be using it? 👀

src/PowerShellEditorServices/Utility/Extensions.cs
```diff
@@ -135,7 +135,14 @@ public static bool Contains(this IScriptExtent scriptExtent, int line, int colum
...
+                {
+                    return scriptExtent.StartColumnNumber <= column && scriptExtent.EndColumnNumber >= column;
+                }
+                else
+                {
+                    return scriptExtent.StartColumnNumber <= column;
+                }
             }
 
             if (scriptExtent.EndLineNumber == line)
```
andschwa, line 148: I don't remember changing this, and I don't remember using it either. Possibly dead code? We _do_ still have a whole other thing going on with _another_ `IScriptExtent` implementation that isn't the `ScriptRegion` being used by the symbols' logic.

src/PowerShellEditorServices/Utility/VisitorUtils.cs
andschwa: Good point, any better fix? @fflaten any ideas?

test/PowerShellEditorServices.Test.Shared/Symbols/DSCFile.ps1
```diff
@@ -0,0 +1,4 @@
+﻿# This file represents a script with a DSC configuration
+configuration AConfiguration {
```
andschwa, line 2: @fflaten did you have DSC working? I had to disable the test entirely, the visitor was never getting called. Patrick and I thought it might be a module-import issue.

src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -0,0 +1,245 @@
...
+        // Enum members and properties are the "same" according to PowerShell, so the symbol name's
+        // must be the same since we can't distinguish them in VisitMemberExpression.
+        SymbolType symbolType =
+            propertyMemberAst.Parent is TypeDefinitionAst typeAst && typeAst.IsEnum
+                ? SymbolType.EnumMember
+                : SymbolType.Property;
+
+        return _action(new SymbolReference(
+            symbolType,
+            "var " + propertyMemberAst.Name,
```
fflaten, line 180:
```suggestion
            "prop " + propertyMemberAst.Name,
```

Avoid conflict between properties and variables.

src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -0,0 +1,245 @@
...
+        string? memberName = memberExpressionAst.Member is StringConstantExpressionAst stringConstant ? stringConstant.Value : null;
+        if (string.IsNullOrEmpty(memberName))
+        {
+            return AstVisitAction.Continue;
+        }
+
+        // TODO: It's too bad we can't get the property's real symbol and reuse its display string.
+        return _action(new SymbolReference(
+            SymbolType.Property,
+            "var " + memberName,
```
fflaten, line 199:
```suggestion
            "prop " + memberName,
```

Avoid conflict between properties and variables.

test/PowerShellEditorServices.Test.Shared/Symbols/DSCFile.ps1
fflaten: AFAIK all tests passed (DSC only on Windows, skipped on rest). I know I had problems making it DSC work on Unix even after installing all the dependencies and I develop in devcontainer so might not have tested it manually.

I'm not against skipping it. Mostly cosmetic feature since you usually have one configuration per file (I assume). Isn't the `configuration` keyword also being removed in DSC v3?

src/PowerShellEditorServices/Utility/VisitorUtils.cs
fflaten: Good catch! `IndexOf()`, which I really don't understand why I didn't use originally. 😕 
You're not allowed to use `(` before function name so shouldn't be any conflicts.

src/PowerShellEditorServices/Utility/VisitorUtils.cs
fflaten: Agree. Should've been `IndexOf()` also.

`$` aren't used in neither types or keywords like static, hidden etc. EnumMember can't have anything prefixed AFAIK, so shouldn't be an issue, right?

src/PowerShellEditorServices/Utility/Extensions.cs
fflaten: https://github.com/fflaten/PowerShellEditorServices/commit/f587eededad7b709d15141aca3f0be7ec4b78473

Not dead. IIRC, that's a fix for a false positive this:
```ps1
function demo { 'hello' }; demo
#                          ^
# Looking for this --------|
```

andschwa: But what uses it? I see no references to this method 👀

test/PowerShellEditorServices.Test.Shared/Symbols/DSCFile.ps1
andschwa: No worries and yeah, not super concerned about it right now.

andschwa:
Thanks @fflaten and @SeeminglyScience, resolved those things!

src/PowerShellEditorServices/Utility/Extensions.cs
fflaten: Oh right 🤦‍♂️ I focused on the diff, but the method itself was only used with FindSymbolVisitor. All replaced by `TryGetSymbolAtPosition(line, column)` 👍

andschwa: Ah, thanks!

SeeminglyScience (approved):
LGTM! These changes are awesome. Huge thank you @fflaten (and ofc @andschwa) ❤️

PR #1989, opened 2023-02-07
Keep only first assignment as declaration

andschwa:
As noted by @fflaten in https://github.com/PowerShell/vscode-powershell/issues/1465#issuecomment-1421401557, our initial logic counted _every_ assignment as a declaration. While there's no true way to determine which is _really_ the declaration, we can make a relatively safe assumption that the first encountered declaration is it. This is efficient, easy, and seems to work quite well.

### DIFF
--- a/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/ReferenceTable.cs
@@ -94,6 +94,15 @@ internal sealed class ReferenceTable
             _ => new ConcurrentBag<SymbolReference> { symbol },
             (_, existing) =>
             {
+                // Keep only the first variable encountered as a declaration marked as such. This
+                // keeps the first assignment without also counting every reassignment as a
+                // declaration (cleaning up e.g. Code's outline view).
+                if (symbol.Type is SymbolType.Variable && symbol.IsDeclaration
+                    && existing.Any(i => i.IsDeclaration))
+                {
+                    symbol = symbol with { IsDeclaration = false };
+                }
+
                 existing.Add(symbol);
                 return existing;
             });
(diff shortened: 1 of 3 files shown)


SeeminglyScience (approved):
I'm cool with this since there isn't really a good solution without some sort of flow analysis work. Just want to put it on record that we know this will miss **a lot** of stuff in large files, but that we think it'll be better to do that than to list every single assignment.

For example, folks building one large psm1 with all their functions, if they reuse variable names we'll miss all but the first one.

(ofc please correct me if I'm wrong)

andschwa:
Also quoting a great example from @fflaten in https://github.com/PowerShell/vscode-powershell/issues/1465#issuecomment-1421446054:

> I believe that would make the inner assignment the definition in the example below.
> 
> ```ps1
> Function abc {
>     $myvar = 'inner'
> }
> 
> $myvar = 'outer'
> ```
> 
> Maybe track the first assignment per parent (when that is tracked)?

He's absolutely right that's going to happen, but we can't improve it until we have parent/child relationship tracking (and even then it might prove tricky).

andschwa:
I think the most annoying bit will be that for a commonly re-used symbol name like `$i`, if you jump to definition on a local use of it, instead of showing you all the potential definitions, it'll just take you to the very first one.

That one has me thinking that maybe we should just filter _for outline view_?

fflaten:
> Just want to put it on record that we know this will miss **a lot** of stuff in large files, but that we think it'll be better to do that than to list every single assignment.

This 👍. Before this PR, open ex. `Pester.psm1` and get overwhelmed by variables in outline and go to symbol.

> That one has me thinking that maybe we should just filter _for outline view_? 

Yes. I could see a value in listing all possible definitions, but outline and symbol search get easily crowded.

Something to revisit later :-)

PR #1990, opened 2023-02-07
Strip scope from function references

andschwa:
Like methods we just bucket them as overloads, which means that it works better than before despite not being perfect.

Resolves https://github.com/PowerShell/vscode-powershell/issues/1089.


SeeminglyScience (approved):
LGTM!

PR #1991, opened 2023-02-08
Declare scope prefixes array as static readonly

andschwa:
I forgot that C# won't just optimize away static arrays.


SeeminglyScience (approved):
LGTM!

PR #1992, opened 2023-02-09
WIP: Support multi-root workspace

andyleejordan:
This time in the server. Resolves https://github.com/PowerShell/vscode-powershell/issues/2112 but needs tests and some more work.

### DIFF
--- a/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
+++ b/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
@@ -306,18 +306,23 @@ namespace Microsoft.PowerShell.EditorServices.Services
         /// <summary>
         /// Gets the workspace-relative path of the given file path.
         /// </summary>
-        /// <param name="filePath">The original full file path.</param>
         /// <returns>A relative file path</returns>
-        public string GetRelativePath(string filePath)
+        public string GetRelativePath(ScriptFile scriptFile)
         {
-            string resolvedPath = filePath;
+            string resolvedPath = scriptFile.FilePath;
 
-            if (!IsPathInMemory(filePath) && !string.IsNullOrEmpty(InitialWorkingDirectory))
+            if (!scriptFile.IsInMemory)
             {
-                Uri workspaceUri = new(InitialWorkingDirectory);
-                Uri fileUri = new(filePath);
-
-                resolvedPath = workspaceUri.MakeRelativeUri(fileUri).ToString();
+                Uri fileUri = scriptFile.DocumentUri.ToUri();
+                foreach (WorkspaceFolder workspaceFolder in WorkspaceFolders)
+                {
+                    Uri workspaceUri = workspaceFolder.Uri.ToUri();
+                    if (workspaceUri.IsBaseOf(fileUri))
+                    {
+                        resolvedPath = workspaceUri.MakeRelativeUri(fileUri).ToString();
+                        break;
+                    }
+                }
 
                 // Convert the directory separators if necessary
                 if (Path.DirectorySeparatorChar == '\\')
--- a/src/PowerShellEditorServices/Extensions/FileContext.cs
+++ b/src/PowerShellEditorServices/Extensions/FileContext.cs
@@ -58,8 +58,7 @@ namespace Microsoft.PowerShell.EditorServices.Extensions
         /// <summary>
         /// Gets the workspace-relative path of the file.
         /// </summary>
-        public string WorkspacePath => editorOperations.GetWorkspaceRelativePath(
-                        scriptFile.FilePath);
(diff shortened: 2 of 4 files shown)


src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -350,20 +362,29 @@ bool ignoreReparsePoints
...
+                    logger);
+
+                PatternMatchingResult fileMatchResult = matcher.Execute(fsFactory.RootDirectory);
+                foreach (FilePatternMatch item in fileMatchResult.Files)
+                {
+                    // item.Path always contains forward slashes in paths when it should be backslashes on Windows.
+                    // Since we're returning strings here, it's important to use the correct directory separator.
+                    string path = VersionUtils.IsWindows ? item.Path.Replace('/', Path.DirectorySeparatorChar) : item.Path;
+                    yield return Path.Combine(rootPath, path);
+                }
```
andschwa, line 387: This worked in my testing of a multi-root workspace, now PowerShell symbols from both (and so presumably all) folders showed up!

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -293,18 +300,23 @@ public void CloseFile(ScriptFile scriptFile)
...
-        /// <param name="filePath">The original full file path.</param>
         /// <returns>A relative file path</returns>
-        public string GetRelativePath(string filePath)
+        public string GetRelativePath(ScriptFile scriptFile)
         {
-            string resolvedPath = filePath;
+            string resolvedPath = scriptFile.FilePath;
 
-            if (!IsPathInMemory(filePath) && !string.IsNullOrEmpty(WorkspacePath))
+            if (!scriptFile.IsInMemory)
```
andschwa, line 308: I think `IsPathInMemory` can go away because it's only used in two tests now, and `IsInMemory` is used everywhere else. Not sure why we had two ways to go about that.

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -293,18 +300,23 @@ public void CloseFile(ScriptFile scriptFile)
...
-                resolvedPath = workspaceUri.MakeRelativeUri(fileUri).ToString();
+                Uri fileUri = scriptFile.DocumentUri.ToUri();
+                foreach (WorkspaceFolder workspaceFolder in WorkspaceFolders)
+                {
+                    Uri workspaceUri = workspaceFolder.Uri.ToUri();
+                    if (workspaceUri.IsBaseOf(fileUri))
+                    {
+                        resolvedPath = workspaceUri.MakeRelativeUri(fileUri).ToString();
+                        break;
+                    }
```
andschwa, line 318: I think this works 😅

src/PowerShellEditorServices/Server/PsesLanguageServer.cs
```diff
@@ -149,13 +144,18 @@ public async Task StartAsync()
...
+                                InitialWorkingDirectory = initializationOptions?.GetValue("initialWorkingDirectory")?.Value<string>()
+                                    ?? workspaceService.WorkspaceFolders.FirstOrDefault()?.Uri.GetFileSystemPath(),
+                                ShellIntegrationEnabled = initializationOptions?.GetValue("shellIntegrationEnabled")?.Value<bool>()
+                                    ?? false
                             };
 
+                            // NOTE: We're still maintaining a root path while things transition to
+                            // support multi-root workspaces.
+                            workspaceService.WorkspacePath = hostStartOptions.InitialWorkingDirectory;
+
```
andschwa, line 158: The main thing we still use this for is resolving the path to the PSScriptAnalyzer settings file. People give relative paths...but we have no way of asking "ok, but relative to which root?" and so for now the logic is "your initial working directory setting" first (so it can be corrected if necessary) and then fall back to the "first" (but presumably random sorting) workspace folder.

src/PowerShellEditorServices/Extensions/EditorWorkspace.cs
```diff
@@ -19,6 +19,7 @@ public sealed class EditorWorkspace
 
         /// <summary>
         /// Gets the current workspace path if there is one or null otherwise.
+        /// TODO: This API needs to be deprecated for new `Paths` instead.
         /// </summary>
         public string Path => editorOperations.GetWorkspacePath();
```
andschwa, line 24: Why oh why did we ever write a public API 💁

andschwa:
Whacha think @SeeminglyScience?

andyleejordan:
Closing in favor of https://github.com/PowerShell/PowerShellEditorServices/pull/2053.

PR #1994, opened 2023-02-10
Update fabricbot.json

SydneyhSmith:
# PR Summary

Updating fabricbot to not comment when a collaborator takes action

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1244,6 +1244,17 @@
       "version": "1.0",
       "config": {
         "taskName": "Clean emails"
+            {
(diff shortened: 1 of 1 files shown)


andschwa:
I think we need to look into this further because turns out we have this exact condition applied to the rule that adds "hey don't comment on old issues!" but it's not working. I'm guessing we're not counted as collaborators? Not sure what we are then.

https://github.com/PowerShell/PowerShellEditorServices/blob/5228a59fcc60fa58e6e9bc00a54171cee472b97c/.github/fabricbot.json#L1211-L1228

PR #1995, opened 2023-02-11
Add `WorkspaceFolders` and use it when enumerating files

andschwa:
Resolves https://github.com/PowerShell/vscode-powershell/issues/2112. Should probably write a test for it, does _not_ yet include updates to the API to handle getting the correct workspace for an open editor.


SeeminglyScience (approved):
LGTM!

PR #1998, opened 2023-02-20
Support module-qualified calls for Pester keywords

fflaten:
# PR Summary

Support Pester features when using module-qualified commands.

## PR Context

Fix PowerShell/vscode-powershell#928

### DIFF
--- a/src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
@@ -62,7 +62,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
             // Ensure the first word is a Pester keyword and in Pester-module if using module-qualified call
             string commandName = CommandHelpers.StripModuleQualification(commandAst.GetCommandName(), out ReadOnlyMemory<char> module);
             if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandName) ||
-                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase))
+                (!module.IsEmpty && !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase)))
             {
                 return false;
             }
--- a/src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
@@ -62,7 +62,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
             // Ensure the first word is a Pester keyword and in Pester-module if using module-qualified call
             string commandName = CommandHelpers.StripModuleQualification(commandAst.GetCommandName(), out ReadOnlyMemory<char> module);
             if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandName) ||
-                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase)
+                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase))
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
```diff
@@ -59,8 +59,10 @@ private static bool IsPesterCommand(CommandAst commandAst)
                 return false;
             }
 
-            // Ensure the first word is a Pester keyword
-            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandAst.GetCommandName()))
+            // Ensure the first word is a Pester keyword and in Pester-module if using module-qualified call
+            string commandName = CommandHelpers.StripModuleQualification(commandAst.GetCommandName(), out ReadOnlyMemory<char> module);
+            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandName) ||
+                (!module.IsEmpty && !string.Equals(module.ToString(), "pester", StringComparison.OrdinalIgnoreCase)))
```
fflaten, line 65: More efficient way to compare the module? Unfamiliar with the Memory/Span types

fflaten:
Had to test if this was even supported by Pester. Never seen anyone use qualified names before I found this old issue 😄

src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
SeeminglyScience: Yep!

```suggestion
                module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase)
```

andschwa: I asked ChatGPT and while this is the most straightforward way to do it, it can be done more efficiently like so:
```c#
module.Span.SequenceEqual("pester".AsSpan())
```
And I guess the `"pester"` string could be made as a span itself.

As to _why_ that `out` is a ROM<char> it was because:

> Storing moduleName as ROMemory saves a string allocation in the common case where it is not needed.

/cc @SeeminglyScience

andschwa, line 65:
```suggestion
                !module.Span.SequenceEqual("pester".AsSpan()))
```

andschwa (approved):
Great fix, thanks @fflaten! I think the code I suggested should work...

src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
```diff
@@ -59,8 +59,10 @@ private static bool IsPesterCommand(CommandAst commandAst)
                 return false;
             }
 
-            // Ensure the first word is a Pester keyword
-            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandAst.GetCommandName()))
+            // Ensure the first word is a Pester keyword and in Pester-module if using module-qualified call
+            string commandName = CommandHelpers.StripModuleQualification(commandAst.GetCommandName(), out ReadOnlyMemory<char> module);
+            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandName) ||
+                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase)
```
fflaten, line 65:
```suggestion
                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase))
```

src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
fflaten: Thanks! 🙂

andschwa:
Oops we totally broke it.

src/PowerShellEditorServices/Services/Symbols/PesterDocumentSymbolProvider.cs
```diff
@@ -59,8 +59,10 @@ private static bool IsPesterCommand(CommandAst commandAst)
                 return false;
             }
 
-            // Ensure the first word is a Pester keyword
-            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandAst.GetCommandName()))
+            // Ensure the first word is a Pester keyword and in Pester-module if using module-qualified call
+            string commandName = CommandHelpers.StripModuleQualification(commandAst.GetCommandName(), out ReadOnlyMemory<char> module);
+            if (!PesterSymbolReference.PesterKeywords.ContainsKey(commandName) ||
+                !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase))
```
andschwa, line 65:
```suggestion
                (!module.IsEmpty && !module.Span.Equals("pester".AsSpan(), StringComparison.OrdinalIgnoreCase)))
```

We are SO BAD @SeeminglyScience. The `IsEmpty` is because it might _not_ be module qualified, it wasn't a guard!

fflaten: I guess we all learned something today 👏🥳

PR #2003, opened 2023-02-24
Add document symbols for `#region`

andschwa:
And some technical debt clean-ups.

Replaces (and re-uses!) @fflaten's https://github.com/PowerShell/PowerShellEditorServices/pull/1911.

Resolves https://github.com/PowerShell/vscode-powershell/issues/3604 and https://github.com/PowerShell/vscode-powershell/issues/3210.

### DIFF
--- a/src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
+++ b/src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
@@ -123,6 +123,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Analysis
 
         /// <summary>
         /// Format a script given its contents.
+        /// TODO: This needs to be cancellable.
         /// </summary>
         /// <param name="scriptDefinition">The full text of a script.</param>
         /// <param name="formatSettings">The formatter settings to use.</param>
--- a/src/PowerShellEditorServices/Services/Symbols/SymbolDetails.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/SymbolDetails.cs
@@ -34,6 +34,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
 
         #region Constructors
 
+        // TODO: This should take a cancellation token!
         internal static async Task<SymbolDetails> CreateAsync(
             SymbolReference symbolReference,
             IRunspaceInfo currentRunspace,
--- a/src/PowerShellEditorServices/Services/Symbols/SymbolDetails.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/SymbolDetails.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Management.Automation;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerShell.EditorServices.Services.PowerShell;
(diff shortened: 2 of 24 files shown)


daxian-dbw (approved):
LGTM

src/PowerShellEditorServices/Services/TextDocument/Handlers/CodeLensHandlers.cs
```diff
@@ -40,12 +39,17 @@ protected override CodeLensRegistrationOptions CreateRegistrationOptions(CodeLen
...
         public override Task<CodeLensContainer> Handle(CodeLensParams request, CancellationToken cancellationToken)
         {
+            _logger.LogDebug($"Handling code lens request for {request.TextDocument.Uri}");
+
             ScriptFile scriptFile = _workspaceService.GetFile(request.TextDocument.Uri);
-            CodeLens[] codeLensResults = ProvideCodeLenses(scriptFile, cancellationToken);
-            return Task.FromResult(new CodeLensContainer(codeLensResults));
+            IEnumerable<CodeLens> codeLensResults = ProvideCodeLenses(scriptFile, cancellationToken);
+
+            return !codeLensResults.Any()
```
SeeminglyScience, line 47: `ProvideCodeLenses` returns a yield enumerator right? If so then the call to `Any` will double enumerate the first item at least (meaning it'll double invoke the method basically). Best to just pass the enumerator along, not worth the potential allocation save

Side note, we should probably look at a way to make this actually async. I know this is already the case, but the method being entirely synchronous while pretending to be async is a little concerning

src/PowerShellEditorServices/Services/TextDocument/Handlers/DocumentHighlightHandler.cs
```diff
@@ -45,7 +46,7 @@ protected override DocumentHighlightRegistrationOptions CreateRegistrationOption
                 request.Position.Line + 1,
                 request.Position.Character + 1);
 
-            if (occurrences is null)
+            if (!occurrences.Any())
```
SeeminglyScience, line 49: I'd examine any use of `Any` to make sure it's not being called on a yield enumerator. Double enumeration can be fine for entirely "pure" yield methods but it's a foot gun I'd like to avoid if we can

SeeminglyScience (changes requested):
Lookin' good, just one thing we need to change up

src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
```diff
@@ -123,6 +123,7 @@ public PssaCmdletAnalysisEngine Build(string pssaModulePath)
 
         /// <summary>
         /// Format a script given its contents.
+        /// TODO: This needs to be cancellable.
```
andschwa, line 126: This one is a big ol' "waiting on PSSA" right?

src/PowerShellEditorServices/Services/Symbols/SymbolDetails.cs
```diff
@@ -34,6 +34,7 @@ internal class SymbolDetails
 
         #region Constructors
 
+        // TODO: This should take a cancellation token!
```
andschwa, line 37: Whereas this one I think could actually be done and I meant to add a commit to do it.

SeeminglyScience (approved):
LGTM!

PR #2006, opened 2023-03-09
Fix declaration detection for variables with type constraints

andschwa:
Resolves https://github.com/PowerShell/vscode-powershell/issues/4443.

### DIFF
--- a/src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
@@ -110,6 +110,27 @@ internal sealed class SymbolVisitor : AstVisitor2
             ));
         }
 
+        // Traverse the parents to determine if this is a declaration.
+        static bool isDeclaration(Ast current)
+        {
+            Ast next = current.Parent;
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices/Services/Symbols/Visitors/SymbolVisitor.cs
```diff
@@ -110,6 +110,27 @@ public override AstVisitAction VisitVariableExpression(VariableExpressionAst var
             ));
         }
 
+        // Traverse the parents to determine if this is a declaration.
+        static bool isDeclaration(Ast current)
+        {
+            Ast next = current.Parent;
+            while (true)
+            {
+                // Should come from an assignment statement.
```
SeeminglyScience, line 119:
```suggestion
                if (next is ParameterAst parameter && parameter.DefaultValue != current)
                {
                    return true;
                }

                // Should come from an assignment statement.
```

SeeminglyScience (changes requested):
Just one change to check for parameters as well (unless you already have this and I missed it)

SeeminglyScience (approved):
Nvm the suggestion, it works! LGTM!

PR #2008, opened 2023-03-21
Update build documentation to include platyPS requirement

andschwa:
Since it's always used post-build to generate the documentation, we have a `#requires` for it but forgot to add it to the documentation.

Also suppress stderr for sysctl when checking if macOS is an Intel or Arm processor. On the former, it doesn't know what field we're querying.

### DIFF
--- a/PowerShellEditorServices.build.ps1
+++ b/PowerShellEditorServices.build.ps1
@@ -42,7 +42,8 @@ $script:dotnetTestArgs = @("test") + $script:dotnetBuildArgs + $TestArgs + @(
 $script:IsNix = $IsLinux -or $IsMacOS
 # For Apple M1, pwsh might be getting emulated, in which case we need to check
 # for the proc_translated flag, otherwise we can check the architecture.
-$script:IsAppleM1 = $IsMacOS -and ((sysctl -n sysctl.proc_translated) -eq 1 -or (uname -m) -eq "arm64")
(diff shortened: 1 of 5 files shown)


PowerShellEditorServices.build.ps1
```diff
@@ -42,7 +42,7 @@ $script:dotnetTestArgs = @("test") + $script:dotnetBuildArgs + $TestArgs + @(
 $script:IsNix = $IsLinux -or $IsMacOS
 # For Apple M1, pwsh might be getting emulated, in which case we need to check
 # for the proc_translated flag, otherwise we can check the architecture.
-$script:IsAppleM1 = $IsMacOS -and ((sysctl -n sysctl.proc_translated) -eq 1 -or (uname -m) -eq "arm64")
+$script:IsAppleM1 = $IsMacOS -and ((sysctl -n sysctl.proc_translated 2> $null) -eq 1 -or (uname -m) -eq "arm64")
```
SteveL-MSFT, line 45: May be better to use `[system.runtime.InteropServices.RuntimeInformation]::ProcessArchitecture`

SteveL-MSFT (changes requested):
Use .NET API instead of calling executable and parsing output

PowerShellEditorServices.build.ps1
andschwa: Thanks!

SteveL-MSFT (approved):

PR #2018, opened 2023-04-24
Set session failure with reason when applicable

andschwa:
This updates our session failure logic (we were no longer ever sending a "failed" status back, hence the awful experience in the client). If PowerShell does start, but detects that it's not going to work due to unsupported PowerShell or .NET Framework versions (applicable only to Windows PowerShell), we set the session status to failed and actually report the PowerShell version for use by the client's error reporting.

Note that as far as we know, our minimum .NET Framework support for Windows PowerShell is now 4.8, per https://github.com/PowerShell/vscode-powershell/issues/4483.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4509.


SeeminglyScience (approved):
LGTM!

PR #2020, opened 2023-05-05
Fix repeated failure to load DSC module

andschwa:
While debugging the `Write-Output` bug (hence the branch name) I came across another bug where the server was attempting to load the DSC module repeatedly. The `SetLineBreakpoint` logic queries for DSC breakpoint capability, which was importing the module, catching the failure to import as an error, and then trying again every single time. Instead we should import (or fail to import) only the first time we try. Also, we don't need to transform the `BreakpointDetails` list into an array and back, just leave it as `IEnumerable`.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
@@ -22,24 +22,19 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Debugging
     {
         private string[] dscResourceRootPaths = Array.Empty<string>();
 
-        private readonly Dictionary<string, int[]> breakpointsPerFile =
-            new();
+        private readonly Dictionary<string, int[]> breakpointsPerFile = new();
 
-        public async Task<BreakpointDetails[]> SetLineBreakpointsAsync(
+        public async Task<IEnumerable<BreakpointDetails>> SetLineBreakpointsAsync(
             IInternalPowerShellExecutionService executionService,
             string scriptPath,
-            BreakpointDetails[] breakpoints)
+            IEnumerable<BreakpointDetails> breakpoints)
         {
-            List<BreakpointDetails> resultBreakpointDetails =
-                new();
-
             // We always get the latest array of breakpoint line numbers
             // so store that for future use
-            if (breakpoints.Length > 0)
+            if (breakpoints.Any())
             {
                 // Set the breakpoints for this scriptPath
-                breakpointsPerFile[scriptPath] =
-                    breakpoints.Select(b => b.LineNumber).ToArray();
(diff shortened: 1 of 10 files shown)


src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
```diff
@@ -1,45 +1,37 @@
...
-            BreakpointDetails[] breakpoints)
+            IEnumerable<BreakpointDetails> breakpoints)
         {
-            List<BreakpointDetails> resultBreakpointDetails =
-                new();
-
             // We always get the latest array of breakpoint line numbers
             // so store that for future use
-            if (breakpoints.Length > 0)
+            if (breakpoints.Any())
```
SeeminglyScience, line 31: So I don't think it's particularly impactful here but it's worth noting that double enumeration will happen for the first item. Granted the only code being executed twice is `BreakpointDetails.Create` but it adds more surface for easy bugs as if at some point `Create` is no longer "pure" then we'd get unintended side effects.

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -264,8 +264,8 @@ public async Task DebuggerSetsAndClearsFunctionBreakpoints()
                 }).ConfigureAwait(true);
 
             Assert.Equal(2, breakpoints.Length);
-            Assert.Equal("Write-Host", breakpoints[0].Name);
-            Assert.Equal("Get-Date", breakpoints[1].Name);
+            Assert.Equal("Write-Host", breakpoints.ElementAt(0).Name);
```
SeeminglyScience, line 267: Same double enumeration warning here, this line and the line below will invoke the producing code twice first the first element

(I know this is just tests and doesn't necessarily impact anything, just worth pointing out because the behavior is not exactly intuitive)

SeeminglyScience (approved):
LGTM

I'd personally remove the enumerable change for maintainability but I'll leave that up to your discretion

src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
andschwa: How do you avoid that?

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
andschwa: What's the right way to do it?

SeeminglyScience: I would say `ToArray` it first is the most common answer. You could `foreach` it and save the first and second instance, but that gets a lot more complicated the more you want to do with it. Like further on when you do `Count()` that'll enumerate everything to get the count, then when you do `ElementAt(0).LineNumber` it'll run `Create` again, and again when you do `ElementAt(0).Verified`, etc

PR #2022, opened 2023-05-10
Only get aliases when looking at function symbols

andschwa:
Instead of every symbol reference lookup. Though we should probably still do this less often, like in the `prompt`.


andschwa:
@SeeminglyScience it's _an_ improvement.

PR #2023, opened 2023-05-12
Fix disappearing output in PowerShell 5.1

andschwa:
What a bug! In PowerShell 5.1 our artificial pipeline for event handling (among other possible PowerShell tasks) could cause the output to just disappear. Turns out a [bug in 5.1](https://github.com/PowerShell/PowerShell/pull/3436) with the transcript was being triggered. Scripts without `Out-Default` would get it appended with `-TranscribeOnly True` in order for the transcription to happen...but the `TranscribeOnly` field could erroneously not be set back to `false`! This would cause all output to just "disappear." It was fixed in PowerShell 7, but we still have to work around it for PowerShell 5.1 by resetting it.

Resolves https://github.com/PowerShell/vscode-powershell/issues/3991, and fixes a few other things too.

Unfortunately due to the implementation of PowerShell 5.1's transcript, we still get a flood of `[System.Diagnostics.DebuggerHidden()]param() 0` in the transcript sometimes. It's annoying, and can lead to massive transcript files, but I don't believe it's the root cause. We're still looking for a way to fix that.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
+++ b/src/PowerShellEditorServices.Hosting/EditorServicesLoader.cs
@@ -350,7 +350,9 @@ namespace Microsoft.PowerShell.EditorServices.Hosting
         private static string GetPSOutputEncoding()
         {
             using SMA.PowerShell pwsh = SMA.PowerShell.Create();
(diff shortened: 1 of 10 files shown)


SeeminglyScience (approved):
LGTM!

andschwa:
I half want to make this a bunch of PRs because it's quite a few fixes, oh well!

PR #2031, opened 2023-05-30
Fix the `TranscribeOnly` bug (take three)

andschwa:
Yet again fixing compiler constants, just go this plain wrong last time. Also caught in testing that we need to disable `TranscribeOnly` in the `finally` of `ExecuteDebugger` too, so that the original `Out-Default` (from the invocation of the debugged script) isn't broken.


SeeminglyScience (approved):
LGTM!

Issue #2050, opened 2023-08-09
Shell integration with PowerShell 5.1 breaks with `$ErrorActionPreference = Stop` and `Set-StrictMode`


Pxtl:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

In `$PROFILE.CurrentUserAllHosts`, set the content to 

```
$ErrorActionPreference = 'stop'
Set-StrictMode -Version Latest
```

then start VSCode.

### Expected behavior

```console
The Powershell extension starts successfully.
```

### Actual behavior

```console
The VSCode Powershell extension fails to start.
```

### Error details

```console
During startup:

2023-08-09 17:03:50.695 -04:00 [VRB] Assembly resolve event fired for System.Management.Automation.resources, Version=3.0.0.0, Culture=en-US, PublicKeyToken=31bf3856ad364e35
2023-08-09 17:03:50.698 -04:00 [WRN] Runtime exception occurred while executing command:

System.Management.Automation.RuntimeException: The variable '$IsWindows' cannot be retrieved because it has not been set.
   at System.Management.Automation.Runspaces.PipelineBase.Invoke(IEnumerable input)
   at System.Management.Automation.PowerShell.Worker.ConstructPipelineAndDoWork(Runspace rs, Boolean performSyncInvoke)
   at System.Management.Automation.PowerShell.Worker.CreateRunspaceIfNeededAndDoWork(Runspace rsToUse, Boolean isSync)
   at System.Management.Automation.PowerShell.CoreInvokeHelper[TInput,TOutput](PSDataCollection`1 input, PSDataCollection`1 output, PSInvocationSettings settings)
   at System.Management.Automation.PowerShell.CoreInvoke[TInput,TOutput](PSDataCollection`1 input, PSDataCollection`1 output, PSInvocationSettings settings)
   at System.Management.Automation.PowerShell.Invoke[T](IEnumerable input, PSInvocationSettings settings)
   at Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility.PowerShellExtensions.InvokeAndClear[TResult](PowerShell pwsh, PSInvocationSettings invocationSettings)
   at Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution.SynchronousPowerShellTask`1.ExecuteNormally(CancellationToken cancellationToken)
[...]

PR #2052, opened 2023-08-11
Fix shell integration for PowerShell 5.1 with strict mode

andyleejordan:
Since `$IsWindows` doesn't exist in Windows PowerShell, with strict mode enabled and the `ErrorActionPreference` set to `Stop`, shell integration would cause the startup to crash. This was reported and fixed upstream too.

Resolves https://github.com/PowerShell/PowerShellEditorServices/issues/2050.


andyleejordan:
Same change as integrated upstream in https://github.com/microsoft/vscode/pull/190192.

SydneyhSmith (approved):

PR #2064, opened 2023-09-08
Fix debugging script blocks that aren't in files

andyleejordan:
The code relied on running `list 1 <MaxInt>` but that wasn't being run in the debug context because we allow-list commands to prevent pollution of the history, and missed it. Like the `prompt` and interactive commands (which `list` could be but is not when we run it) we need to check for this exact `list` command and run it under the debugger.

Moreover, we also weren't locking the `debugInfoHandle`, nor were we correctly checking if `scriptListingLines` was empty (it was never null), and our shortcut to skip allocation was broken. Actually we can't skip allocation, but we can at least skip superfluous conversions.

Resolves https://github.com/PowerShell/vscode-powershell/issues/3904.


SeeminglyScience (approved):
LGTM!

PR #2068, opened 2023-09-15
Wrap import of DSC module with `ProgressPreference = SilentlyContinue`

andyleejordan:
And set it back afterwards. Since it has `ErrorAction = Ignore` it should always get set back correctly. This prevents the progress pane from getting stuck in certain circumstances.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
@@ -92,10 +92,13 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Debugging
             if (!isDscInstalled.HasValue)
             {
                 PSCommand psCommand = new PSCommand()
+                    .AddScript($"$global:{DebugService.PsesGlobalVariableNamePrefix}prevProgressPreference = $ProgressPreference")
+                    .AddScript("$ProgressPreference = 'SilentlyContinue'")
                     .AddCommand(@"Microsoft.PowerShell.Core\Import-Module")
-                    .AddParameter("-Name", "PSDesiredStateConfiguration")
+                    .AddParameter("Name", "PSDesiredStateConfiguration")
                     .AddParameter("PassThru")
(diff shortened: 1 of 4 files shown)


src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
```diff
@@ -92,10 +92,13 @@ public bool IsDscResourcePath(string scriptPath)
             if (!isDscInstalled.HasValue)
             {
                 PSCommand psCommand = new PSCommand()
+                    .AddScript("$prevProgressPreference = $ProgressPreference")
```
SeeminglyScience, line 95:
```suggestion
                    .AddScript("${++prevProgressPreference} = $ProgressPreference")
```

Since this is will be ran in global scope, `$prevProgressPreference` will show up in tab completion and may conflict with a user variable. This one can still show up, but less likely (and won't conflict)

src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
```diff
@@ -92,10 +92,13 @@ public bool IsDscResourcePath(string scriptPath)
...
+                    .AddScript("$ProgressPreference = 'SilentlyContinue'")
                     .AddCommand(@"Microsoft.PowerShell.Core\Import-Module")
-                    .AddParameter("-Name", "PSDesiredStateConfiguration")
+                    .AddParameter("Name", "PSDesiredStateConfiguration")
                     .AddParameter("PassThru")
-                    .AddParameter("ErrorAction", ActionPreference.Ignore);
+                    .AddParameter("ErrorAction", ActionPreference.Ignore)
+                    .AddScript("$ProgressPreference = $prevProgressPreference");
```
SeeminglyScience, line 101:
```suggestion
                    .AddScript("$ProgressPreference = ${++prevProgressPreference}");
```

SeeminglyScience (changes requested):
Just one change real quick

SeeminglyScience (approved):
LGTM!

PR #2079, opened 2023-09-18
Update OmniSharp

andyleejordan:
Dependabot got confused when I updated the configuration and so it accidentally closed #2074 and then opened #2077, but this has the commits which gets the project building again with this update.


andyleejordan:
Replaced with https://github.com/PowerShell/PowerShellEditorServices/pull/2083

PR #2081, opened 2023-09-20
Silence progress output of `Get-DscResource` (take two)

andyleejordan:
It wasn't the module import that was being noisy but the next command. It sure looked like the last commit "fixed" it in testing because the output was certainly silenced...but that was actually due to the side effect of having added an `AddScript` to the `PSCommand` causing it to no longer return the imported module at all. Without that return value, the noisy code wasn't executing either. Now the progress preference is saved, set to silent, `Get-DscResource` is run (silently) and then the preference is restored. Double-checked in the debugger.


andyleejordan:
This is a redo of #2068 which didn't quite work as expected.

SeeminglyScience (approved):
LGTM!

PR #2090, opened 2023-10-18
Ignore `didOpen` notifications for `git` schemed documents from VS Code

andyleejordan:
For some reason VS Code, in some repositories and configurations, will send us `DidOpenTextDocument` notifications for special documents with the `git` scheme (not `file` or `untitled`). It's probably some internal representation that VS Code is using to supply VCS information. Because we were specifically ignoring the URI in this handler these files were getting erroneously added to our workspace service's list of open files, which then caused duplicate references.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4784.


silverqx:
This should solve the problem? Please let me know how I can test it or when there will be some preview release or something. Thx

SeeminglyScience (approved):
LGTM!

andyleejordan:
> This should solve the problem? Please let me know how I can test it or when there will be some preview release or something. Thx

It should solve the problem based on how I was able to reproduce it. The fix will first be available in the next pre-release, so please try that out when it comes out and let us know!

silverqx:
Ok thx I will wait for the new release.

PR #2091, opened 2023-10-19
Run `SetInitialWorkingDirectoryAsync` before `LoadProfiles`

andyleejordan:
So that if a user runs `Set-Location` in their profile, it's not later overridden.

Fixes https://github.com/PowerShell/vscode-powershell/issues/4788


SeeminglyScience (approved):
LGTM!

PR #2096, opened 2023-11-08
Allow `WorkspacePaths` to be empty if we're not in a workspace

andyleejordan:
We were previously adding the initial working directory as a fallback if there was no workspace. However, this led to new VS Code windows (not in a workspace) enumerating all files in the home directory (if that's what the client had to default to, without a workspace folder). And this could even spam permission requests on more secure systems such as macOS. So we just need to let be an empty enumerable.

This may resolve https://github.com/PowerShell/vscode-powershell/issues/4814 and https://github.com/PowerShell/vscode-powershell/issues/4821. Bug was introduced in https://github.com/PowerShell/PowerShellEditorServices/pull/1995.


SeeminglyScience (approved):
LGTM!

PR #2097, opened 2023-11-10
When the built-in `$null` was watched its value was incorrect

andyleejordan:
No idea _why_ its value was `""`, but it was, so we special case it and return `$null` just like you get from watched variables that were assigned the built-in `$null`.

Resolves https://github.com/PowerShell/vscode-powershell/issues/4816.


andyleejordan:
image

Tada!

andyleejordan:
Hm, I hadn't tried `$VarThatDoesNotExist` and it wasn't in the issue, so probably not. I can take a look. Anything else you know about?

PR #2110, opened 2023-12-08
Switch to GitHub Actions for all CI

andyleejordan:
~~Does it work?~~ It works great!


src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
```diff
@@ -360,6 +361,14 @@ private static RunspacePool CreatePssaRunspacePool(string pssaModulePath)
             // We intentionally use `CreateDefault2()` as it loads `Microsoft.PowerShell.Core`
             // only, which is a more minimal and therefore safer state.
             InitialSessionState sessionState = InitialSessionState.CreateDefault2();
+
+            // We set the runspace's execution policy `Bypass` so we can always import our bundled
+            // PSScriptAnalyzer module.
+            if (VersionUtils.IsWindows)
+            {
+                sessionState.ExecutionPolicy = ExecutionPolicy.Bypass;
```
andyleejordan, line 369: Hey @dkattan I bet this was at least _a_ source of those irregular failures. We consistently set this elsewhere for the extension on the CLI (so its inherited to the whole process) or for tests manually in the test host (which these tests weren't using) _but_ in your hosted case I bet it wasn't getting set here either and so failing to load the module.

SeeminglyScience (approved):
LGTM!

andyleejordan:
Naming things is hard.

PR #2115, opened 2023-12-15
Added null check to `GetCompletionsAsync`

dkattan:
# PR Summary

Occasionally TabExpansion2 may return null. This could be due to a custom implementation of TabExpansion2 or an underlying issue with the FileSystemProvider or LocationGlobber. 

When this happens the value of commandCompletion is not assigned and remains null. This causes a NullReferenceException to be thrown when commandCompletion.ReplacementLength is accessed in the subsequent call to LogTrace.

### DIFF
--- a/src/PowerShellEditorServices/Services/Symbols/Visitors/AstOperations.cs
+++ b/src/PowerShellEditorServices/Services/Symbols/Visitors/AstOperations.cs
@@ -76,13 +76,10 @@ namespace Microsoft.PowerShell.EditorServices.Services.Symbols
             CancellationToken cancellationToken)
         {
             IScriptPosition cursorPosition = s_clonePositionWithNewOffset(scriptAst.Extent.StartScriptPosition, fileOffset);
-
+            Stopwatch stopwatch = new();
             logger.LogTrace($"Getting completions at offset {fileOffset} (line: {cursorPosition.LineNumber}, column: {cursorPosition.ColumnNumber})");
 
-            Stopwatch stopwatch = new();
-
-            CommandCompletion commandCompletion = null;
-            await executionService.ExecuteDelegateAsync(
+            CommandCompletion commandCompletion = await executionService.ExecuteDelegateAsync(
                 representation: "CompleteInput",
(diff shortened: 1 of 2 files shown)


src/PowerShellEditorServices/Services/Symbols/Visitors/AstOperations.cs
```diff
@@ -136,7 +140,7 @@ static AstOperations()
                         commandCompletion.ReplacementLength)
                     : null,
                 commandCompletion.CompletionMatches.Count);
-
+            }
             return commandCompletion;
```
andyleejordan, line 144: Since it now returns when not null, the `stopwatch` is superfluous and the the `LogTrace` is never triggered. Now, I don't really care about the trace log or the stopwatch, so I propose we just remove them. @SeeminglyScience what do you think about that?

andyleejordan (changes requested):
Good catch, just need to fix up what's now dead code around it.

src/PowerShellEditorServices/Services/Symbols/Visitors/AstOperations.cs
andyleejordan: @dkattan I am happy to take this PR we just needed to fix this bit up. Did you want to do that? Otherwise I can get to it eventually.

andyleejordan: I will handle it so I can get this in.

dkattan: Thanks!

andyleejordan: Once i had it in my editor I saw what was going on 😄

andyleejordan (approved):

andyleejordan:
> ERROR: Your current culture is (Invariant Language (Invariant Country)), which is not associated with any language, consider changing your system culture or install the English-US help content using: Update-Help -UICulture en-US.

LMAO.

PR #2122, opened 2023-12-26
Add `UseNullPSHostUI` config so apps hosting PSES can disable it

dkattan:
# PR Summary

Meaningful output from PSES's internal PowerShell is suppressed if you don't specify -EnableConsoleRepl.

## PR Context

I'd like to be able to see the output from PSScriptAnalyzer but my application has a read-only terminal. 
This SwitchParameter prevents the UI from being set to NullPSHostUI without enabling Repl functionality.

Perhaps a better way to implement this would be to condition off the use of named pipes as I believe the intention of the following code is to prevent clobbering stdio since ConsoleRepl can only be used when connecting via named pipes.

```csharp
     UI = hostInfo.ConsoleReplEnabled || hostInfo.UseCurrentPSHostUI
                ? new EditorServicesConsolePSHostUserInterface(loggerFactory, hostInfo.PSHost.UI)
                : new NullPSHostUI();
```

### DIFF
--- a/src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
+++ b/src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
@@ -292,10 +292,17 @@ namespace Microsoft.PowerShell.EditorServices.Services.Analysis
             catch (CmdletInvocationException ex)
             {
                 // We do not want to crash EditorServices for exceptions caused by cmdlet invocation.
-                // Two main reasons that cause the exception are:
+                // The main reasons that cause the exception are:
                 // * PSCmdlet.WriteOutput being called from another thread than Begin/Process
                 // * CompositionContainer.ComposeParts complaining that "...Only one batch can be composed at a time"
-                _logger.LogError(ex.Message);
+                // * PSScriptAnalyzer not being able to find its PSScriptAnalyzer.psd1 because we are hosted by an Assembly other than pwsh.exe
+                string message = ex.Message;
+                if (!string.IsNullOrEmpty(ex.ErrorRecord.FullyQualifiedErrorId))
+                {
+                    // Microsoft.PowerShell.EditorServices.Services.Analysis.PssaCmdletAnalysisEngine: Exception of type 'System.Exception' was thrown. |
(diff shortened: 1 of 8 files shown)


src/PowerShellEditorServices/Services/Analysis/PssaCmdletAnalysisEngine.cs
```diff
@@ -292,10 +292,17 @@ private PowerShellResult InvokePowerShell(PSCommand command)
             catch (CmdletInvocationException ex)
             {
                 // We do not want to crash EditorServices for exceptions caused by cmdlet invocation.
-                // Two main reasons that cause the exception are:
+                // The main reasons that cause the exception are:
                 // * PSCmdlet.WriteOutput being called from another thread than Begin/Process
                 // * CompositionContainer.ComposeParts complaining that "...Only one batch can be composed at a time"
-                _logger.LogError(ex.Message);
+                // * PSScriptAnalyzer not being able to find its PSScriptAnalyzer.psd1 because we are hosted by an Assembly other than pwsh.exe
```
andyleejordan, line 298: Ooooh

andyleejordan (changes requested):
I like quite a bit of this but would want to clean it up a bit, put it through some thorough testing, and I think you are quite right that we should condition this based on named pipes instead of adding (yet another) parameter to the startup script. Let me know if you want to do some of that, and please tell me about the best way to go about testing this!

dkattan:
> I like quite a bit of this but would want to clean it up a bit, put it through some thorough testing, and I think you are quite right that we should condition this based on named pipes instead of adding (yet another) parameter to the startup script. Let me know if you want to do some of that, and please tell me about the best way to go about testing this!

Sweet, I'll go ahead and make that change

dkattan:
> please tell me about the best way to go about testing this!

This is going to be hard to test since it appears that all the tests use Stdio, unless I’m mistaken.

andyleejordan (approved):

andyleejordan:
That CI failure was a red herring, it was actually caused by GitHub Actions giving us a mix of PowerShell versions leading to a missed bug in a unit test, fixed by https://github.com/PowerShell/PowerShellEditorServices/pull/2135

PR #2124, opened 2024-01-03
Test end-to-end with PowerShell Daily

andyleejordan:
This downloads the PowerShell Daily build on CI and executes the end-to-end tests with it. It's set to run on a schedule and fail to alert us to issues (but won't block PRs because it won't run on them at all).


andyleejordan:
Honestly the only thing I'd want to change is to somehow run this "daily" like on a schedule instead of against PRs.

JustinGrote (approved):
LGTM if the rebase is OK, good addition!

PR #2129, opened 2024-01-10
Add sane defaults to `Start-EditorServices`

### LINKED ISSUE #1855: Simplify PSES CLI so startup with non-Code editors is easier

andyleejordan:
We do not need to require so many CLI flags.

This continues to run the existing Emacs and Vim tests utilizing those flags as a regression scenario, and adds an additional pair of tests that launch PSES with a minimal set of flags.

Resolves #1855.

Slight breaking change for CLI users of the `-LogPath` parameter. It now takes a directory name instead of a log name, which makes a lot more sense because multiple log files get written to that location.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
+++ b/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
@@ -33,6 +33,10 @@ namespace Microsoft.PowerShell.EditorServices.Commands
 
         private HostLogger _logger;
 
+        // NOTE: Ignore the suggestion to use Environment.ProcessId as it doesn't work for
+        // .NET 4.6.2 (for Windows PowerShell), and this won't be caught in CI.
+        private static readonly int s_currentPID = Process.GetCurrentProcess().Id;
+
         public StartEditorServicesCommand()
         {
             // Sets the distribution channel to "PSES" so starts can be distinguished in PS7+ telemetry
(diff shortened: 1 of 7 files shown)


src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
```diff
@@ -39,33 +39,44 @@ public StartEditorServicesCommand()
             Environment.SetEnvironmentVariable("POWERSHELL_DISTRIBUTION_CHANNEL", "PSES");
             _disposableResources = new List<IDisposable>();
             _loggerUnsubscribers = new List<IDisposable>();
+
+            // These are the actual defaults. Because of how PowerShell cmdlets work, the "defaults"
+            // in the corresponding .PS1 file don't matter (though they match for discoverability),
+            // nor can these be set at the field declaration because the engine resets parameters to
+            // null on invocation.
```
andyleejordan, line 46: This was so annoying...thanks GitHub Copilot for the help because I was pounding sand wondering why the defaults were still null.

src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
```diff
@@ -287,9 +298,12 @@ private void StartLogging()
...
         private string GetLogDirPath()
         {
-            string logDir = !string.IsNullOrEmpty(LogPath)
-                ? Path.GetDirectoryName(LogPath)
-                : Path.GetDirectoryName(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
+            string logDir = Path.GetDirectoryName(LogPath);
+            if (string.IsNullOrEmpty(logDir))
+            {
+                logDir = Directory.GetCurrentDirectory();
+                // Path.GetDirectoryName(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
```
andyleejordan, line 305: I really don't know why this was the previous safe default, do you Patrick?

andyleejordan (commented):
@SeeminglyScience please take a look!

src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
SeeminglyScience: While neither are particularly safe, I'd say the location of the assembly is marginally safer of the two. Process working directory is often somewhat arbitrary, could be a project directory, could be system32

Maybe `Path.GetTempPath()` + `\PowerShellEditorServices` would be good?

src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
```diff
@@ -118,7 +129,7 @@ public StartEditorServicesCommand()
         /// <summary>
         /// The path to where PowerShellEditorServices and its bundled modules are.
         /// </summary>
-        [Parameter]
+        [Parameter(Mandatory = true)]
         [ValidateNotNullOrEmpty]
         public string BundledModulesPath { get; set; }
```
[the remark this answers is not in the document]
SeeminglyScience: You mean that we should be able to figure out bundled modules path based on the assembly path? Yeah I'd agree

src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs
andyleejordan: I vaguely remember having an issue with using the temp path, not sure why...

SeeminglyScience (approved):
LGTM!

PR #2130, opened 2024-01-13
Fix up debugger attach handlers

andyleejordan:
First off, the error messages were never actually displayed to the user because the `RpcErrorException` constructor takes three arguments. Secondly, we do not support attaching to PowerShell Editor Services. It sure looked like we did (because we had special logic for it) but once attached, nothing worked. So it was half-baked. Now we throw an error if the user is trying to do that. Thirdly, because of that half-baked implementation, the process ID field was typed as a string (to support "current" as a shortcut)but that caused a mess here and an error in the VS Code client. Now it's just always an integer. (Same for the runspace ID.) Fourthly, a big mess was cleaned up by refactoring using functions, who'd have thought? Fifth and finally, superfluous version checking around PowerShell <5.1 was removed (as those versions are no longer supported whatsoever).

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebuggerActionHandlers.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebuggerActionHandlers.cs
@@ -51,7 +51,7 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
             }
             catch (NotSupportedException e)
             {
-                throw new RpcErrorException(0, e.Message);
+                throw new RpcErrorException(0, null, e.Message);
             }
         }
     }
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
@@ -76,9 +76,9 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
     {
         public string ComputerName { get; set; }
 
-        public string ProcessId { get; set; }
+        public int ProcessId { get; set; }
 
-        public string RunspaceId { get; set; }
+        public int RunspaceId { get; set; }
 
         public string RunspaceName { get; set; }
 
@@ -87,6 +87,7 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
     internal class LaunchAndAttachHandler : ILaunchHandler<PsesLaunchRequestArguments>, IAttachHandler<PsesAttachRequestArguments>, IOnDebugAdapterServerStarted
     {
+        private static readonly int currentProcessId = System.Diagnostics.Process.GetCurrentProcess().Id;
         private static readonly Version s_minVersionForCustomPipeName = new(6, 2);
         private readonly ILogger<LaunchAndAttachHandler> _logger;
         private readonly BreakpointService _breakpointService;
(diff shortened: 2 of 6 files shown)


src/PowerShellEditorServices/Services/PowerShell/Handlers/PSHostProcessAndRunspaceHandlers.cs
```diff
@@ -11,92 +11,97 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
...
-            request.ProcessId ??= "current";
+            IReadOnlyList<PSObject> runspaces = [];
 
-            // If the processId is a valid int, we need to run Get-Runspace within that process
-            // otherwise just use the current runspace.
-            if (int.TryParse(request.ProcessId, out int pid))
+            // If we're the host process we just use Get-Runspace.
+            if (request.ProcessId == currentProcessId)
+            {
+                PSCommand psCommand = new PSCommand().AddCommand(@"Microsoft.PowerShell.Utility\Get-Runspace");
```
andyleejordan, line 69: // TODO: Throw here, we don't handle this.

src/PowerShellEditorServices/Services/PowerShell/Handlers/PSHostProcessAndRunspaceHandlers.cs
```diff
@@ -11,92 +11,97 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
...
+            foreach (dynamic runspace in runspaces)
             {
-                foreach (dynamic runspace in runspaces)
+                // When we are attached to ourself, we cannot include ConsoleHost (Id 1) nor the
+                // current runspace.
+                //
+                // NOTE: Attaching to ourself doesn't currently work at all, but if it did, we would
+                // need to do this.
+                if (request.ProcessId == currentProcessId
+                    && (runspace.Id == 1 || runspace.Id == _runspaceContext.CurrentRunspace.Runspace.Id))
```
andyleejordan, line 93: Delete, will no longer be relevant. (Leave a note.)

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebuggerActionHandlers.cs
```diff
@@ -51,7 +51,7 @@ public override Task<PauseResponse> Handle(PauseArguments request, CancellationT
             }
             catch (NotSupportedException e)
             {
-                throw new RpcErrorException(0, e.Message);
+                throw new RpcErrorException(0, e, e.Message);
```
andyleejordan, line 54: Make all the second arguments null.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -395,38 +323,27 @@ void RunspaceChangedHandler(object s, RunspaceChangedEventArgs _)
                         _debugStateService.RunspaceId = id;
                         break;
 
-                        // TODO: If we don't end up setting this, we should throw
+                        // TODO: If we don't end up setting this, we should throw!
```
andyleejordan, line 326: If the IEnumerable is empty, throw an RpcException because we couldn't get the ID of the named runspace.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -395,38 +323,27 @@ void RunspaceChangedHandler(object s, RunspaceChangedEventArgs _)
...
-
-                debugRunspaceCmd.AddParameter("Id", runspaceId);
+                _debugStateService.RunspaceId = request.RunspaceId;
+                debugRunspaceCmd.AddParameter("Id", request.RunspaceId);
             }
             else
             {
                 _debugStateService.RunspaceId = 1;
-
                 debugRunspaceCmd.AddParameter("Id", 1);
```
andyleejordan, line 347: > throw new RpcErrorException(0, "A positive integer must be specified for the RunspaceId field.");

daxian-dbw (approved):

PR #2132, opened 2024-01-18
Remove `PowerShellEditorServices.VSCode` module / `ContentViews` feature

andyleejordan:
As it took a year (presumably, maybe a more recent VS Code update is what broke it) to come up that indicates its usage is too low for us to go about fixing. If someone wants to figure it out we'll be happy to take a PR. For now we're removing the code (which is also a massive simplification because it was the entirety of the "VSCode" module we had to build and ship with the extension).


andyleejordan:
Me too! Goal is to reduce the surface area we have to maintain. I want a stable, quality extension even if it's smaller! Not an extension trying to implement everything possible under the sun but failing at all the core features.

PR #2133, opened 2024-01-25
Inlay Hint Support

JustinGrote:
# PR Summary

This PR implements Inlay Hints to resolve the type in PowerShell when it can be determined by the PowerShell Type Inference completer. This is considered best effort and not exhaustive.
image

NOTE: This PR uses some modern C# syntax features that do not match the prevailing style. Worth exploring if these are OK to adopt:

- File Scoped Namespace (reduces indentation)
- Primary Constructor (simplifies dependency injection)
- LINQ Query Syntax (easier to follow symbol selection process)


JustinGrote:
@MartinGC94 @andyleejordan @SeeminglyScience  need to find where PSES gets this info from:
image

PR #2135, opened 2024-02-12
Fix PowerShell 7.2 E2E unit test for last change

andyleejordan:
PR #2115 added a new error message when no completion is returned (such as when it was interrupted) and our test covering a specific bug with PowerShell 7.2 was not updated. Investigating how CI missed this.


andyleejordan:
CI missed it because it's a real edge case that requires the end-to-end tests to run with the path explicitly set to use PowerShell 7.2 or 7.3 (since the path given in the build script is just "pwsh") and we somehow got an image for the PR and merge group that had 7.4, and then one for the push to main after the merge with 7.3. I guess.

SeeminglyScience (approved):
LGTM!

PR #2136, opened 2024-02-13
Support provider-backed workspace URIs via pspath

dkattan:
## Summary

- add `pspath://` document support so provider-backed scripts can participate in workspace operations without pretending to be `file://` paths
- route provider-backed workspace enumeration and content reads through PowerShell cmdlets when a host is available
- keep normal file-backed documents and workspace traversal on direct filesystem I/O to avoid cross-thread/apartment issues in ordinary editor requests
- preserve untitled/notebook handling and watched-file reload behavior for non-file URIs
- keep the workspace service surface synchronous and cover the provider-backed launch path in tests

## Why

The goal of this branch is to let PSES honor custom PSProviders for workspace content while still behaving normally for ordinary files. The final shape is intentionally hybrid: provider-backed URIs use PowerShell so custom providers are respected, and file-backed URIs stay on the existing filesystem path to avoid regressions in debugger, symbol, and code lens flows.

This also supports web-based VS Code scenarios where the editor does not share the same filesystem as the language server. In that setup, an extension can contribute a VS Code FileProvider for `pspath:` while PSES loads a matching PSProvider in its runspace; both sides can point at the same backing persistence mechanism, such as an API, so the client and the language server address the same logical content.

## Files changed

- `WorkspaceService` adds `pspath` conversion helpers plus provider-backed enumeration/read support
- `ScriptFile` explicitly recognizes supported URI schemes and treats only true in-memory schemes as untitled
- `DidChangeWatchedFilesHandler` reloads content through the workspace service and handles non-file URIs more safely
- tests cover provider-backed launch and updated URI classification

## Validation

- targeted provider-path unit coverage passes locally
- current PR CI is green on Ubuntu, macOS, and Windows, along with the policy checks

### DIFF
--- a/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
+++ b/src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
@@ -117,7 +117,8 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
         #region Public Methods
 
-        public IEnumerable<string> WorkspacePaths => WorkspaceFolders.Select(i => i.Uri.GetFileSystemPath());
+        public IEnumerable<string> WorkspacePaths => WorkspaceFolders.Select(
+            folder => folder.Uri.ToUri().IsFile ? folder.Uri.GetFileSystemPath() : GetPowerShellPath(folder.Uri));
 
         /// <summary>
         /// Gets an open file in the workspace. If the file isn't open but exists on the filesystem, load and return it.
@@ -391,21 +392,26 @@ namespace Microsoft.PowerShell.EditorServices.Services
             int maxDepth,
             bool ignoreReparsePoints)
         {
-            string[] workspacePaths = GetPowerShellWorkspacePaths()
+            string[] powerShellWorkspacePaths = GetPowerShellWorkspacePaths()
                 .Where(path => !string.IsNullOrEmpty(path))
                 .Distinct(StringComparer.OrdinalIgnoreCase)
                 .ToArray();
 
-            if (workspacePaths.Length == 0)
+            string[] fileSystemWorkspacePaths = GetFileSystemWorkspacePaths()
+                .Where(path => !string.IsNullOrEmpty(path))
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+
+            if (powerShellWorkspacePaths.Length == 0 && fileSystemWorkspacePaths.Length == 0)
             {
                 yield break;
             }
 
-            if (psesInternalHost is not null)
+            if (psesInternalHost is not null && powerShellWorkspacePaths.Length > 0)
             {
                 PSCommand psCommand = new PSCommand()
                     .AddCommand(@"Microsoft.PowerShell.Management\Get-ChildItem")
-                        .AddParameter("LiteralPath", workspacePaths)
+                        .AddParameter("LiteralPath", powerShellWorkspacePaths)
                         .AddParameter("Recurse")
                         .AddParameter("ErrorAction", ActionPreference.SilentlyContinue)
                         .AddParameter("Force")
@@ -437,7 +443,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
             foreach (string pattern in includeGlobs) { matcher.AddInclude(pattern); }
             foreach (string pattern in excludeGlobs) { matcher.AddExclude(pattern); }
 
-            foreach (string rootPath in workspacePaths)
+            foreach (string rootPath in fileSystemWorkspacePaths)
             {
                 if (!Directory.Exists(rootPath))
                 {
@@ -478,7 +484,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
         internal string ReadFileContents(DocumentUri uri)
         {
-            if (psesInternalHost is null)
+            if (uri.ToUri().IsFile || psesInternalHost is null)
             {
                 using StreamReader reader = OpenStreamReader(uri);
                 return reader.ReadToEnd();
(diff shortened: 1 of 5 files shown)


dkattan:
@andyleejordan Curious your thoughts on this. It ended up being heavier-handed than I wanted with all the async changes, but I'm wondering if that would be desirable if I split it into a separate MR? I know @SeeminglyScience doesn't want me changing the public interfaces, but they likely should've been Task<> from the start, async or not.

SeeminglyScience:
> I know @SeeminglyScience doesn't want me changing the public interfaces, but they likely should've been Task<> from the start, async or not.

Can you talk through what you're trying to solve? When you make a synchronous method return `Task<>` all you're really doing is requiring wrapper object be created that will be immediately discarded. It doesn't change the behavior at all if that's the assumption.

dkattan:
Rebased this branch onto the current `main` and force-pushed `feature/get-content` at `c9425602`.

On the sync/async concern: I kept `WorkspaceService` synchronous and reused the current `PsesInternalHost.InvokePSCommand(...)` synchronous helpers instead of widening the API surface or asyncifying the workspace methods. The branch now routes file reads / workspace enumeration through PowerShell when a host is available, keeps the existing non-host fallback for current callers/tests, and preserves `pspath` support for provider-backed documents.

I also added targeted coverage for provider-backed launch (`PSProviderPathsLaunchCorrectly`) and scheme handling.

test/PowerShellEditorServices.Test/Debugging/DebugServiceTests.cs
```diff
@@ -94,6 +104,15 @@ public async Task DisposeAsync()
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "VSTHRD110:Observe result of async calls", Justification = "This intentionally fires and forgets on another thread.")]
         private void OnDebuggerStopped(object sender, DebuggerStoppedEventArgs e) => Task.Run(() => debuggerStoppedQueue.Add(e));
 
+        private static string ConvertPSPathToUri(string psPath)
```
dkattan, line 111: Add some structured comments here with examples of what the input and output strings will look like

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -439,10 +482,113 @@ internal static StreamReader OpenStreamReader(DocumentUri uri)
...
+            catch (ActionPreferenceStopException ex)
+                when (ex.ErrorRecord.CategoryInfo.Category == ErrorCategory.ObjectNotFound
+                    && ex.ErrorRecord.TargetObject is string[] missingFiles
+                    && missingFiles.Length == 1)
+            {
+                throw new FileNotFoundException(ex.ErrorRecord.ToString(), missingFiles[0], ex.ErrorRecord.Exception);
+            }
+        }
+
+        private IEnumerable<string> GetFileSystemWorkspacePaths()
```
dkattan, line 519: Add structured comments with example input/output

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -439,10 +482,113 @@ internal static StreamReader OpenStreamReader(DocumentUri uri)
...
+                    .Where(uri => uri.ToUri().IsFile)
+                    .Select(uri => uri.GetFileSystemPath());
+            }
+
+            return string.IsNullOrEmpty(InitialWorkingDirectory)
+                ? Array.Empty<string>()
+                : new[] { InitialWorkingDirectory };
+        }
+
+        private IEnumerable<string> GetPowerShellWorkspacePaths()
```
dkattan, line 538: Add structured comments with example input/output

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -439,10 +482,113 @@ internal static StreamReader OpenStreamReader(DocumentUri uri)
...
+                return WorkspaceFolders
+                    .Select(folder => folder.Uri)
+                    .Where(uri => !uri.ToUri().IsFile)
+                    .Select(GetPowerShellPath);
+            }
+
+            return Array.Empty<string>();
+        }
+
+        private static string ConvertWorkspaceItemPath(PSObject item)
```
dkattan, line 555: Add structured comments with example input/output

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -439,10 +482,113 @@ internal static StreamReader OpenStreamReader(DocumentUri uri)
...
+            {
+                return fullName;
+            }
+
+            return item.Properties["PSPath"]?.Value is string psPath && !string.IsNullOrEmpty(psPath)
+                ? CreatePowerShellPathUri(psPath)
+                : null;
+        }
+
+        private static string GetPowerShellPath(DocumentUri uri)
```
dkattan, line 571: Add structured comments with example input/output

src/PowerShellEditorServices/Services/Workspace/WorkspaceService.cs
```diff
@@ -439,10 +482,113 @@ internal static StreamReader OpenStreamReader(DocumentUri uri)
...
+
+                return string.IsNullOrEmpty(provider)
+                    ? path.TrimStart('/')
+                    : $"{provider}::{path}";
+            }
+
+            throw new NotSupportedException($"Unsupported URI scheme '{uri.Scheme}'.");
+        }
+
+        private static string CreatePowerShellPathUri(string psPath)
```
dkattan, line 600: Add structured comments with example input/output

PR #2137, opened 2024-02-15
Fixed TextReader disposal

dkattan:
# PR Summary

Small nitpick I found when refactoring other stuff

https://stackoverflow.com/questions/2988977/passing-idisposable-objects-through-constructor-chains


andyleejordan (approved):
I like it, thanks!

PR #2156, opened 2024-04-02
Use a provided shell integration script directly

andyleejordan:
Instead of having to maintain an edited copy (which was really annoying) I stubbed out `PSConsoleHostReadLine` to do what's expected. So now we can just use the existing shell integration script directly!

Since we can't reliably find the script using `code --locate-shell-integration-path pwsh` we now rely on it being sent by the client on initialization. Its presence implies the feature is on.

This is pretty VS Code specific, but not necessarily so.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
@@ -315,16 +315,28 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
                 _logger.LogDebug("Profiles loaded!");
             }
 
-            if (startOptions.ShellIntegrationEnabled)
+            if (!string.IsNullOrEmpty(startOptions.ShellIntegrationScript))
             {
-                _logger.LogDebug("Enabling shell integration...");
+                _logger.LogDebug("Enabling Terminal Shell Integration...");
                 _shellIntegrationEnabled = true;
-                await EnableShellIntegrationAsync(cancellationToken).ConfigureAwait(false);
+                // TODO: Make the __psEditorServices prefix shared (it's used elsewhere too).
+                string setupShellIntegration = $$"""
+                    # Setup Terminal Shell Integration.
+
+                    # Define a fake PSConsoleHostReadLine so the integration script's wrapper
+                    # can execute it to get the user's input.
+                    $global:__psEditorServices_userInput = "";
+                    function global:PSConsoleHostReadLine { $global:__psEditorServices_userInput }
+
+                    # Execute the provided shell integration script.
+                    try { . '{{startOptions.ShellIntegrationScript}}' } catch {}
+                    """;
+                await EnableShellIntegrationAsync(setupShellIntegration, cancellationToken).ConfigureAwait(false);
                 _logger.LogDebug("Shell integration enabled!");
             }
             else
             {
-                _logger.LogDebug("Shell integration not enabled!");
+                _logger.LogDebug("Terminal Shell Integration not enabled!");
             }
 
             await _started.Task.ConfigureAwait(false);
@@ -495,6 +507,7 @@ namespace Microsoft.PowerShell.EditorServices.Services.PowerShell.Host
                 new SynchronousDelegateTask(_logger, representation, executionOptions, action, cancellationToken));
         }
 
+        // TODO: One day fix these so the cancellation token is last.
         public Task<IReadOnlyList<TResult>> ExecutePSCommandAsync<TResult>(
             PSCommand psCommand,
             CancellationToken cancellationToken,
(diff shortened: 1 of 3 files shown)


andyleejordan:
This was prompted by a) being annoyed at having to keep doing this and b) the changes in https://github.com/microsoft/vscode/commit/d60e424ff398a41f5e65827ec6e06b235273e019 being much more difficult to incorporate except by doing it this way! /cc @Tyriar

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -35,6 +35,16 @@ internal class PsesInternalHost : PSHost, IHostSupportsInteractiveSession, IRuns
...
         internal const string DefaultPrompt = "> ";
 
+        // TODO: Make the __psEditorServices prefix shared (it's used elsewhere too).
+        private const string ShellIntegrationScript = @"
+# Define fake PSConsoleHostReadLine so VS Code's script can execute it to get the user's input.
+$global:__psEditorServices_userInput = """";
+function global:PSConsoleHostReadLine { $global:__psEditorServices_userInput }
+
+# Execute VS Code's built-in shell integration script.
+try { . ""$(code --locate-shell-integration-path pwsh)"" } catch {}
```
andyleejordan, line 45: My main concern is not knowing if `code` will reliably in `PATH`...

Tyriar: You should be able to get the executable from the extension host's `process`. Also keep in mind there are other flavors like `code-insiders` and `code-exploration`.

andyleejordan: Perhaps the best thing to do would be to get the path to the script in the client and pass that along to the server to use here.

andyleejordan: Fixed it up.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -315,16 +315,27 @@ public async Task<bool> TryStartAsync(HostStartOptions startOptions, Cancellatio
...
 
-            if (startOptions.ShellIntegrationEnabled)
+            if (!string.IsNullOrEmpty(startOptions.ShellIntegrationScript))
             {
-                _logger.LogDebug("Enabling shell integration...");
+                _logger.LogDebug("Enabling VS Code's shell integration...");
                 _shellIntegrationEnabled = true;
-                await EnableShellIntegrationAsync(cancellationToken).ConfigureAwait(false);
+                // TODO: Make the __psEditorServices prefix shared (it's used elsewhere too).
+                string setupShellIntegration = $$"""
```
andyleejordan, line 323: I like C# 11's triple-quote double-dollar thing. Very useful.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -1262,16 +1078,34 @@ private void InvokeInput(string input, CancellationToken cancellationToken)
...
+                        new PowerShellExecutionOptions { ThrowOnError = false, WriteOutputToHost = false },
+                        cancellationToken);
+
+                    // Invoke the PSConsoleHostReadLine wrapper. We don't write the output because it
+                    // returns the command line (user input) which would then be duplicate noise. Fortunately
+                    // it writes the shell integration sequences directly using [Console]::Write.
+                    InvokePSCommand(
+                        new PSCommand().AddScript("PSConsoleHostReadLine"),
+                        new PowerShellExecutionOptions { ThrowOnError = false, WriteOutputToHost = false },
+                        cancellationToken);
```
andyleejordan, line 1098: @Tyriar just a heads-up here that your `PSConsoleHostReadLine` wrapper _must_ continue to write those escape sequences directly, as we're intentionally not emitting the output (like the returned value) to the host. `Write-Host` might work (though you don't use it), but we'd need to test.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -1262,16 +1078,34 @@ private void InvokeInput(string input, CancellationToken cancellationToken)
...
+
+                    // Reset our global variable.
+                    InvokePSCommand(
+                        new PSCommand().AddScript("$global:__psEditorServices_userInput = \"\""),
+                        new PowerShellExecutionOptions { ThrowOnError = false, WriteOutputToHost = false },
+                        cancellationToken);
                 }
 
                 InvokePSCommand(
-                    new PSCommand().AddScript(input, useLocalScope: false),
```
andyleejordan, line 1274: No real change, it's the default.

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -315,16 +315,28 @@ public async Task<bool> TryStartAsync(HostStartOptions startOptions, Cancellatio
...
+                string setupShellIntegration = $$"""
+                    # Setup Terminal Shell Integration.
+
+                    # Define a fake PSConsoleHostReadLine so the integration script's wrapper
+                    # can execute it to get the user's input.
+                    $global:__psEditorServices_userInput = "";
+                    function global:PSConsoleHostReadLine { $global:__psEditorServices_userInput }
+
+                    # Execute the provided shell integration script.
+                    try { . "{{startOptions.ShellIntegrationScript}}" } catch {}
```
SeeminglyScience, line 332:
```suggestion
                    try { . '{{startOptions.ShellIntegrationScript}}' } catch {}
```

^ assuming we don't rely on a variable in the path. Just disables variable expansion so we don't need to worry about potential bugs there or code injection

src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs
```diff
@@ -1262,16 +1078,34 @@ private void InvokeInput(string input, CancellationToken cancellationToken)
...
-                // to users of PSES, since we already wrap ReadLine ourselves.
+                // For the terminal shell integration feature, we call PSConsoleHostReadLine specially as it's been wrapped.
+                // Normally it would not be available (since we wrap ReadLine ourselves),
+                // but in this case we've made the original just emit the user's input so that the wrapper works as intended.
                 if (_shellIntegrationEnabled)
                 {
-                    System.Console.Write("\x1b]633;C\a");
+                    // Save the user's input to our special global variable so PSConsoleHostReadLine can read it.
+                    InvokePSCommand(
+                        new PSCommand().AddScript($"$global:__psEditorServices_userInput = \"{input}\""),
```
SeeminglyScience, line 1088:
```suggestion
                        new PSCommand().AddScript("$global:__psEditorServices_userInput = $args[0]").AddArgument(input),
```

Similar reasons, plus don't have to worry about escaping. Like if the REPL line was `"normal string here"` that would have issues.

SeeminglyScience (changes requested):
Lookin' good, just two fixes real quick

SeeminglyScience (approved):
LGTM!

PR #2167, opened 2024-07-16
Look in all workspace folders for PSSA settings file

andyleejordan:
Mostly investigating https://github.com/PowerShell/vscode-powershell/issues/4743.

### DIFF
--- a/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
+++ b/src/PowerShellEditorServices/Services/Analysis/AnalysisService.cs
@@ -322,7 +322,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
                 return false;
             }
 
-            settingsFilePath = _workspaceService?.ResolveWorkspacePath(configuredPath);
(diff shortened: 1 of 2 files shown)


andyleejordan:
@JustinGrote @SeeminglyScience would you two mind reviewing this? I think there's reason enough to get it in. It doesn't fix _everything_ since the logic is just "find the first PSSA config" and not a specific folder's config, but the latter is a big ask and the former at least gives a solution to multi-root repos. And it's come up a few times now.

JustinGrote:
@andyleejordan I couldn't really tell but does it look thru the workspaces in a deterministic fashion? My thought was that since you can specify the folder order in the workspace, if it always goes thru the folder order, then a user can define the "primary" analyzer by putting that folder to the top.

Also agreed a per-folder analysis config that also applies different rules to each folder is a way larger ask.

andyleejordan:
Yes, it should iterate over the workspace folders deterministically and in the order supplied. So you're right, you can force it by ordering those in the workspace settings.

SeeminglyScience (approved):
LGTM!

PR #2173, opened 2024-08-23
Mention debugging support in intellij-powershell

ForNeVeR:
Hi! I am a maintainer of intellij-powershell, and we recently have released a plugin version that supports debugging using PSES (that works good, thank you folks so much!)

Since it is mentioned in the language section of the README already, I decided to add it to the debugger section as well.


andyleejordan (approved):

andyleejordan:
Thanks so much!

Issue #2182, opened 2024-10-02
VIM E2E Test Broken
Assignees: @andyleejordan @JustinGrote
Labels: bug
Milestone: 
Projects: 


JustinGrote:
https://github.com/PowerShell/PowerShellEditorServices/actions/runs/11132769096/job/30937371296?pr=2152

JustinGrote:
Test works fine in upterm, weird...
![image](https://github.com/user-attachments/assets/d5a803c8-39f1-411d-b8d5-7c5d6538b53c)

JustinGrote:
@andyleejordan I think whatever VIM you're downloading, something in the latest version broke the tests. If I remove the ENV to the executable, the test completes, see this PR:
https://github.com/PowerShell/PowerShellEditorServices/pull/2183

JustinGrote:
Looks like package updated to vim-gtk3 8.2.3995-1ubuntu2.19, it was ubuntu2.18 before, something in that update appears to have broken it.

JustinGrote:
@andyleejordan nightly works fine, so I adjusted the github action to use nightly, see this PR:
https://github.com/PowerShell/PowerShellEditorServices/pull/2183

We will need to actually verify if the extension is actually broken due to this security update package, since it is the current stable ubuntu package and will have widespread usage.

andyleejordan:
Yay for fun new bugs.

andyleejordan:
Resolved — the Vim E2E workflow installs Vim via `rhysd/action-setup-vim` (nightly), bypassing the broken apt package. Closing as completed.

---
_Drafted by GitHub Copilot (Claude Opus 4.8) at @andyleejordan's request._

PR #2183, opened 2024-10-02
Switch VIM E2E tests to use nightly

JustinGrote:
The current ubuntu2.19 package breaks E2E tests. Latest VIM works fine, so switching to nightly for testing.

Band-Aid for https://github.com/PowerShell/PowerShellEditorServices/issues/2182 but will not close it as we should probably find out what in that 2.19 ubuntu security update broke it, as it is currently being distributed widely and we should make sure it hasn't broken ubuntu VIM users.

### DIFF
--- a/.github/workflows/vim-test.yml
+++ b/.github/workflows/vim-test.yml
@@ -35,7 +35,10 @@ jobs:
         run: Invoke-Build Build
 
       - name: Install Vim
+        id: vim
         uses: rhysd/action-setup-vim@v1
+        with:
+          version: nightly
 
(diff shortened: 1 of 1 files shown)


JustinGrote:
Rebased and ready.

andyleejordan (approved):
Thanks Justin!

andyleejordan:
I wonder why it's passing in the PR but failing in the merge queue?!

andyleejordan:
Oh I can answer that. Only the merge queue installs and tests against PowerShell daily (since that takes a bit longer) and that's not currently working.

PR #2186, opened 2024-10-10
Allow daily PowerShell test to fail

andyleejordan:
Really, it's the installation that's failing because literally the upstream posting of new daily PowerShell builds is broken.

If it succeeds, TestFull and the daily schedule will run it. If it doesn't, they'll ignore it because the build script will skip it.


SeeminglyScience (approved):
LGTM!

PR #2190, opened 2024-10-12
Add Breakpoint Label frame to optimize debug stepping performance

JustinGrote:
# PR Summary
Implements DAP `DelayedStackTraceLoading` and returns an artificial breakpoint label frame based on the invocation info ASAP once the debugger stops. The rest of the stack trace is then returned on a future request that doesn't block the client UI.

https://github.com/user-attachments/assets/b909655b-1085-4f8e-a453-99b8af21b877

## PR Context
Debug stepping takes 300ms+ due to waiting for stack trace collection via the slow `Get-PSCallStack` "hack". Further improvements will come around this later.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -6,8 +6,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Management.Automation;
-using System.Management.Automation.Language;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -50,7 +48,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
         private VariableContainerDetails scriptScopeVariables;
         private VariableContainerDetails localScopeVariables;
         private StackFrameDetails[] stackFrameDetails;
-        private readonly PropertyInfo invocationTypeScriptPositionProperty;
 
         private readonly SemaphoreSlim debugInfoHandle = AsyncUtils.CreateSimpleLockingSemaphore();
         #endregion
@@ -74,6 +71,12 @@ namespace Microsoft.PowerShell.EditorServices.Services
         /// is stopped.
         /// </summary>
         public DebuggerStoppedEventArgs CurrentDebuggerStoppedEventArgs { get; private set; }
+
+        /// <summary>
+        /// Returns a task that completes when script frames and variables have completed population
+        /// </summary>
+        public Task StackFramesAndVariablesFetched { get; private set; }
+
 
         /// <summary>
         /// Tracks whether we are running <c>Debug-Runspace</c> in an out-of-process runspace.
@@ -111,12 +114,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
             _debugContext.DebuggerResuming += OnDebuggerResuming;
             _debugContext.BreakpointUpdated += OnBreakpointUpdated;
             _remoteFileManager = remoteFileManager;
-
-            invocationTypeScriptPositionProperty =
-                typeof(InvocationInfo)
-                    .GetProperty(
-                        "ScriptPosition",
-                        BindingFlags.NonPublic | BindingFlags.Instance);
         }
 
         #endregion
@@ -981,8 +978,8 @@ namespace Microsoft.PowerShell.EditorServices.Services
                     }
                 }
 
-                // Get call stack and variables.
-                await FetchStackFramesAndVariablesAsync(noScriptName ? localScriptPath : null).ConfigureAwait(false);
+                // Begin call stack and variables fetch. We don't need to block here.
+                StackFramesAndVariablesFetched = FetchStackFramesAndVariablesAsync(noScriptName ? localScriptPath : null);
 
                 // If this is a remote connection and the debugger stopped at a line
                 // in a script file, get the file contents
@@ -994,53 +991,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
                         await _remoteFileManager.FetchRemoteFileAsync(
                             e.InvocationInfo.ScriptName,
                             _psesHost.CurrentRunspace).ConfigureAwait(false);
-                }
-
-                if (stackFrameDetails.Length > 0)
-                {
-                    // Augment the top stack frame with details from the stop event
-                    if (invocationTypeScriptPositionProperty.GetValue(e.InvocationInfo) is IScriptExtent scriptExtent)
-                    {
-                        StackFrameDetails targetFrame = stackFrameDetails[0];
-
-                        // Certain context changes (like stepping into the default value expression
-                        // of a parameter) do not create a call stack frame. In order to represent
-                        // this context change we create a fake call stack frame.
-                        if (!string.IsNullOrEmpty(scriptExtent.File)
-                            && !PathUtils.IsPathEqual(scriptExtent.File, targetFrame.ScriptPath))
-                        {
-                            await debugInfoHandle.WaitAsync().ConfigureAwait(false);
-                            try
-                            {
-                                targetFrame = new StackFrameDetails
-                                {
-                                    ScriptPath = scriptExtent.File,
-                                    // Just use the last frame's variables since we don't have a
-                                    // good way to get real values.
-                                    AutoVariables = targetFrame.AutoVariables,
-                                    CommandVariables = targetFrame.CommandVariables,
-                                    // Ideally we'd get a real value here but since there's no real
-                                    // call stack frame for this, we'd need to replicate a lot of
-                                    // engine code.
-                                    FunctionName = "<ScriptBlock>",
-                                };
-
-                                StackFrameDetails[] newFrames = new StackFrameDetails[stackFrameDetails.Length + 1];
-                                newFrames[0] = targetFrame;
-                                stackFrameDetails.CopyTo(newFrames, 1);
-                                stackFrameDetails = newFrames;
-                            }
-                            finally
-                            {
-                                debugInfoHandle.Release();
-                            }
-                        }
-
-                        targetFrame.StartLineNumber = scriptExtent.StartLineNumber;
-                        targetFrame.EndLineNumber = scriptExtent.EndLineNumber;
-                        targetFrame.StartColumnNumber = scriptExtent.StartColumnNumber;
-                        targetFrame.EndColumnNumber = scriptExtent.EndColumnNumber;
-                    }
                 }
 
                 CurrentDebuggerStoppedEventArgs =
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -77,7 +77,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
         /// </summary>
         public Task StackFramesAndVariablesFetched { get; private set; }
 
-
         /// <summary>
         /// Tracks whether we are running <c>Debug-Runspace</c> in an out-of-process runspace.
         /// </summary>
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
@@ -20,14 +20,17 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
         public Task<ScopesResponse> Handle(ScopesArguments request, CancellationToken cancellationToken)
         {
-            VariableScope[] variableScopes =
-                _debugService.GetVariableScopes(
-                    (int)request.FrameId);
+            //We have an artificial breakpoint label, so just copy the stacktrace from the first stack entry for this.
+            int frameId = request.FrameId == 0 ? 0 : (int)request.FrameId - 1;
+
+            VariableScope[] variableScopes = _debugService.GetVariableScopes(frameId);
 
             return Task.FromResult(new ScopesResponse
             {
-                Scopes = new Container<Scope>(variableScopes
-                    .Select(LspDebugUtils.CreateScope))
+                Scopes = new Container<Scope>(
+                    variableScopes
+                    .Select(LspDebugUtils.CreateScope)
+                )
             });
         }
     }
--- a/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
@@ -18,12 +18,20 @@ namespace Microsoft.PowerShell.EditorServices.Handlers
 
         public ScopesHandler(DebugService debugService) => _debugService = debugService;
 
+        /// <summary>
+        /// Retrieves the variable scopes (containers) for the currently selected stack frame. Variables details are fetched via a separate request.
+        /// </summary>
         public Task<ScopesResponse> Handle(ScopesArguments request, CancellationToken cancellationToken)
         {
-            //We have an artificial breakpoint label, so just copy the stacktrace from the first stack entry for this.
-            int frameId = request.FrameId == 0 ? 0 : (int)request.FrameId - 1;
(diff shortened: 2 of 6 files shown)


JustinGrote:
@SeeminglyScience @andyleejordan I made this a fairly minimal PR to be easy to review. Additional things I plan to do next as future incrementals
- Add start/end/column positioning to all stack frames (it's available in the position property, just has to be serialized properly)
- Separate variable and frame processing into independent resolution paths
- Implement "fast path" for local runspace sessions to use the runspace debugger properties and methods rather than going thru the pipeline to fetch the frame info, ~2-3ms vs 300ms-800ms
- Move all the scope/stack/variable resolution to a new async-based `DebugStoppedContext` class that can be disposed and re-instantiated cleanly and kept separate from the `DebugService` concerns.

andyleejordan:
> @SeeminglyScience @andyleejordan I made this a fairly minimal PR to be easy to review. Additional things I plan to do next as future incrementals
> 
> * Add start/end/column positioning to all stack frames (it's available in the position property, just has to be serialized properly)
> * Separate variable and frame processing into independent resolution paths
> * Implement "fast path" for local runspace sessions to use the runspace debugger properties and methods rather than going thru the pipeline to fetch the frame info, ~2-3ms vs 300ms-800ms
> * Move all the scope/stack/variable resolution to a new async-based `DebugStoppedContext` class that can be disposed and re-instantiated cleanly and kept separate from the `DebugService` concerns, since LSP spec says no IDs/etc. should be reused between debug adapter stop/resumes

I'm super excited for all of this.

andyleejordan (commented):
@SeeminglyScience and I are going to take a closer look at this just to be sure nothing got missed.

JustinGrote:
This has been rebased to the latest prerelease and is good for re-review.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -75,6 +72,12 @@ internal class DebugService
         /// </summary>
         public DebuggerStoppedEventArgs CurrentDebuggerStoppedEventArgs { get; private set; }
 
+        /// <summary>
+        /// Returns a task that completes when script frames and variables have completed population
+        /// </summary>
+        public Task StackFramesAndVariablesFetched { get; private set; }
+
+
```
SeeminglyScience, line 80:
```suggestion
```

Extra line, hopefully this suggestion works

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
```diff
@@ -20,14 +20,17 @@ internal class ScopesHandler : IScopesHandler
 
         public Task<ScopesResponse> Handle(ScopesArguments request, CancellationToken cancellationToken)
         {
-            VariableScope[] variableScopes =
-                _debugService.GetVariableScopes(
-                    (int)request.FrameId);
+            //We have an artificial breakpoint label, so just copy the stacktrace from the first stack entry for this.
```
andyleejordan, line 23:
```suggestion
            // We have an artificial breakpoint label, so just copy the stacktrace from the first stack entry for this.
```

Could ou explain why this is either 0, or (and this is what's changed) `FrameId - 1`? That means there are two instances where we get scope 0: `FrameId == 0` _and_ `FrameId == 1`.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
-            long maxFrameCount = stackFrameDetails.Length;
-
-            // If the number of requested levels == 0 (or null), that means get all stack frames
-            // after the specified startFrame index. Otherwise get all the stack frames.
-            long requestedFrameCount = request.Levels ?? 0;
-            if (requestedFrameCount > 0)
-            {
-                maxFrameCount = Math.Min(maxFrameCount, startFrameIndex + requestedFrameCount);
-            }
+        StackFrame breakpointLabel = CreateBreakpointLabel(invocationInfo);
```
andyleejordan, line 40: So by "label" we mean a fake `StackFrame` always pushed on top of the stack right?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+        if (skip == 0 && take == 1) // This indicates the client is doing an initial fetch, so we want to return quickly to unblock the UI and wait on the remaining stack frames for the subsequent requests.
+        {
+            return new StackTraceResponse()
             {
-                // Create the new StackFrame object with an ID that can
-                // be referenced back to the current list of stack frames
-                newStackFrames.Add(LspDebugUtils.CreateStackFrame(stackFrameDetails[i], id: i));
-            }
+                StackFrames = new StackFrame[] { breakpointLabel },
+                TotalFrames = INITIAL_PAGE_SIZE //Indicate to the client that there are more frames available
```
andyleejordan, line 47:
```suggestion
                TotalFrames = INITIAL_PAGE_SIZE // Indicate to the client that there are more frames available
```

Does this have any effect on the client's UI for different values of `TotalFrames`? I ask since we arbitrarily set it to 20.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+            };
+        }
+
+        // Wait until the stack frames and variables have been fetched.
+        await debugService.StackFramesAndVariablesFetched.ConfigureAwait(false);
```
andyleejordan, line 52: Should this be cancellable?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+    /// Because we don't know the size of the stacktrace beforehand, we will tell the client that there are more frames available, this is effectively a paging size, as the client should request this many frames after the first one.
+    /// </summary>
+    private const int INITIAL_PAGE_SIZE = 20;
+
+    public async Task<StackTraceResponse> Handle(StackTraceArguments request, CancellationToken cancellationToken)
     {
-        private readonly DebugService _debugService;
+        if (!debugService.IsDebuggerStopped)
+        {
+            throw new NotSupportedException("Stacktrace was requested while we are not stopped at a breakpoint.");
```
SeeminglyScience, line 29: Do we know for sure this is caught somewhere? I want to make sure we don't crash if our assumptions are incorrect and this can happen.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+        {
+            StackFrames = newStackFrames,
+            TotalFrames = newStackFrames.Count
+        };
+    }
+
+    public static StackFrame CreateStackFrame(StackFrameDetails stackFrame, long id)
+    {
+        SourcePresentationHint sourcePresentationHint =
+            stackFrame.IsExternalCode ? SourcePresentationHint.Deemphasize : SourcePresentationHint.Normal;
```
andyleejordan, line 91: Is this new?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+
+        // When debugging an interactive session, the ScriptPath is <No File> which is not a valid source file.
+        // We need to make sure the user can't open the file associated with this stack frame.
+        // It will generate a VSCode error in this case.
+        Source? source = null;
+        if (!stackFrame.ScriptPath.Contains("<"))
```
andyleejordan, line 97: So, I get you're checking for `<No File>` which is fair _but_ `<` is a valid file character on non-Windows (err NTFS) filesystems, so you need to check for the whole thing `<No File>` (and we might also need to check for translations of that? Unsure...and while important, we're already not handling localization super great).

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -1,64 +1,132 @@
...
+                StackFrames = new StackFrame[] { breakpointLabel },
+                TotalFrames = INITIAL_PAGE_SIZE //Indicate to the client that there are more frames available
+            };
+        }
+
+        // Wait until the stack frames and variables have been fetched.
+        await debugService.StackFramesAndVariablesFetched.ConfigureAwait(false);
+
+        StackFrameDetails[] stackFrameDetails = await debugService.GetStackFramesAsync(cancellationToken)
+                                                                    .ConfigureAwait(false);
```
SeeminglyScience, line 55:
```suggestion
            .ConfigureAwait(false);
```

Style nit, prefer no hanging indent

src/PowerShellEditorServices/Utility/LspDebugUtils.cs
```diff
@@ -56,38 +56,6 @@ public static Breakpoint CreateBreakpoint(
             };
         }
 
-        public static StackFrame CreateStackFrame(
-            StackFrameDetails stackFrame,
-            long id)
-        {
-            SourcePresentationHint sourcePresentationHint =
-                stackFrame.IsExternalCode ? SourcePresentationHint.Deemphasize : SourcePresentationHint.Normal;
```
andyleejordan, line 64: Nope, not new.

SeeminglyScience (dismissed):
Thanks Justin! Couple of questions and minor requests

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
andyleejordan, line 29: Patrick may have also said this but we should be sure in the cases where we throw that we _really mean to throw_ because if not handled, it'll crash (and so often it's better to just log a warning and return if that's more graceful).

andyleejordan (changes requested):
Loving this, we finally got to review through it. Just a couple changes requested and then we want to get into pre-release.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: I will verify.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: Yes, [a label is a special kind of stackframe in the DAP protocol](https://microsoft.github.io/debug-adapter-protocol/specification#Types_StackFrame) - See presentation hint property.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: IDK,[ it's part of the spec](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Source).

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: Fair, this was supposed to be one of those "this should never happen" exceptions that if it happens its 99.99% chance its a bug and we *want* it to be reported. I was going to update the handling of that in my reporter updates but for now I can do that.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: Good point, will do.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: Probably, I'll wire that up if I can.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: No, according to the spec[ `a client should be prepared to receive fewer frames than requested, which is an indication that the end of the stack has been reached.`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_StackTrace)

JustinGrote:
@andyleejordan @SeeminglyScience your opinions have officially been deemed irrelevant by our AI overlords, please approve as is /s
image

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
JustinGrote: Uh, there was a reason, but I honestly don't remember at this point. Let me check

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
JustinGrote: Didn't work in vscode but I fixed it anyways.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: I converted it to RPCErrorException so that it only errors out the handler. I have a HandlerErrorException in the rename PR that really simplifies this kind of stuff.

JustinGrote: I changed it to RpcErrorException so that it will only kill the handler and log an error to PSES. I have a better Exception for this in the rename PR that eventually we will backport all these RpcErrorExceptions to.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: So this task is created by the debugService, and there's no cancellation anywhere in that process anyways
https://github.com/PowerShell/PowerShellEditorServices/blob/939e2c4604b20fd7731a3eaa63e7c2b62046d339/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs#L981-L982

So I'm gonna say for the scope of this PR, it's not worth wiring all that up when it doesn't exist today, and that the task has been completing without fail AFAIK, but that's definitely a good refactor to consider, especially if we do the dependency injection.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: I think I meant for it to be specifically that and maybe I got off on another idea or something. Good Catch.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
JustinGrote: My vscode identation checker hates you but OK :)

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/ScopesHandler.cs
JustinGrote: OK, I remember now, it's done so that the variables populate correctly in the debug pane for the label. I've added more of a comment explaining this.

JustinGrote:
All feedback should be addressed. I also removed some of the TraceOutput log stuff I was experimenting with, I'm going to submit a separate test-specific PR to revamp the DAP E2E tests.

andyleejordan (approved):
YOLO, I mean I love it and it's been thoroughly reviewed. Thanks Justin! Excited to get this out.

JustinGrote:
@andyleejordan this wording sounded so unnecessarily passive aggressive, LOL
image

Thanks! Excited to speed this up, it's annoyed me for YEARS.

PR #2193, opened 2024-10-29
Fix unit test broken by GitHub Action image update

andyleejordan:
Which now includes an AWS module with a conflicting completion (`AWS.Tools.SocialMessaging.psd1`). Why GitHub Actions installed that, I don't know. Why our test made such a broad assumption, also unknown.


SeeminglyScience (approved):
LGTM!

PR #2194, opened 2024-10-29
Bump packages for component governance

andyleejordan:
Which means the PowerShell SDK so its transitive dependencies update.


SeeminglyScience (approved):
LGTM!

PR #2206, opened 2024-12-04
Enable Roslynator Analyzers via Nuget Package

JustinGrote:
**This requires the following packages to be added to the Build Feed**
```
Roslynator.Analyzers
Roslynator.CodeAnalysis.Analyzers
Roslynator.Formatting.Analyzers
```

Rosylnator used to be used via vscode extension but since the switch to C# Dev Kit, Support for that has been non-functional (https://github.com/dotnet/vscode-csharp/issues/6790)

The analysis packages are now made available as Nuget packages, and this PR integrates them into the non-test repositories. This is the first step to later revise EditorConfig to enforce consistent code style among repositories.

Style-related recommendations based on current editorconfig (these are info recommendations and not enforced)
```
Analyzed solution 'D:\PowerShellEditorServices\PowerShellEditorServices.sln' (in 22.4 s)

  4 CA1510    Use ArgumentNullException throw helper
  2 CA1826    Do not use Enumerable methods on indexable collections
  2 CA1837    Use 'Environment.ProcessId'
  6 CA1859    Use concrete types when possible for improved performance
 10 CA1861    Avoid constant arrays as arguments
  1 CA2208    Instantiate argument exceptions correctly
  1 CA2249    Consider using 'string.Contains' instead of 'string.IndexOf'
  3 CS1998    Async method lacks 'await' operators and will run synchronously
  4 CS4014    Because this call is not awaited, execution of the current method continues before the call is completed
 74 IDE0028   Simplify collection initialization
 12 IDE0090   Use 'new(...)'
  2 RCS1015   Use nameof operator
  8 RCS1036   Remove unnecessary blank line
  1 RCS1075   Avoid empty catch clause that catches System.Exception
  4 RCS1097   Remove redundant 'ToString' call
  9 RCS1163   Unused parameter
  2 RCS1214   Unnecessary interpolated string
  1 RCS1218   Simplify code branching
 14 RCS1251   Remove unnecessary braces from record declaration
  2 RCS1267   Use string interpolation instead of 'string.Concat'
  1 VSTHRD104 Offer async methods
  3 xUnit1004 Test methods should not be skipped
  4 xUnit1042 The member referenced by the MemberData attribute returns untyped data rows
 26 xUnit2023 Do not use collection methods for single-item collections
```

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,6 +13,18 @@
     <PackageVersion Include="OmniSharp.Extensions.LanguageClient" Version="0.19.9" />
     <PackageVersion Include="OmniSharp.Extensions.LanguageServer" Version="0.19.9" />
     <PackageVersion Include="PowerShellStandard.Library" Version="5.1.1" />
+    <PackageVersion Include="Roslynator.Analyzers" Version="4.12.9">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageVersion>
+    <PackageVersion Include="Roslynator.CodeAnalysis.Analyzers" Version="4.12.9">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageVersion>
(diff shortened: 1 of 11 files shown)


andyleejordan:
Ok I built locally which ensured those packages were pulled into the Azure Artifacts feed, and then I deleted a bunch of blank lines etc. to satisfy the existing Roslynator rules.

andyleejordan (approved):
Thanks as always Justin! Happy new year!!!

PR #2218, opened 2025-02-26
OneBranch now uses PowerShell 7.4 so remove workarounds

andyleejordan:
We can rely on the pre-installed PSResourceGet and we can re-enable tests!

@JustinGrote seriously we can _finally_ run tests again in the release pipeline 🤞


SeeminglyScience (approved):
LGTM!

PR #2220, opened 2025-02-26
Skip test that's too slow in ADO

andyleejordan:
And remove test that's no longer run since <7.4 has been dropped.


SeeminglyScience (approved):
LGTM!

PR #2222, opened 2025-02-28
Surface Additional Parameter Documentation/Tooltips

JustinGrote:
Fixes an issue in type parsing where, if a type is discovered, additional documentation gets trimmed off. This restores that to the `Documentation` field.

I did a little experimenting with the new LabelDetail fields in LSP 3.17, but at least in vscode there was little to no visual advantage, something to maybe explore more later.

Would be nice to not have to text parse a string like this but it's where we are at currently with PS API

Related: https://github.com/PowerShell/PowerShell/pull/25108
image

Related: https://github.com/PowerShell/PowerShell/pull/25112
image

- [x] Needs Tests


JustinGrote:
@andyleejordan ready for review. Won't really do anything until this stuff gets populated on the PS side, but it does fix some tests and lints as well.

andyleejordan (approved):
Amazing, thanks.

PR #2226, opened 2025-03-18
Package and module updates

andyleejordan:
Minus a test package that would require bumping to `net472`.


SeeminglyScience (approved):
LGTM!

PR #2230, opened 2025-05-27
Update deprecated API method in docs

924312:
# PR Summary

Not super familiar with Neovim but this is deprecated
```lua
vim.api.nvim_buf_set_option(bufnr, 'omnifunc', 'v:lua.vim.lsp.omnifunc')
```
and I believe this is the fix
```lua
vim.api.nvim_set_option_value("omnifunc", "v:lua.vim.lsp.omnifunc", { buf = bufnr })
```


andyleejordan (approved):
TY

PR #2234, opened 2025-06-20
Allow new log levels in Start-EditorServices.ps1

sharpchen:
# PR Summary

Server yells against to be deprecated log levels, while new log levels are not allow for `$LogLevel` parameter in Start-EditorServices.ps1

## PR Context

https://github.com/PowerShell/PowerShellEditorServices/commit/d9de5bd8b1ed624a687d701fa2616e5991db2e2c

### DIFF
--- a/module/PowerShellEditorServices/Start-EditorServices.ps1
+++ b/module/PowerShellEditorServices/Start-EditorServices.ps1
@@ -46,7 +46,7 @@ param(
     [ValidateNotNullOrEmpty()]
     $LogPath,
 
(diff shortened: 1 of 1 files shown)


JustinGrote:
I'm at Psconfeu but I will review this! Looks straightforward, just actually need to test it.

Thanks for plugging a gap in my PR!

JustinGrote:
Thanks so much for your PR but I'm closing this as a dupe of #2236 as it has more detail. Check over there and stay tuned! Just one clarification needed before I merge it.

Issue #2235, opened 2025-06-21
Missing MEL levels from LogLevel param in Start-EditorServices.ps1


eitamal:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [_open and closed_ issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been reported.
- [x] I have read the [troubleshooting](https://github.com/PowerShell/vscode-powershell/blob/main/docs/troubleshooting.md) guide.
- [x] I am sure this issue is with _PowerShell Editor Services itself_ and does not reproduce in a standalone [PowerShell](https://github.com/PowerShell/PowerShell/issues/new/choose) instance, and is not an issue with my editor.
- [x] I have verified that I am using the latest version of PowerShell Editor Services.
- [ ] If this is a security issue, I have read the [security issue reporting guidance](https://github.com/PowerShell/vscode-powershell/blob/main/SECURITY.md).

### Summary

When you try starting PSES by calling `Start-EditorServices.ps1` with a [Microsoft.Extensions.Logging](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel) log level value, if it doesn't match one of the legacy log levels you get an error, e.g.:
```sh
19:22:57 ❯ pwsh -nol -nop -Command "& '$PWD/PowerShellEditorServices/Start-EditorServices.ps1' -BundledModulesPath '$PWD' -LogPath '$PWD/powershell_es.log' -SessionDetailsPath '$HOME/.cache/nvim/powershell_es.session.json' -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Information"
Start-EditorServices.ps1: Cannot validate argument on parameter 'LogLevel'. The argument "Information" does not belong to the set "Diagnostic,Verbose,Normal,Warning,Error" specified by the ValidateSet attribute. Supply an argument that is in the set
and then try the command again.
```
As you can see, it's expecting one of the legacy log levels:  `Diagnostic`, `Verbose`, `Normal`, `Warning`, `Error`, while the MEL log level defines: `Trace`, `Debug`, `Information`, `Warning`, `Error`, `Critical`, `None`.

[...]

PR #2200, opened 2024-11-15
Enhanced Logging

JustinGrote:
- Align PsesLogLevel to MEL log levels
- Adjust some logging verbosity levels for the PSES startup logging flow to move better.
- Create a new LSP Log Debug Provider to get around limitations with the CSharp Omnisharp one.
- Drop all Omnisharp logs to trace since they are just LSP related
- Address LSP bugs by wiring up log levels in-band.

Should be merged with https://github.com/PowerShell/vscode-powershell/pull/5065

PR #2236, opened 2025-06-21
Add missing MEL log levels to the `LogLevel` validation (#2235)

eitamal:
# Summary

This PR adds `Trace`, `Debug`, `Information`, `Critical`, and `None` to the validation set on the `LogLevel` parameter when starting PSES using `Start-EditorServices.ps1` to allow for the adoption of the newer MEL log levels (#2200), closing #2235.

## Context

The `ValidationSet` defined for `LogLevel` here: [Start-EditorServices.ps1:49](https://github.com/PowerShell/PowerShellEditorServices/blob/3a429eac5544fd7613c281c0814f94ac03fad551/module/PowerShellEditorServices/Start-EditorServices.ps1#L49) was left out during the work done to enhance logging, meaning that the only MEL levels you can practically use are the ones that are the same as the legacy log levels. I chose to add the missing ones as opposed to replacing the old with the new for compatibility reasons which matched the sentiment [conveyed in the code](https://github.com/PowerShell/PowerShellEditorServices/blob/3a429eac5544fd7613c281c0814f94ac03fad551/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs#L268). I've included a bit more details in the issue I opened (#2235).

### DIFF
--- a/module/PowerShellEditorServices/Start-EditorServices.ps1
+++ b/module/PowerShellEditorServices/Start-EditorServices.ps1
@@ -107,5 +107,13 @@ param(
     $DebugServiceOutPipeName
 )
 
+#Translate legacy PSES log levels to MEL levels
+$LogLevel = switch ($LogLevel) {
+    'Diagnostic' { 'Trace' }
+    'Verbose' { 'Debug' }
+    'Normal' { 'Information' }
+    default { $LogLevel }
+}
+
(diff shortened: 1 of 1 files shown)


JustinGrote (approved):
Approved with MEL translation added

andyleejordan (approved):
Thanks!

Issue #2244, opened 2025-07-23
Support startDebugging DAP reverse request


jborean93:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been requested.

### Summary

I would like to spawn multiple debug launch configurations through a single debug session. This allows me to start a single PowerShell script that might launch separate runspaces or processes that I want to debug without manually having to setup the launch configuration JSON, going to the debug tag, selecting the configuration, then clicking run.

The [startDebugging request](https://microsoft.github.io/debug-adapter-protocol/specification#Reverse_Requests_StartDebugging) is a reverse request that can be sent by the DAP server to the DAP client to launch extra configuration options for the same debug type. This would allow the initial debugged script to be able to launch whatever debug session on demand rather than having to make the end user do all that manual work.

The scenario I have this in mind for is for a remote session where the remote runspace is configured with the script to run and will send back a `startDebugging` request for VSCode to attach to that runspace before continuing on.

### Proposed Design

My proposal is to implement this as a new function in PSES like the following:

```powershell
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

using namespace System.Collections
using namespace System.Threading
using namespace System.Threading.Tasks

function Start-NewDebugSession {
    <#
    .EXTERNALHELP ..\PowerShellEditorServices.Commands-help.xml
    #>
    [OutputType([Task])]
    [CmdletBinding()]
    param(
        [Parameter(Mandatory)]
        [ValidateSet('Attach', 'Launch')]
        [string]
        $Request,

        [Parameter(Mandatory)]
        [IDictionary]
        $Configuration
    )

    # Var will be set by PSES in configurationDone before launching script
    if (-not (Get-Variable -Name __psEditorServices_DebugServer -ErrorAction Ignore)) {
        $err = [ErrorRecord]::new(
            [Exception]::new("Cannot start a new debug session unless running in an existing debug session"),
            "NotUnderDebugSession",
            [ErrorCategory]::InvalidOperation,
            $null)
        $err.ErrorDetails = [ErrorDetails]::new("")
        $err.ErrorDetails.RecommendedAction = 'Launch script with debugging'
        $PSCmdlet.WriteError($err)
        return
    }

    $config = $configuration.Clone()
    # Must be true or else a deadlock occurs trying to launch
[...]

PR #2249, opened 2025-07-25
Add Start-DebugAttachSession function

jborean93:
# PR Summary
Adds the `Start-DebugAttachSession` cmdlet which can be used to launch a new attach session debug request inside an existing launched debug session. This allows a script being debugged to launch a child debugging session using more dynamic environment information without requiring end users to manually setup the launch.json configuration.

## PR Context

Fixes: https://github.com/PowerShell/PowerShellEditorServices/issues/2244

Examples added with: https://github.com/PowerShell/vscode-powershell/pull/5246

image

### DIFF
--- a/test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs
+++ b/test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerShell.EditorServices.Handlers;
 using Nerdbank.Streams;
@@ -54,11 +55,15 @@ namespace PowerShellEditorServices.Test.E2E
         /// </summary>
         private Task<StoppedEvent> nextStopped => nextStoppedTcs.Task;
 
-        private readonly TaskCompletionSource<StartDebuggingAttachRequestArguments> startDebuggingAttachRequestTcs = new();
         /// <summary>
         /// This task is useful for waiting until a StartDebuggingAttachRequest is received.
         /// </summary>
-        private Task<StartDebuggingAttachRequestArguments> startDebuggingAttachRequest => startDebuggingAttachRequestTcs.Task;
+        private readonly TaskCompletionSource<StartDebuggingAttachRequestArguments> startDebuggingAttachRequestTcs = new();
+
+        /// <summary>
+        /// This task is useful for waiting until the debug session has terminated.
+        /// </summary>
+        private readonly TaskCompletionSource<TerminatedEvent> terminatedTcs = new();
 
         public async Task InitializeAsync()
         {
@@ -110,6 +115,11 @@ namespace PowerShellEditorServices.Test.E2E
                     .OnRequest("startDebugging", (StartDebuggingAttachRequestArguments request) =>
                     {
                         startDebuggingAttachRequestTcs.SetResult(request);
+                        return Task.CompletedTask;
+                    })
+                    .OnTerminated((TerminatedEvent e) =>
+                    {
+                        terminatedTcs.SetResult(e);
                         return Task.CompletedTask;
                     })
                 ;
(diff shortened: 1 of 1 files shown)


module/PowerShellEditorServices/Commands/Public/Start-DebugAttachSession.ps1
```diff
@@ -0,0 +1,174 @@
...
+        }
+        else {
+            $configuration.name = 'Attach Session'
+        }
+
+        if ($ComputerName) {
+            $configuration.computerName = $ComputerName
+        }
+
+        if ($RunspaceId) {
```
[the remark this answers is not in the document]
jborean93: 0 isn't a valid Runspace ID but I'll update to make it clearer.

module/PowerShellEditorServices/Commands/Public/Start-DebugAttachSession.ps1
```diff
@@ -0,0 +1,174 @@
...
+
+    try {
+        if ($PSBoundParameters.ContainsKey('RunspaceId') -and $RunspaceName) {
+            $err = [ErrorRecord]::new(
+                [ArgumentException]::new("Cannot specify both RunspaceId and RunspaceName parameters"),
+                "InvalidRunspaceParameters",
+                [ErrorCategory]::InvalidArgument,
+                $null)
+            $err.ErrorDetails = [ErrorDetails]::new("")
+            $err.ErrorDetails.RecommendedAction = 'Specify only one of RunspaceId or RunspaceName.'
```
JustinGrote, line 56: *chefs kiss on recommendedaction*

module/PowerShellEditorServices/Commands/Public/Start-DebugAttachSession.ps1
```diff
@@ -0,0 +1,174 @@
...
+        if ($PSBoundParameters.ContainsKey('RunspaceId')) {
+            $configuration.runspaceId = $RunspaceId
+        }
+        elseif ($RunspaceName) {
+            $configuration.runspaceName = $RunspaceName
+        }
+
+        # https://microsoft.github.io/debug-adapter-protocol/specification#Reverse_Requests_StartDebugging
+        $resp = $debugServer.SendRequest(
+            "startDebugging",
```
JustinGrote, line 132: nit: single quotes

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -178,6 +178,23 @@ public async Task<LaunchResponse> Handle(PsesLaunchRequestArguments request, Can
                 }
 
                 _logger.LogTrace("Working dir " + (string.IsNullOrEmpty(workingDir) ? "not set." : $"set to '{workingDir}'"));
+
+                if (!request.CreateTemporaryIntegratedConsole)
+                {
+                    // Start-DebugAttachSession attaches in a new temp console
+                    // so we cannot set this var if already running in that
+                    // console.
+                    PSCommand setVariableCmd = new PSCommand().AddCommand("Set-Variable")
```
JustinGrote, line 187: Nice way to pass this in.

module/PowerShellEditorServices/Commands/Public/Start-DebugAttachSession.ps1
jborean93: Gotta play to the crowd :) It has made me notice that it's awkward to set it in PowerShell outside of `Write-Error`. Might be something to look into some more.

JustinGrote (approved):
LGTM for the most part, this will be a very useful addtiion!

- doing detach in the runspace creates an unhandled exception, we should probably handle this nicer if at all possible but it might not be.

- having tests for this would be ideal, but scaffolding it may be difficult. It should be doable without needing to E2E it via the vscode-powershell repo. I'll think on it.

I wonder if there's any way we can *hook* into child pwsh process starts from an active debug session and do "smart" attach similar to what node does, this way you could debug child processes without having to make code changes. That would be a separate PR though.

In any case, this is more or less nonexistent/broken functionality so this isn't a risky thing to merge.

module/PowerShellEditorServices/Commands/Public/Start-DebugAttachSession.ps1
jborean93, line 132:
```suggestion
            'startDebugging',
```

jborean93:
> I think the difficult bit will be that Omnisharp doesn't actually know about startDebugging so I'm unsure if there is going to be any way to "expect" the request on the test side.

Looks like there is a way to wait for any request by name in the test runner so it should be possible to test this out. I'll probably want to look at adding it in in a separate PR though as the tests in my various PRs are starting to rely on each other or repeat the same things.

PR #2250, opened 2025-07-28
Delay launch/attach until after config done

### LINKED ISSUE #2191: Bring launch sequencing in line with DAP sec
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been requested.

### Summary

Currently, our DAP conversation doesn't follow the spec explicitly in terms of conversation.

What is supposed to happen:
1. Initialize is sent and server responds, then responds with capabilities
1. Launch is sent by client after receiving capabilities. **should be held and deferred until configurationDone**
1. Server sends initialized event (preferably when ready to receive launch requests)
3. Client does config after initialized 
4. Configuration is done (breakpoints etc.)
5. launch request is now responded to after configuration is completed

In actuality, we respond to the launch request immediately before configuration
image

The spec states:
```
Once the debug adapter is ready to receive configuration from the client, it sends an initialized event to the client. As described above, the client sends zero or more configuration-related requests before sending a configurationDone request.

After the response to configurationDone is sent, the debug adapter may respond to the launch or attach request, and then the debug session has started.
```

### Proposed Design

We need to hold the launch response until after configurationDone is completed. Fixing this would likely be part of a larger refactor where the jobs are started within configurationDone, rather than configurationDone triggering an event or otherwise to start/release any pending launch messages.

### LINKED ISSUE #2245: Write event on configurationDone for attach requests
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been requested.

### Summary

On attach scenarios, it is not possible to wait for all the breakpoints set by the client to be set in the remote runspace. A client could have something like

```powershell
$originalRunspaces = Get-Runspace

Read-Host -Prompt "Please attach to process $pid with runspace $([Runspce]::DefaultRunspace.Id)"

# Waits for VSCode to attach to runspace, an extra runspace will be created by
# Enter-PSHostProcess
while ($true) {
    $newRunspace = Get-Runspace | Where-Object { $_.Id -notin $runspaces.Id } | Select-Object -First 1
    if ($newRunspace) {
        break
    }
    Start-Sleep -Milliseconds 300
}

& .\script.ps1
```

But this only checks that PSES has attached to the named pipe. It won't wait for PSES to have called `Debug-Runspace` on the target runspace and more importantly any breakpoints inside `script.ps1` to be set. It's also error prone if you cannot control when the attach actually happens as the additional runspace check could have occurred before the first `Get-Runspace` call.

What would be ideal is for a well known event to be created by PSES when `configurationDone` was received so that the client only had to do

```powershell
# Attach to this runspace somehow

$e = Wait-Event -SourceIdentifier PSES.ConfigurationDone -Timeout 30
if ($e) {
    $e | Remove-Event
}
else {
    throw "Timed out waiting for attach event"
}
```

### Proposed Design

My proposal requires some changes to PowerShell itself but I wanted to see if the general idea was acceptable before trying to propose the change there. The benefit of using an event is that no custom functions needed to be loaded, and subsequently checked by the code, and `Wait-Event` has a simple timeout operation.

Ultimately what needs to happen is for PowerShell to expose a `__New-Event` command, similar to the new `__Set-PSBreakpoint` commands, that is treated specially by the PSRemoting server. This would be able to do the following but without having to spin up a new pipeline.

```
$runspace = Get-Runspace -Id $runspaceId
$runspace.Events.GenerateEvent($SourceId, $Sender, $EventArgs, $EventData)
```

We cannot just have PSES run this script or call `New-Event` directly as the runspace pipeline is busy running `Debug-Runspace`. The only time it can run commands is if the caller ran `Wait-Debugger` which is not ideal as it will pause the execution and require the client to continue on manually.

An alternative that I haven't tried out yet is to delay `Debug-Runspace` until after `configurationDone` is sent. This would allow both the breakpoints and event to be manually emited as commands on the runspace specified without having to block the pipeline with `Debug-Runspace`. I am not sure if that will be possible though but it could be an option that doesn't require any changes to PowerShell itself.

jborean93:
# PR Summary
Updates the sequence of launching and attaching scripts until after
configuration done is received. This aligns the behviour of starting a
debug session with other debug adapters. A side effect of this change is
that it is now possible to set breakpoints and do other actions in an
attached runspace target during the initialisation.

Also adds a new option `NotifyOnAttach` for an attach request which will
create the `PSES.Attached` event before calling `Debug-Runspace`
allowing attached scripts to wait for an attach event.

## PR Context
Fixes: https://github.com/PowerShell/PowerShellEditorServices/issues/2191
Fixes: https://github.com/PowerShell/PowerShellEditorServices/issues/2245

The changes include the commit from https://github.com/PowerShell/PowerShellEditorServices/pull/2251 which includes some test runner changes and to verify all the recent stuff added by myself work properly. I'll take this out of draft once we know what is happening with #2251.

PR to document new option in `vscode-powershell` https://github.com/PowerShell/vscode-powershell/pull/5249.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/BreakpointService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/BreakpointService.cs
@@ -18,6 +18,154 @@ namespace Microsoft.PowerShell.EditorServices.Services
 {
     internal class BreakpointService
     {
+        private const string _getPSBreakpointLegacy = @"
+            [CmdletBinding()]
+            param (
+                [Parameter()]
+                [string]
+                $Script,
+
+                [Parameter()]
+                [int]
+                $RunspaceId = [Runspace]::DefaultRunspace.Id
+            )
+
+            $runspace = if ($PSBoundParameters.ContainsKey('RunspaceId')) {
+                Get-Runspace -Id $RunspaceId
+                $null = $PSBoundParameters.Remove('RunspaceId')
+            }
+            else {
+                [Runspace]::DefaultRunspace
+            }
+
+            $debugger = $runspace.Debugger
+            $getBreakpointsMeth = $debugger.GetType().GetMethod(
+                'GetBreakpoints',
+                [System.Reflection.BindingFlags]'NonPublic, Public, Instance',
+                $null,
+                [type[]]@(),
+                $null)
+
+            $runspaceIdProp = [System.Management.Automation.PSNoteProperty]::new(
+                'RunspaceId',
+                $runspaceId)
+
+            @(
+                if (-not $getBreakpointsMeth) {
+                    if ($RunspaceId -ne [Runspace]::DefaultRunspace.Id) {
+                        throw 'Failed to find GetBreakpoints method on Debugger.'
+                    }
+
+                    Microsoft.PowerShell.Utility\Get-PSBreakpoint @PSBoundParameters
+                }
+                else {
+                    $getBreakpointsMeth.Invoke($debugger, @()) | Where-Object {
+                        if ($Script) {
+                            $_.Script -eq $Script
+                        }
+                        else {
+                            $true
+                        }
+                    }
+                }
+            ) | ForEach-Object {
+                $_.PSObject.Properties.Add($runspaceIdProp)
+                $_
+            }
+        ";
+
+        private const string _removePSBreakpointLegacy = @"
+            [CmdletBinding(DefaultParameterSetName = 'Breakpoint')]
+            param (
+                [Parameter(Mandatory, ValueFromPipeline, ParameterSetName = 'Breakpoint')]
+                [System.Management.Automation.Breakpoint[]]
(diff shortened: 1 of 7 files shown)


jborean93:
Also have opened https://github.com/PowerShell/PowerShell/pull/25788 which adds an event and public API to determine when `Debug-Runspace` has attached itself to a Runspace. This will help provide an official public API for new PowerShell versions if attached but the proposed `notifyAttach` option here will be good for older PowerShell versions.

JustinGrote (commented):
@copilot rebase this to main and resolve merge conflicts, favoring the main branch and summarizing conflicts and the reasoning for how it was resolved.

Issue #2242, opened 2025-07-23
Path Mapping on Attach request


jborean93:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all [issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been requested.

### Summary

When debugging a remote runspace the current logic is for PSES to retrieve back a temporary file of anything hit in a breakpoint and display it VSCode. This can be prone to some bugs and makes debugging a runspace that is not local a lot more difficult than it has to be.

Other languages work around this issue by offering an option to map local path(s) to their remote equivalent and vice versa.

For example .NET's DAP launch configuration has an option called [sourceFileMap](https://code.visualstudio.com/docs/csharp/debugger-settings#_source-file-map). You can do something like

```json
{
    "sourceFileMap": {
        "C:\\remote\\path": "/home/local/path"
    }
}
```

Python is similar but calls the option [pathMappings](https://code.visualstudio.com/docs/python/debugging#_remote-script-debugging-with-ssh) and uses a different structure. For example:

```json
{
    "pathMappings": [
        {
            "localRoot": "/home/local/path",
            "remoteRoot": "C:\\remote\\path"
        }
    ]
}
```

I would like to extend the attach request in PSES to support a way to map these local <-> remote paths so we can avoid all the psedit and file copying problems.

### Proposed Design

I'm more in favour of the `pathMappings` option that Python uses as the C# one is more geared around to source compilation paths than actual local/remote translations. It is slightly more verbose but it is easier to see what is the local and what is the remote.

As for the implementation there are a few locations that need to do the translation:

+ [SetLineBreakpointsAsync](https://github.com/PowerShell/PowerShellEditorServices/blob/6bb322ec2150ed341991ffc7db7180db7a49539d/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs#L129-L173) when translating the local to remote breakpoint path
+ [FetchStackFramesAsync](https://github.com/PowerShell/PowerShellEditorServices/blob/6bb322ec2150ed341991ffc7db7180db7a49539d/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs#L779-L892) when translating the remote stack frame file path to local
[...]

PR #2251, opened 2025-07-31
Add `pathMappings` option to debugger

jborean93:
# PR Summary

Adds the `pathMappings` option to the debugger that can be used to map a local to remote path and vice versa. This is useful if the local environment has a checkout of the files being run on a remote target but at a different path. The mappings are used to translate the paths that will the breakpoint will be set to in the target PowerShell instance. It is also used to update the stack trace paths received from the remote.

For a launch scenario, the path mappings are also used when launching a script if the integrated terminal has entered a remote runspace.

## PR Context

Fixes: https://github.com/PowerShell/PowerShellEditorServices/issues/2242

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -48,7 +48,6 @@ namespace Microsoft.PowerShell.EditorServices.Services
         private VariableContainerDetails scriptScopeVariables;
         private VariableContainerDetails localScopeVariables;
         private StackFrameDetails[] stackFrameDetails;
-        private PathMapping[] _pathMappings;
 
         private readonly SemaphoreSlim debugInfoHandle = AsyncUtils.CreateSimpleLockingSemaphore();
         #endregion
@@ -86,6 +85,11 @@ namespace Microsoft.PowerShell.EditorServices.Services
             get => _debugContext.IsDebuggingRemoteRunspace;
             set => _debugContext.IsDebuggingRemoteRunspace = value;
         }
+
+        /// <summary>
+        /// Gets or sets an array of path mappings for the current debug session.
+        /// </summary>
+        public PathMapping[] PathMappings { get; set; } = [];
 
         #endregion
 
@@ -603,27 +607,20 @@ namespace Microsoft.PowerShell.EditorServices.Services
             };
         }
 
-        internal void SetPathMappings(PathMapping[] pathMappings) => _pathMappings = pathMappings;
-
-        internal void UnsetPathMappings() => _pathMappings = null;
-
         internal bool TryGetMappedLocalPath(string remotePath, out string localPath)
         {
-            if (_pathMappings is not null)
+            foreach (PathMapping mapping in PathMappings)
             {
-                foreach (PathMapping mapping in _pathMappings)
+                if (string.IsNullOrWhiteSpace(mapping.LocalRoot) || string.IsNullOrWhiteSpace(mapping.RemoteRoot))
                 {
-                    if (string.IsNullOrWhiteSpace(mapping.LocalRoot) || string.IsNullOrWhiteSpace(mapping.RemoteRoot))
-                    {
-                        // If either path mapping is null, we can't map the path.
-                        continue;
-                    }
+                    // If either path mapping is null, we can't map the path.
+                    continue;
+                }
 
-                    if (remotePath.StartsWith(mapping.RemoteRoot, StringComparison.OrdinalIgnoreCase))
-                    {
-                        localPath = mapping.LocalRoot + remotePath.Substring(mapping.RemoteRoot.Length);
-                        return true;
-                    }
+                if (remotePath.StartsWith(mapping.RemoteRoot, StringComparison.OrdinalIgnoreCase))
+                {
+                    localPath = mapping.LocalRoot + remotePath.Substring(mapping.RemoteRoot.Length);
+                    return true;
                 }
             }
 
@@ -633,22 +630,19 @@ namespace Microsoft.PowerShell.EditorServices.Services
 
         internal bool TryGetMappedRemotePath(string localPath, out string remotePath)
         {
-            if (_pathMappings is not null)
+            foreach (PathMapping mapping in PathMappings)
             {
-                foreach (PathMapping mapping in _pathMappings)
+                if (string.IsNullOrWhiteSpace(mapping.LocalRoot) || string.IsNullOrWhiteSpace(mapping.RemoteRoot))
                 {
-                    if (string.IsNullOrWhiteSpace(mapping.LocalRoot) || string.IsNullOrWhiteSpace(mapping.RemoteRoot))
-                    {
-                        // If either path mapping is null, we can't map the path.
-                        continue;
-                    }
+                    // If either path mapping is null, we can't map the path.
+                    continue;
+                }
 
-                    if (localPath.StartsWith(mapping.LocalRoot, StringComparison.OrdinalIgnoreCase))
-                    {
-                        // If the local path starts with the local path mapping, we can replace it with the remote path.
-                        remotePath = mapping.RemoteRoot + localPath.Substring(mapping.LocalRoot.Length);
-                        return true;
-                    }
+                if (localPath.StartsWith(mapping.LocalRoot, StringComparison.OrdinalIgnoreCase))
+                {
+                    // If the local path starts with the local path mapping, we can replace it with the remote path.
+                    remotePath = mapping.RemoteRoot + localPath.Substring(mapping.LocalRoot.Length);
+                    return true;
                 }
             }
 
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -15,7 +15,6 @@ using Microsoft.PowerShell.EditorServices.Services.PowerShell.Debugging;
 using Microsoft.PowerShell.EditorServices.Services.PowerShell.Execution;
 using Microsoft.PowerShell.EditorServices.Services.PowerShell.Host;
 using Microsoft.PowerShell.EditorServices.Services.PowerShell.Utility;
-using Microsoft.PowerShell.EditorServices.Services.TextDocument;
 using Microsoft.PowerShell.EditorServices.Utility;
 
 namespace Microsoft.PowerShell.EditorServices.Services
@@ -49,6 +48,7 @@ namespace Microsoft.PowerShell.EditorServices.Services
         private VariableContainerDetails scriptScopeVariables;
         private VariableContainerDetails localScopeVariables;
         private StackFrameDetails[] stackFrameDetails;
+        private PathMapping[] _pathMappings;
 
         private readonly SemaphoreSlim debugInfoHandle = AsyncUtils.CreateSimpleLockingSemaphore();
         #endregion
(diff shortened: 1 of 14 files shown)


jborean93:
An example of how the `pathMappings` can be used for launching when the interactive terminal has entered a remote runspace. The editor sees the file under `/home/jborean/dev/vscode-powershell/examples/test.ps1` but the `pathMappings` will map it to `C:\temp\test.ps1` where that same script is located on the remote host.

```json
{
  "name": "PowerShell Launch Script",
  "type": "PowerShell",
  "request": "launch",
  "script": "${workspaceFolder}/test.ps1",
  "cwd": "${cwd}",
  "pathMappings": [
    {
      "localRoot": "${workspaceFolder}/",
      "remoteRoot": "C:\\temp\\"
    }
  ]
},
```

image

It looks like I need to update the logic for the client breakpoint to contain the original workspace path and not the mapped remote file so I will fix that.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -603,6 +603,59 @@ public VariableScope[] GetVariableScopes(int stackFrameId)
             };
         }
 
+        internal void SetPathMappings(PathMapping[] pathMappings) => _pathMappings = pathMappings;
+
+        internal void UnsetPathMappings() => _pathMappings = null;
```
andyleejordan, line 608: Could this just be a property?

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -603,6 +603,59 @@ public VariableScope[] GetVariableScopes(int stackFrameId)
...
+
+        internal bool TryGetMappedLocalPath(string remotePath, out string localPath)
+        {
+            if (_pathMappings is not null)
```
andyleejordan, line 612: If _pathMappings is initialized to an empty dictionary then foreach over it in that scenario is a no-op and we don't have to do this check (and nested statement).

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
```diff
@@ -603,6 +603,59 @@ public VariableScope[] GetVariableScopes(int stackFrameId)
...
+                {
+                    if (string.IsNullOrWhiteSpace(mapping.LocalRoot) || string.IsNullOrWhiteSpace(mapping.RemoteRoot))
+                    {
+                        // If either path mapping is null, we can't map the path.
+                        continue;
+                    }
+
+                    if (remotePath.StartsWith(mapping.RemoteRoot, StringComparison.OrdinalIgnoreCase))
+                    {
+                        localPath = mapping.LocalRoot + remotePath.Substring(mapping.RemoteRoot.Length);
```
andyleejordan, line 624: It does seem to me like the mapping would also be a directory (which may or may not be supplied with a trailing path separator); is there a good reason _not_ to use path.combine?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -128,6 +139,20 @@ public LaunchAndAttachHandler(
...
+        {
+            _debugService.SetPathMappings(request.PathMappings);
+            try
+            {
+                return await HandleImpl(request, cancellationToken).ConfigureAwait(false);
+            }
+            catch
+            {
+                _debugService.UnsetPathMappings();
+                throw;
```
andyleejordan, line 151: I feel like there's a better pattern to apply here, but I'm not sure what. @SeeminglyScience for a suggestion (is there a way this can be done with a `using`? Is that too complicated?)

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -222,10 +247,19 @@ await _executionService.ExecutePSCommandAsync(
...
-                        _runspaceContext.CurrentRunspace);
+                if (_debugService.TryGetMappedRemotePath(_debugStateService.ScriptToLaunch, out string remoteMappedPath))
+                {
+                    _debugStateService.ScriptToLaunch = remoteMappedPath;
+                }
+                else
+                {
+                    // If the script is not mapped, we will map it to the remote path
+                    // using the RemoteFileManagerService.
+                    _debugStateService.ScriptToLaunch =
```
andyleejordan, line 258: Can you simplify this so the assignment to ScriptToLaunch happens only once, with the if/else covering the two logical paths to get the (err) path?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -250,11 +284,13 @@ public async Task<AttachResponse> Handle(PsesAttachRequestArguments request, Can
...
             try
             {
+                _debugService.SetPathMappings(request.PathMappings);
                 return await HandleImpl(request, cancellationToken).ConfigureAwait(false);
             }
             catch
             {
                 _debugService.IsDebuggingRemoteRunspace = false;
+                _debugService.UnsetPathMappings();
                 throw;
```
andyleejordan, line 294: Ohh I see where that pattern came from now.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
```diff
@@ -486,6 +522,7 @@ private async Task OnExecutionCompletedAsync(Task executeTask)
             _debugEventHandlerService.UnregisterEventHandlers();
 
             _debugService.IsDebuggingRemoteRunspace = false;
+            _debugService.UnsetPathMappings();
```
andyleejordan, line 525: This is probably not actually something to be solved in this PR, it seems like DebugService needs some sort of finalizer to handle "hey we're wrapping up now." Surprised it's not just disposed and recreated but I'd have to dig in to understand that more.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/StackTraceHandler.cs
```diff
@@ -38,7 +38,12 @@ public async Task<StackTraceResponse> Handle(StackTraceArguments request, Cancel
         InvocationInfo invocationInfo = debugService.CurrentDebuggerStoppedEventArgs?.OriginalEvent?.InvocationInfo
             ?? throw new RpcErrorException(0, null!, "InvocationInfo was not available on CurrentDebuggerStoppedEvent args. This is a bug and you should report it.");
 
-        StackFrame breakpointLabel = CreateBreakpointLabel(invocationInfo);
+        string? scriptNameOverride = null;
+        if (debugService.TryGetMappedLocalPath(invocationInfo.ScriptName, out string mappedLocalPath))
+        {
+            scriptNameOverride = mappedLocalPath;
+        }
+        StackFrame breakpointLabel = CreateBreakpointLabel(invocationInfo, scriptNameOverride: scriptNameOverride);
```
andyleejordan, line 46: This is how I think the other assignment should be done. Much more clear.

src/PowerShellEditorServices/Services/PowerShell/Debugging/DscBreakpointCapability.cs
```diff
@@ -91,10 +91,12 @@ public static async Task<DscBreakpointCapability> GetDscCapabilityAsync(
...
                     .AddParameter("PassThru")
-                    .AddParameter("ErrorAction", ActionPreference.Ignore);
+                    .AddParameter("ErrorAction", ActionPreference.Ignore)
+                    .AddCommand(@"Microsoft.PowerShell.Utility\Select-Object")
+                    .AddParameter("ExpandProperty", "Name");
 
-                IReadOnlyList<PSModuleInfo> dscModule =
-                    await executionService.ExecutePSCommandAsync<PSModuleInfo>(
+                IReadOnlyList<string> dscModule =
+                    await executionService.ExecutePSCommandAsync<string>(
```
andyleejordan, line 99: Ok so these are now strings since we added the parameter to get the names instead of the modules directly. Why, what are we changing here? FWIW this is really old legacy code.

test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs
```diff
@@ -105,12 +107,18 @@ send until a launch is sent.
...
+                    })
                     // This TCS is useful to wait for a breakpoint to be hit
-                    .OnStopped(async (StoppedEvent e) =>
+                    .OnStopped((StoppedEvent e) =>
                     {
-                        nextStoppedTcs.SetResult(e);
+                        TaskCompletionSource<StoppedEvent> currentStoppedTcs = nextStoppedTcs;
                         nextStoppedTcs = new();
+
+                        currentStoppedTcs.SetResult(e);
```
andyleejordan, line 121: I feel like this must fix something that's been broken/buggy with the tests. Can you elaborate?

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
SeeminglyScience: @jborean93 should this be a `finally` instead of a `catch`? or are the path mappings unset somewhere else when there's no error? I guess that would have to be the case right? `Handle` returns before debugging is actually complete right?

(if the answer to the above is "no it needs to be a catch", then @andyleejordan: nah `using` wouldn't work here unfortunately as it's only unset when an exception occurs)

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
jborean93: It can, I think I started with just a normal property then went with a method when it seemed like I wanted to set some other stuff but that went away. Can definitely just use a property here if you wish.

src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs
jborean93: Yea it's done in the catch block to undo any setup work for the session in case of a fatal failure. The normal cleanup will happen in the terminal/disconnect handler which occurs when the debug session ends.

test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs
jborean93: It was a race condition I (think I) encountered. Consider the following code from one of the tests

```csharp
...
StoppedEvent stoppedEvent = await nextStoppedTask;
nextStoppedTask = nextStopped;
...

// Later in the test
stoppedEvent = await nextStoppedTask;
```

When the current code did

```csharp
nextStoppedTcs.SetResult(e);
nextStoppedTcs = new();
```

The `await` on the test side could have set `nextStoppedTask` from the `nextStopped` field before the `OnStopped` delegate created the new value. This meant the next time the test awaited that task it would get the old `StoppedEvent` from the original await. What this change does is to set the field first with a fresh TCS then complete the task so that when the tests return from the await the next `nextStoppedTask = nextStopped` will always be for a subsequent stop event.

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
jborean93: Root just means the root part of the path, this is how the python one works.If you prefer a different name I can certainly change it but:

+ I'm just working based on how the Python works and is named
+ I need to be able to specify just a name based on how Ansible works

test/PowerShellEditorServices.Test.E2E/DebugAdapterProtocolMessageTests.cs
andyleejordan: Nice, thanks!

src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
andyleejordan: Ok, that's reasonable given Python's precedent.

andyleejordan (approved):
Thanks for this!

PR #2262, opened 2025-11-25
Update .NET, PSReadLine, and Nuget packages

andyleejordan:
Seems like a test is hung locally, seeing if it repros in CI.


SeeminglyScience (approved):
LGTM!

PR #2303, opened 2026-06-12
Match strong-name identity when resolving PSES dependencies

andyleejordan:
## Summary

A focused **trial** of tightening how `PsesLoadContext` decides whether a candidate assembly can satisfy a dependency request. Patrick and I have suspected for a while that our version-only matching is inadequate; this proposes we trial requiring the full strong-name identity to match.

## Problem

On .NET Core, `PsesLoadContext.IsSatisfyingAssembly` gates whether a DLL in `$PSHOME` or our bundled `Common` directory can satisfy a dependency request, using only:

- the simple name (case-insensitive), and
- `candidate.Version >= required.Version`.

That ignores the rest of the assembly identity. A same-named assembly with a *different* public key token (i.e. a genuinely different assembly) was treated as a drop-in replacement, and the mismatch only surfaced later as a `FileLoadException`/`TypeLoadException` at bind time instead of being declined up front.

## Change

Also require:

- **Public key token** — if the reference is strong-named, the candidate's token must match exactly; a non-strong-named reference imposes no token requirement.
- **Culture** — must match, so we never substitute a satellite resource assembly for the neutral one (or vice versa).

The pure comparison moved into an `internal static` overload taking two `AssemblyName`s so it can be unit-tested without DLLs on disk.

## Why this is safe to trial

- The check can only return `false` in *more* cases, and only for assemblies that could not have satisfied the reference anyway.
- On a mismatch we decline to short-circuit and fall through to the default load context's own (laxer) resolution — equivalent to returning "not mine".
- Measured against a current build, **no presently-bundled dependency changes resolution** under the new rules (zero token mismatches across `Common` vs `$PSHOME`), so today this is purely added protection.

## Tests

`PsesLoadContextTests` (net8.0 / `CoreCLR` only, since the Hosting assembly is .NET Core only) covers exact match, newer/older version, name case-insensitivity, differing public key token, strong-named-vs-unsigned, no-required-token, and culture match/mismatch. All 10 pass; `net462` still compiles (reference and tests are guarded).

## Context / scope

Part of the broader ALC isolation investigation behind the recurring *"Assembly with same name is already loaded"* / *"Could not load file or assembly"* reports. This is the **resolver-correctness** piece only — it does not by itself address the feature-side eager-loading (completion / `Get-Help` importing user modules) or the Windows PowerShell *"no ALC at all"* class of issues.

## Open questions

- Whether to also tighten the `Version >=` rule (e.g. require major-version compatibility) — deliberately left as-is here.
- Whether the trial framing is enough to merge as-is, or if we'd rather gate it.

---
🤖 Drafted by Copilot (Claude Opus 4.8) for @andyleejordan to review and edit before merging.

### DIFF
--- a/src/PowerShellEditorServices.Hosting/Internal/PsesLoadContext.cs
+++ b/src/PowerShellEditorServices.Hosting/Internal/PsesLoadContext.cs
@@ -76,10 +76,72 @@ namespace Microsoft.PowerShell.EditorServices.Hosting
                 return false;
             }
 
-            AssemblyName asmToLoadName = AssemblyName.GetAssemblyName(assemblyPath);
+            return IsSatisfyingAssembly(requiredAssemblyName, AssemblyName.GetAssemblyName(assemblyPath));
+        }
 
-            return string.Equals(asmToLoadName.Name, requiredAssemblyName.Name, StringComparison.OrdinalIgnoreCase)
-                && asmToLoadName.Version >= requiredAssemblyName.Version;
+        // Internal (rather than private) purely so it can be unit tested with constructed
+        // AssemblyName instances; it has no file-system dependency of its own.
+        internal static bool IsSatisfyingAssembly(AssemblyName requiredAssemblyName, AssemblyName asmToLoadName)
+        {
+            // The simple name must match (case-insensitively, as assembly names are).
+            if (!string.Equals(asmToLoadName.Name, requiredAssemblyName.Name, StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            // The candidate must be at least the requested version. We still accept newer
+            // versions, since shared framework and $PSHOME assemblies are generally
+            // forward-compatible via the runtime's binding.
+            if (asmToLoadName.Version < requiredAssemblyName.Version)
+            {
+                return false;
+            }
+
+            // The strong-name identity must match. Previously only the simple name and version
+            // were compared, so a same-named assembly with a *different* public key token (i.e.
+            // a genuinely different assembly) was treated as a drop-in replacement and would then
+            // fail at runtime with a FileLoadException/TypeLoadException. Requiring the public key
+            // token to match means we only short-circuit to a $PSHOME/Common assembly that can
+            // actually satisfy the reference; otherwise we fall through and let the default load
+            // context resolve it with its own (laxer) rules.
+            if (!PublicKeyTokensMatch(requiredAssemblyName, asmToLoadName))
+            {
+                return false;
+            }
+
+            // The culture must match so we never substitute a satellite resource assembly for the
+            // neutral one (or vice versa).
+            return string.Equals(
+                asmToLoadName.CultureName ?? string.Empty,
+                requiredAssemblyName.CultureName ?? string.Empty,
+                StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool PublicKeyTokensMatch(AssemblyName requiredAssemblyName, AssemblyName candidateAssemblyName)
+        {
+            byte[] requiredToken = requiredAssemblyName.GetPublicKeyToken();
+
+            // A reference to a non-strong-named assembly imposes no public key token requirement.
+            if (requiredToken is null || requiredToken.Length == 0)
+            {
+                return true;
+            }
+
+            byte[] candidateToken = candidateAssemblyName.GetPublicKeyToken();
+            if (candidateToken is null || candidateToken.Length != requiredToken.Length)
+            {
+                return false;
+            }
+
+            for (int i = 0; i < requiredToken.Length; i++)
+            {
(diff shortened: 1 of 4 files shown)


JustinGrote:
Would this have any issues with different PowerShell versions providing different base assemblies with different strongnames/publickey than what we are trying to match? That'd be my only concern.

andyleejordan:
@JustinGrote that's a solid question worth testing...

andyleejordan:
@JustinGrote Good news: tested and it holds up. Short answer — no, this doesn't cause issues across PowerShell versions, because **a newer PowerShell changes the assembly _version_ but never the strong-name (public key token)**. The matcher accepts newer versions (`>=`) and the token stays identical, so there are no false rejections.

### How I tested
Built this branch, linked it into a sibling `vscode-powershell`, and ran the extension's full E2E suite (which starts a real PSES session) against each PowerShell, pointing the session at each binary via `powershell.powerShellAdditionalExePaths` + `powerShellDefaultVersion`. The 7.x binaries were downloaded from GitHub releases and extracted (not installed). Also did diagnostic-level direct launches to capture which ALC each assembly lands in.

| PowerShell | Runtime | Result |
|---|---|---|
| 7.4.16 | .NET 8 | E2E **112/112**, PSES connected |
| 7.5.7 | .NET 9 | E2E **112/112**, PSES connected |
| 7.6.2 | .NET 10 | E2E **112/112**, PSES connected |
| 5.1 | .NET Framework | loads cleanly (not affected — see below) |

Plus the new `PsesLoadContextTests` pass 10/10.

### The proof for your exact concern
Here's the *same* base assembly under .NET 8 vs .NET 10 — version bumps `8.0.0.0` → `10.0.0.0`, **public key token unchanged**:

```
# 7.4.16 (.NET 8) — StartEditorServices log
Loaded into load context "Default" ...: System.Net.Http, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

# 7.6.2 (.NET 10) — StartEditorServices log
Loaded into load context "Default" ...: System.Net.Http, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Loaded into load context "Default" ...: System.Text.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
```

The well-known tokens (`b03f5f7f11d50a3a`, `cc7b13ffcd2ddd51`, `31bf3856ad364e35`, `adb9793829ddae60`) are stable across .NET 8/9/10. PSES + its bundled deps go into the isolated context, `$PSHOME`/shared-framework assemblies resolve from Default — exactly as intended, with zero `FileLoad`/`TypeLoad` errors:

```
# 7.6.2 (.NET 10)
Loaded into load context "PsesLoadContext" ...: Microsoft.PowerShell.EditorServices, Version=4.6.0.0, ..., PublicKeyToken=null
Loaded into load context "PsesLoadContext" ...: Microsoft.Extensions.Logging.Abstractions, Version=10.0.0.0, ..., PublicKeyToken=adb9793829ddae60
PSES Startup Completed. Starting Language Server.
```

And the E2E run asserting the *actual* executable/version that answered, per version:

```
[crossver] PSES connected: version=7.4.16 arch=X64 exePath=...\pwsh\7.4.16\pwsh.exe   →  112 passing
[crossver] PSES connected: version=7.5.7  arch=X64 exePath=...\pwsh\7.5.7\pwsh.exe    →  112 passing
[crossver] PSES connected: version=7.6.2  arch=X64 exePath=C:\Program Files\PowerShell\7\pwsh.exe → 112 passing
```

### Why the public-key-token check is still the right call
The token comparison only rejects a *same-named but genuinely different* assembly (different publisher key) — and in that case falling through to PSES's own bundled copy is the correct, safer behavior (that's the bug this PR fixes). It never rejects a legitimate, forward-compatible `$PSHOME` assembly, since those keep the same token.

### Note on 5.1
`PsesLoadContext` is `#if CoreCLR`-only, so Windows PowerShell 5.1 isn't touched by this change at all — it still uses the net462 `AssemblyResolve` path. Confirmed with a direct launch: `{"status":"started", ... "powerShellVersion":"5.1.26100.8655"}`, deps resolved cleanly. (Side note: the headless `vscode-test` harness can't spin up a Windows PowerShell *terminal* session — an environment artifact unrelated to this change; pwsh 7.x works fine in the same harness.)

---
_Drafted by Copilot, reviewed by @andyleejordan._

andyleejordan:
@JustinGrote @SeeminglyScience I think this is worth landing into a preview and hope it alleviates _some_ of the assembly conflicts users run into.

JustinGrote (approved):
I agree, LGTM lets see if it breaks anything in preview. I tried it on a few different PS versions locally (5.1, 7.4.2, 7.5.3, 7.6.1, 7.6.2) without apparent issue.

PR #2306, opened 2026-06-16
Add `FileContext.GetText(IFileRange)` overload

### LINKED ISSUE #1496: $Context.CurrentFile.GetText not compatible with $Context.SelectedRange
In the Powershell 5.1 version, the following code do not work: 

```
$Context = $psEditor.GetEditorContext()
$Context.CurrentFile.GetText($context.SelectedRange)

```
This seems to be a type error 
```
Cannot find an overload for "GetText" and the argument count: "1".
At line:1 char:1
+ $text = $Context.CurrentFile.GetText($context.SelectedRange)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodException
    + FullyQualifiedErrorId : MethodCountCouldNotFindBest
```

Solution is to do the following

```
    $Context = $psEditor.GetEditorContext()
    $Range = [Microsoft.PowerShell.EditorServices.Extensions.FileRange]::new($context.SelectedRange.Start,$context.SelectedRange.End)
    $Context.CurrentFile.GetText($Range)
```

I would expect the $Context.SelectedRange to work directly with the GetText method which take a FileRange as argument.

That does not end there. 
When working with Powershell 7.2, the **Microsoft.PowerShell.EditorServices.Extensions.FileRange** type does not exist so not only the way you'd expect it to work does not work but also the workaround of creating a new object of the aforementionned type manually and pasting value from the context range do not work either. 

The error in a PS 7,2 VSCode session  with GetText and $Context.SelectedRange 
```
MethodException: Cannot find an overload for "GetText" and the argument count: "1".
```

The error in a PS 7.2 VSCode session when trying to create a filerange object manually
```
InvalidOperation: Unable to find type [Microsoft.PowerShell.EditorServices.Extensions.FileRange].
```

andyleejordan:
Fixes #1496.

`EditorContext.SelectedRange` is typed as `IFileRange`, but `FileContext.GetText`/`GetTextLines` only accepted the concrete `FileRange` class. So the natural `$Context.CurrentFile.GetText($Context.SelectedRange)` failed overload resolution ("Cannot find an overload for GetText and the argument count: 1"), and the documented workaround of constructing a `FileRange` isn't reachable when that type isn't loaded.

This widens both parameters from `FileRange` to `IFileRange`. The bodies already route through the `ToBufferRange()` extension on `IFileRange`, so nothing else changes and existing `FileRange` callers keep working.

Adds focused regression tests covering `GetText`/`GetTextLines` with a `SelectedRange` and with a hand-built `FileRange`.

### DIFF
--- a/src/PowerShellEditorServices/Extensions/FileContext.cs
+++ b/src/PowerShellEditorServices/Extensions/FileContext.cs
@@ -104,7 +104,7 @@ namespace Microsoft.PowerShell.EditorServices.Extensions
         /// </summary>
         /// <param name="bufferRange">The buffer range for which content will be extracted.</param>
         /// <returns>A string with the specified range of content.</returns>
-        public string GetText(FileRange bufferRange)
+        public string GetText(IFileRange bufferRange)
         {
             return
                 string.Join(
@@ -123,7 +123,7 @@ namespace Microsoft.PowerShell.EditorServices.Extensions
         /// </summary>
         /// <param name="fileRange">The buffer range for which content will be extracted.</param>
         /// <returns>An array of strings, each representing a line in the file within the specified range.</returns>
-        public string[] GetTextLines(FileRange fileRange) => scriptFile.GetLinesInRange(fileRange.ToBufferRange());
+        public string[] GetTextLines(IFileRange fileRange) => scriptFile.GetLinesInRange(fileRange.ToBufferRange());
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


andyleejordan:
Software development really has changed. When this was first opened I said to the effect of "too busy to debug this, sorry" in https://github.com/PowerShell/PowerShellEditorServices/issues/1496#issuecomment-861002229 but today it just got automatically fixed when I threw GitHub Copilot across the repo to solve issues it thinks it could handle. And yes, this was an easy fix that was a typing problem (using a concrete type instead of the interface) and it's confirmed with a correct regression test.

andyleejordan:
@SeeminglyScience @JustinGrote this is definitely worth landing, it was a dumb type mismatch, easily fixed.

JustinGrote (approved):
LGTM, tested as working on my local instance.

PR #2309, opened 2026-06-16
Open hidden files in `Open-EditorFile`

### LINKED ISSUE #1750: Open-EditorFile doesn't work with hidden files
On MacOS (and I imagine Linux) dot files are hidden, and as such, are not displayed by `Get-ChildItem` which `Open-EditorFile` uses to get the file(s) to open. This makes using `psedit` as a shortcut to open such files impossible.

image

Initially I was going to suggest simply adding `-Force` to the `Get-ChildItem`, but the unintended consequence of that is that it will then open all the hidden files in a directory when given the directory. An alternative may be to include `-Force` on `Open-EditorFile` and pass that through to `Get-ChildItem`.

andyleejordan:
Fixes #1750.

On macOS and Linux, dot files are hidden, so `Get-ChildItem` skips them unless `-Force` is passed — meaning `psedit ~/.zshrc` (and similar) silently opened nothing. This adds `-Force` to the `Get-ChildItem` call in `Open-EditorFile` so hidden files resolve and open.

As the issue notes, this also means passing a *directory* will now enumerate its hidden files too. That is intentional: `psedit` exposes no switch to include or exclude hidden files, so the only predictable default is to always include them. Treating the directory case as "always include hidden" keeps the behavior consistent rather than special-casing it.

### DIFF
--- a/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
+++ b/module/PowerShellEditorServices/Commands/Public/CmdletInterface.ps1
@@ -195,7 +195,7 @@ function Open-EditorFile {
             $preview = $true
         }
 
-        Get-ChildItem $Paths -File | ForEach-Object {
+        Get-ChildItem $Paths -File -Force | ForEach-Object {
             $psEditor.Workspace.OpenFile($_.FullName, $preview)
         }
     }


andyleejordan:
@SeeminglyScience @JustinGrote ancient issue, one word fix, makes sense to me, no idea why we didn't do it sooner.

JustinGrote (approved):
Agreed, works on my machine.

PR #2311, opened 2026-06-16
Resolve evaluate/watch variables from the correct scope

### LINKED ISSUE #1882: Evaluate expression event returns variable value from wrong scope
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

```powershell
$myVarName = 123
& {
    $myVarName = 456
    $myVarName
}
```

While debugging the script above and stopped at line 4 (`$myVarName`), `evaluate` debugger call using `context: 'watch', expression: '$myVarName'` returns the value from parent scope (123).

Simply changing the expression to `$myVarName;` returns the local value (456). This points us in the direction of this method as source of the bug.
https://github.com/PowerShell/PowerShellEditorServices/blob/00238cdd8322241276b9691dd96f1f1b860bd1f9/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DebugEvaluateHandler.cs#L63-L64

The evaluate-call in my scenario is made through using `new vscode.InlineValueEvaluatableExpression(range, expression)` which VSCode executes [here](https://github.com/microsoft/vscode/blob/13429c254b99fdc0b1d28056d5fd5e43e95f5c07/src/vs/workbench/contrib/debug/browser/debugEditorContribution.ts#L691-L703)

### Expected behavior

```console
`evaluate` should return the local value (456). The local value is shown in Variable Explorer, PSIC and by inline using `vscode.InlineValueVariableLookup` while stopped at the same location.
```

### Actual behavior

```console
Returns 123.

After executing `$script:myVarName` or `$global:myVarName` in PSIC and using step over, `evaluate` returns the newest value (no scope-preference/-priority).

Changing `$myVarName` in PSIC + step over updates variable explorer + `vscode.InlineValueVariableLookup` result, but `evaluate` doesn't care - still showing the latest parent-scope value.
```

### Error details

_No response_

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.1.3
PSEdition                      Core
GitCommitId                    7.1.3
OS                             Linux 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

2022.7.2

### Visuals

_No response_

andyleejordan:
## Problem

Fixes #1882.

While stopped at a breakpoint, a watch / `evaluate` request for a naked variable could return the value from a *parent* scope instead of the current one. Given:

```powershell
$myVarName = 123
& {
    $myVarName = 456
    $myVarName   # break here
}
```

evaluating `$myVarName` returned `123` instead of `456`, even though the Variables explorer (and PowerShell itself) correctly show `456`.

## Cause

`DebugService.GetVariableFromExpression` searched the flat `variables` list with `FirstOrDefault`. That list is populated broadest-to-narrowest (global → script → local → stack frames), so the global/parent copy of a shadowed name was always matched first.

## Fix

Search the `Local`, `Script`, then `Global` scope containers in that order first, falling back to the full flat list so frame-only names still resolve. This matches PowerShell's own variable resolution and what the Variables explorer displays. Only the name lookup changed; ID-based lookups elsewhere are unaffected.

## Testing

- Added `DebuggerResolvesVariableFromMostLocalScope` plus a shared `VariableScopeTest.ps1` that shadows a variable across scopes and asserts the local value wins.
- Confirmed it fails without the fix (returns the parent value) and passes with it.
- All 36 `DebugServiceTests` pass on net8.0.

### DIFF
--- a/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
+++ b/src/PowerShellEditorServices/Services/DebugAdapter/DebugService.cs
@@ -308,7 +308,16 @@ namespace Microsoft.PowerShell.EditorServices.Services
             await debugInfoHandle.WaitAsync(cancellationToken).ConfigureAwait(false);
             try
             {
-                variableList = variables;
+                // Search the scope containers in order of narrowest to broadest scope (local,
+                // then script, then global) so that a variable defined in a more local scope
+                // correctly shadows one of the same name in a parent scope, matching
+                // PowerShell's own variable resolution. The flattened list of every variable is
+                // appended as a fallback so that names not present in those scopes (such as
+                // frame-specific variables) still resolve. See issue #1882.
+                variableList = new[] { localScopeVariables, scriptScopeVariables, globalScopeVariables }
+                    .Where(container => container is not null)
+                    .SelectMany(container => container.Children.Values)
+                    .Concat(variables);
             }
             finally
             {
--- a/test/PowerShellEditorServices.Test.Shared/Debugging/VariableScopeTest.ps1
+++ b/test/PowerShellEditorServices.Test.Shared/Debugging/VariableScopeTest.ps1
@@ -1,0 +1,5 @@
+$scopeTestVariable = "from parent scope"
+& {
+    $scopeTestVariable = "from local scope"
+    Write-Output $scopeTestVariable
+}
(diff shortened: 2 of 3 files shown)


andyleejordan:
I'm flabberghasted. That's pretty much the problem and a proper fix for it.

JustinGrote (approved):
Works for me when tested interactively using the example:




Watch has an interesting formatting difference for the further-out scoped variables, but that should be addressed in another issue.

PR #2312, opened 2026-06-16
Modernize alias expansion to use the language parser

### LINKED ISSUE #2108: Update `ExpandAliasHandler` to use latest PowerShell parser
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all [issues](https://github.com/PowerShell/PowerShellEditorServices/issues?q=is%3Aissue) to ensure it has not already been requested.

### Summary

In https://github.com/PowerShell/PowerShellEditorServices/blob/37f0648ca28358c1fd2d0688801658948f69090d/src/PowerShellEditorServices/Services/PowerShell/Handlers/ExpandAliasHandler.cs#L32-L33 we (repeatedly) define an internal PowerShell script that uses the original parser to expand the alias.

### Proposed Design

This should be updated to use the new parser directly from the PowerShell APIs in C# and return a proper edit (potentially deprecating the need for an entirely separate custom request).

andyleejordan:
Fixes #2108.

The `powerShell/expandAlias` handler built an embedded PowerShell here-string that called `[System.Management.Automation.PsParser]::Tokenize` to find command tokens. That legacy tokenizer is Windows PowerShell / full-framework only and carries its parsing limitations.

This rewrites the logic in C# against the modern, cross-edition APIs:

- Tokenize with `Parser.ParseInput` and select command-name tokens via `TokenFlags.CommandName` instead of the legacy `PSTokenType.Command`.
- Resolve every distinct name to its alias definition in a single `Get-Command -CommandType Alias` round-trip, rather than re-defining a helper function and querying one token at a time.
- Escape wildcard metacharacters with `WildcardPattern.Escape` so aliases whose names are patterns — `?` (Where-Object) and `%` (ForEach-Object) — resolve to themselves. The old script did this by hand with a leading backtick.
- Rebuild the text by substituting from the highest offset down so earlier extents stay valid as the length changes, preserving behavior for multiple aliases on a line and multi-token definitions.

Scripts with no aliases are returned unchanged, and the JSON-RPC contract (`IExpandAliasHandler` / `powerShell/expandAlias` / `ExpandAliasParams.Text` → `ExpandAliasResult.Text`) is untouched.

### Testing

Extended the E2E tests with a multi-alias line that includes the wildcard-named `?` alias — `gci | ? Name | % Name` → `Get-ChildItem | Where-Object Name | ForEach-Object Name` — which the naive modern approach would mishandle. Both ExpandAlias E2E tests pass.

### DIFF
--- a/src/PowerShellEditorServices/Services/PowerShell/Handlers/ExpandAliasHandler.cs
+++ b/src/PowerShellEditorServices/Services/PowerShell/Handlers/ExpandAliasHandler.cs
@@ -1,7 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Management.Automation;
+using System.Management.Automation.Language;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using MediatR;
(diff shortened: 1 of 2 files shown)


andyleejordan:
Well that's an old TODO solved and the tests pass so...

JustinGrote (approved):
Yeah looks fine to me. As a follow-on issue we should probably update the rename handler to use this same code path if it isn't already.

PR #2314, opened 2026-06-16
Fix flaky `CanRunOnIdleTask` by polling instead of sleeping

andyleejordan:
## Summary

`CanRunOnIdleTask` (and its twin `CanRunOnIdleInProfileTask`) in
`PsesInternalHostTests.cs` were intermittently failing on the **net462 (Windows
PowerShell 5.1)** CI leg. `CanRunOnIdleTask` was just caught failing on PR #2298's
Windows job (`Assert.Collection() … Assert.False/Assert.True` at line ~200). Both
tests carried a `// TODO: Why is this racy?` above a hard-coded `Thread.Sleep(2000)`.

## Root cause

`PsesInternalHost.OnPowerShellIdle` calls
`_mainRunspaceEngineIntrinsics.Events.GenerateEvent(PSEngineEvent.OnIdle, ...)`,
which only **enqueues** the OnIdle event. For a subscriber registered with
`-Action { ... }` (exactly what these tests register), PowerShell does **not** run
the action scriptblock inline at `GenerateEvent` time — it becomes a *pending
action* that the engine's event manager dispatches **asynchronously, on the
pipeline thread**, at the next process-pending-actions point (around subsequent
pipeline invocations). `OnPowerShellIdle` then runs a tiny artificial pipeline
(`…param() 0`) to nudge event processing, but the action's actual execution is
never synchronized with that pipeline returning or with the test's later
`$handled` read.

So the fixed `Thread.Sleep(2000)` was only a timing guess. On a fast runner the
action finishes first; on the slower WinPS leg 2 seconds is sometimes not enough,
leaving `$global:handled` still `$false` at the assertion — hence the intermittent
failure.

The key realization: each **additional** pipeline execution gives the engine
another chance to drain the pending action, so re-reading the handler variable in
a loop both **waits for** *and* **drives** completion. That makes polling a real
deterministic fix rather than just a longer sleep.

## Change

- Add a shared `WaitForHandledAsync(psesHost, variableName)` helper that polls the
  handler variable via `ExecutePSCommandAsync<bool>` until it reports `$true`
  (~200 ms between polls, ~15 s ceiling via `CancellationTokenSource`). On timeout
  it returns the last observed value so the existing `Assert.Collection(handled,
  Assert.True)` still fails loudly.
- Use it in `CanRunOnIdleTask` (`$handled`) and `CanRunOnIdleInProfileTask`
  (`$handledInProfile`), replacing both `Thread.Sleep(2000)` calls and removing the
  `// TODO: Why is this racy?` comments. Registration, the pre-assert
  (`Assert.False`), and the `OnPowerShellIdle` delegate call are unchanged.

No production code or test-profile script changes — test file only.

## Validation

- Built and ran both tests on **net8.0** (`Invoke-Build TestPS74 -TestFilter
  'FullyQualifiedName~CanRunOnIdle'`): green across repeated runs, ~0.4 s each vs.
  the old fixed 2 s.
- **net462 can't run on macOS**, but the dispatch mechanism is identical across
  targets — only latency differs. The 15 s ceiling is ~7.5× the old 2 s window and
  self-terminates on success, so it's strictly safer on the slow leg without
  slowing the fast one.

### DIFF
--- a/test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs
+++ b/test/PowerShellEditorServices.Test/Session/PsesInternalHostTests.cs
@@ -19,6 +19,38 @@ namespace PowerShellEditorServices.Test.Session
     using System.Management.Automation;
     using System.Management.Automation.Runspaces;
 
+    // Shared helpers for the OnIdle engine-event tests, whose handler actions are
+    // dispatched asynchronously by PowerShell's event manager.
+    internal static class OnIdleTestHelpers
+    {
+        // The OnIdle engine event's -Action scriptblock is not run inline when
+        // OnPowerShellIdle generates the event; PowerShell enqueues it as a pending
+        // action and dispatches it asynchronously around subsequent pipeline executions.
+        // So instead of sleeping a fixed amount, poll the handler variable until it
+        // reports true (each read is itself a pipeline, giving the engine another chance
+        // to drain the pending action). On timeout we return the last observed value so
+        // the caller's assertion still fails loudly.
+        internal static async Task<IReadOnlyList<bool>> WaitForHandledAsync(
+            PsesInternalHost psesHost, string variableName)
+        {
+            using CancellationTokenSource cancellationSource = new(millisecondsDelay: 15000);
+            IReadOnlyList<bool> handled = Array.Empty<bool>();
+            while (true)
+            {
+                handled = await psesHost.ExecutePSCommandAsync<bool>(
+                    new PSCommand().AddScript(variableName),
+                    CancellationToken.None);
+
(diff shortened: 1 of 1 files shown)


andyleejordan:
That seems reasonable to me!

JustinGrote (approved):
Ah the classic "fix a race condition by introducing a timeout". Good catch copilot!

andyleejordan:
@justingrote _and_ a polling loop.

JustinGrote:
@andyleejordan I mean I'll take a polling loop over a blind timeout any day, at least it's confirming its OK to proceed.

PR #2328, opened 2026-06-24
Fix Windows PowerShell host-start hang in `SetCorrectExecutionPolicy`

### LINKED ISSUE #2323: Windows CI hang: in-box Windows PowerShell attach E2E test wedges after the 20260614 runner image
## Summary

The Windows leg of CI (`CI Tests` workflow) hangs on the E2E debugger-attach
test `CanAttachScriptWithPathMappings` and rides GitHub Actions' default
6-hour job timeout. The root cause is **not** a PSES code change — it is a
`windows-latest` runner-image refresh that broke in-box **Windows PowerShell
5.1**'s cross-process attach. This issue tracks the real fix; PR #2318 is a
stopgap (skip the test on in-box WinPS + a 30-minute `timeout-minutes`
backstop).

## Root cause: runner-image refresh `20260608` → `20260614`

Comparing the last-green and first-red `main` runs:

- Last green (#2304) ran on image `win25-vs2026/20260608.135`.
- First red (#2303) and every red after it ran on image `20260614.141`.

Same image family, same runner agent (`2.335.1`), same `-Preview` PowerShell.
The only thing that moved at the boundary is the weekly OS-servicing patch in
the image. That refresh broke in-box Windows PowerShell 5.1's cross-process
`Debug-Runspace` / `Enter-PSHostProcess` attach — exactly the path this test
exercises. The precise servicing delta (likely a .NET Framework or Windows
named-pipe/IPC update) is still unidentified and is the main open question
here.

## Symptoms

- The hang is specifically the in-box Windows PowerShell E2E suite
  (`TestE2EPowerShell`). PowerShell Core (`TestE2EPwsh`) and the preview pass
  the same attach test; macOS and Linux pass in ~5 minutes.
- The test wedges on the attach handshake and the run sits on that single
  test until the job is killed.
- Its xUnit `[SkippableFact(Timeout = 15000)]` never fires: the wedge blocks
  in a way xUnit's cooperative timeout cannot abort, so it is reported as a
  "Long Running Test" climbing past 20+ minutes.

## What we ruled out

- **Not #2303 (strong-name identity change).** The per-commit bisection
  fingered #2303, but that was confounded: #2303 was the *first merge onto the
  new image*, so the image bump and the code change moved together. Two
  independent proofs exonerate it:
  - `PsesLoadContext.cs` (the file #2303 touched) is `<Compile Remove>`'d for
    `net462` and uses the Core-only `AssemblyLoadContext`. In-box WinPS 5.1
    loads `bin/Desktop/` (net462), so #2303's code is not even compiled into
    the hanging configuration.
  - With #2303 reverted on the #2318 branch, the in-box WinPS attach suite
    still hung on the new image (killed by the 30-minute cap).
- **Not the `-Preview` PowerShell.** The preview pwsh suite passes; it is the
  in-box Windows PowerShell suite that hangs. (An earlier draft of this issue
  hypothesized the preview — that is disproven.)
- **Not `ReadScriptLogLineAsync`.** The test wedges on the attach handshake
  before it reaches that helper. (PR #2318 still hardens that EOF/poll spin as
  general hygiene, but it is not the cause.)

## Mitigation (PR #2318)

- Skip `CanAttachScriptWithPathMappings` on in-box Windows PowerShell
  (`IsWindowsPowerShell`, covering the WinPS and WinPS-CLM suites) so the
  `windows-latest` leg completes; Core / preview / macOS / Linux keep full
  coverage of the attach path.
- `timeout-minutes: 30` on the matrix test job as a permanent backstop.

## Next steps toward a real fix

- Identify the precise `20260614` servicing delta (.NET Framework / Windows
  IPC) that changed in-box WinPS 5.1 attach behavior.
- Capture a hung Windows run's process/runspace state or DAP transcript to see
  where the attach handshake stalls under the new image.
- Investigate the harness's readiness logic (`RunWithAttachableProcess` keys
  off `RunspaceBase.AvailabilityChanged` via reflection) and the
  `Debug-Runspace` handshake for an OS-version-sensitive race.
- Once fixed, remove the `Skip.If` from #2318; the `timeout-minutes` backstop
  can stay.

## References

- Stopgap PR: #2318
- Original reported run: https://github.com/PowerShell/PowerShellEditorServices/actions/runs/27734424409

andyleejordan:
Fixes #2323.

On recent in-box Windows PowerShell 5.1 servicing builds, `SetCorrectExecutionPolicy`'s `Get-ExecutionPolicy -List` call autoloads `Microsoft.PowerShell.Security` into a runspace whose reused `InitialSessionState` already carries that module's `ObjectSecurity` type data, throwing "The member `AuditToString` is already present". That kills the pipeline thread during host startup while `TryStartAsync` is still awaiting queued startup tasks, so the server hangs and rides the CI timeout.

The execution-policy query is best-effort, so this wraps it in try/catch and skips policy configuration on failure (mirroring the existing `Set-ExecutionPolicy` handling), plus a guard against a short result list.

Because the hang is fixed, this also reverts the in-box Windows PowerShell E2E test skips from #2318. The `ci-test.yml` job-timeout backstop is kept.

### Known follow-ups
- Three LSP help tests (`Expand-Archive` synopsis) fail under in-box `powershell.exe` on my local older servicing build (.8655) but pass under `pwsh`; letting CI judge them on its build. The server returns auto-generated syntax instead of the MAML synopsis only in the WinPS server runspace — I could not isolate the mechanism locally.
- Separately, OmniSharp's `From(...)` silently swallows an exception thrown from the DAP `OnInitialize` handler, so a startup failure wedges instead of failing fast — I'll file a tracking issue.


src/PowerShellEditorServices/Services/PowerShell/Utility/PowerShellExtensions.cs
```diff
@@ -138,10 +138,31 @@ public static void SetCorrectExecutionPolicy(this PowerShell pwsh, ILogger logge
...
+                // rather than letting it abort host startup (which manifests as a hang).
+                logger.LogError(e, "Failed to query the execution policy; skipping execution policy configuration.");
+                return;
+            }
+
+            // We need at least the CurrentUser and LocalMachine scopes to proceed.
+            if (policies is null || policies.Count < 2)
+            {
+                return;
+            }
```
andyleejordan, line 165: @JustinGrote I got this whole issue to repro locally on a Windows box and apparently this was the issue that was hanging all the end-to-end tests on the updated image.

SeeminglyScience (approved):
LGTM!

